PR #302, opened 2022-12-13
Add ruleset template

### LINKED ISSUE #99: Ruleset templates
For very common experiment types, create templates which make it easy to put such an experiment together.

Start with documentation: e.g., take the `Base` rule file, and document: what does an experiment need to be defined? How do `Session`s keep track of `rounds_passed`?

Blocked by #928

albertas-jn:
Closes #99


backend/experiment/rules/rules_template.py
```diff
@@ -200,6 +201,6 @@ def get_feedback(cls, session):
             instruction=instruction,
             steps=[],
             button_label=button_label
-        ).action()
+        )
 
         return feedback
```
BeritJanssen: Missing emtpy line

BeritJanssen (commented):
I've been wondering whether we can make the `base.py` class the template, as all other rules files derive from that. It would be neat if the template is in itself a working experiment, then subclassing / overriding functions would make the first steps towards your own experiment easiest. Perhaps something to look into tomorrow?

drikusroor:
Should we close this PR now that we have a ruleset template in the `createexperiment` / `createruleset` command?

BeritJanssen:
We should incorporate this work as a starting place for the 2AFC paradigm, I think. I'm thinking `manage.py createruleset -t 2AFC` , with `-t` as an optional flag to switch between experiment types, and create an extra template based on this PR.

drikusroor:
> We should incorporate this work as a starting place for the 2AFC paradigm, I think. I'm thinking `manage.py createruleset -t 2AFC` , with `-t` as an optional flag to switch between experiment types, and create an extra template based on this PR.

Okay, in that case we might want to update the name of this PR and its corresponding issue to avoid confusion.

albertas-jn:
I have updated this PR to reflect new changes in our infrastructure. 

I am not sure if there is any need to use a command to "generate" this template, because it is already "generated" and sitting in the rules directory. I think we should merge it as it is, close this PR and reopen if we come up with new ideas.

backend/experiment/templates/consent/consent_rules_template.html
BeritJanssen, line 1: We don't need an empty consent form. Consent is now set through admin or otherwise will be lorem ipsum by default.

BeritJanssen:
You are perfectly right that we should merge this PR. I do think that instead of "ruleset_template", the experiment could be renamed to "2afc". And let's refer in the wiki to this experiment as a nice guideline of how to set up an experiment. Perhaps even ship a mini playlist in this repo to use it with? (Wiki / playlist: leave for later.) Calling it "ruleset_template" might suggest to users that this one is not an experiment in its own right, while I think it can be nicely tweaked for multiple purposes.

BeritJanssen (approved):

PR #539, opened 2023-08-07
Selenium Python tests for BeatAligment and Eurovision experiments

albertas-jn:


BeritJanssen (approved):
Nice! I find the Python code also much more readable than the IDE-generated one. Would it be an idea to use the json fixtures for playlists and experiments? Then we could also check `tests-selenium.ini` file into version control, as the slugs would be determined by the database dump in the fixtures. Can be left for later though.

albertas-jn:
We can definitely think about fixtures when we have other major tests (categorization, matching pairs) done.

PR #545, opened 2023-08-17
Feature er/tunetwins external

Evert-R:
This PR is a temporary solution to solve the latency with tunetwins when playing from Japan.
For every playcard an `<audio>` tag is created, with the same attributes as the main `<audio>` tag, but with an `id` labeled `buffer-<section.id>`.
These audio elements are never used, but preload the audio files in cache before the games begins.

### DIFF
--- a/frontend/src/components/PlayButton/PlayCard.js
+++ b/frontend/src/components/PlayButton/PlayCard.js
@@ -1,8 +1,25 @@
 import classNames from "classnames";
 
 import Histogram from "../Histogram/Histogram";
-
+import { MEDIA_ROOT } from "../../config";
 const PlayCard = ({ onClick, registerUserClicks, playing, section }) => {
+
+    // preload the section (temporary solution for ICMPC)
+    const audioBuffer = document.createElement("audio");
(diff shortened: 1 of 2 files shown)


frontend/src/components/PlayButton/PlayCard.js
```diff
@@ -1,8 +1,36 @@
...
+          setFirstRender(true);
+        }
+    }, [firstRender])
+    
+    const createBuffer = () => {
+        // preload the section (temporary solution for ICMPC)
+        const audioBuffer = document.createElement("audio");
+        audioBuffer.id = `buffer-${section.id}`;    
+        audioBuffer.src = MEDIA_ROOT + section.url;
+        audioBuffer.crossorigin = "use-credentials";
```
BeritJanssen, line 21: I don't think we want to set 'use-credentials' here, since then the external source won't play unless its response headers are set to certain values, which may not always be possible.

BeritJanssen: But will test for myself - of course, external audio will not be a problem anymore if we can preload everything from our own server.

PR #550, opened 2023-08-29
Feature/musical pref 2.2

### LINKED ISSUE #371: Change the question order and correct the Like question text for the Music Preferences
1. The questions should flip, so the Know goes first, followed by the Like question.
2. Number the questions and give slightly more space between the two questions. 
3. The text for Like question is wrong, it should be: How much do you like this song?

4531678798324_ pic_hd

### LINKED ISSUE #455: Music Preference: Version 2.2
The design for MP, aka final version of 2.2, is ready, the documents can be found on Teams.
The big change compared to version 1.1 is that in order to address any potential concerns regarding participant engagement, we have introduced the concept of unlocking levels as participants progress. Once they have completed a specific number of songs, they will unlock the "Love" level, followed by "Knowledge" and "Connection" levels in sequential order, where they will receive feedback from us.  And at the end of the experiment, participants can share the results of Chinese social media.  

[Draft_2.2_Final.pdf](https://github.com/Amsterdam-Music-Lab/aml-experiments/files/11712525/Draft_2.2_Final.pdf)

BeritJanssen:
Close #371 and close #455. To be able to use icons from the backend, this branch also introduces fontawesome webfonts - this slightly increases what needs to be cached on load. As fontawesome includes more icons than themify-icons, I decided to replace themify everywhere by fontawesome, and also removed the fontawesome React libraries - so some of the increased size is compensated. From now on, we can just use `<span class="fa-solid fa-something">` in order to get icons.

One request is unaddressed: the ability to show Chinese social media on the final page. This calls for a more rudimentary change to the way the `Social` component works, and I will make a separate issue out of this.


Evert-R:
Works great, I only noticed that smileys in the feedback of score.js are now solid, but used to be regular (in categorization). Looks fine though.

BeritJanssen:
True - for some reason, `fontawesome` doesn't support regular icons as free content. I don't think it's a major problem, though.

PR #553, opened 2023-08-29
Feature/remove consent 404

### LINKED ISSUE #549: Consent not found 'error' on new participant
When a participant loads an experiment for the first time, this 'error' shows up:
`server/result/consent_tunetwins_icmpc/:1     Failed to load resource: the server responded with a status of 404 (Not Found)`
We should find a more elegant way of dealing with a new participant, as this 'bug' has been reported on multiple occasions already.

BeritJanssen:
Close #549.

### DIFF
--- a/frontend/src/components/Consent/Consent.js
+++ b/frontend/src/components/Consent/Consent.js
@@ -15,7 +15,7 @@ const Consent = ({ title, text, experiment, participant, onNext, confirm, deny, 
         if (consent || (new URLSearchParams(url_query_string).get("participant_id"))) {
             onNext();
         }
-    }, [consent, onNext]);
+    }, [consent, onNext, url_query_string]);
 
     // Click on agree button
     const onAgree = async () => {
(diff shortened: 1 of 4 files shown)


frontend/src/components/Consent/Consent.js
```diff
@@ -15,7 +15,7 @@ const Consent = ({ title, text, experiment, participant, onNext, confirm, deny,
         if (consent || (new URLSearchParams(url_query_string).get("participant_id"))) {
             onNext();
         }
-    }, [consent, onNext]);
+    }, [consent, onNext, url_query_string]);
```
albertas-jn, line 18: Why is it needed to add `url_query_string` as an effect dependency? URL query string will not change during the experiment, and Consent will be re-rendered anyway at the beginning of the next experiment. It will not hurt, but seems unnecessary.
Otherwise, everything seems fine!

albertas-jn (approved):

frontend/src/components/Consent/Consent.js
BeritJanssen: This came from a React warning. Basically, any variable in a `useEffect` function is considered being subject to change.

Issue #561, opened 2023-09-26
Reorganize player config


BeritJanssen:
Currently, there are numerous player types, and the config includes settings that may or may not apply to them: e.g., animations aren't applicable to play buttons, 'play_once' is not applicable for Autoplay players.

Instead, we should subclass the `Playback` type, comparable to the `Question` class and subclasses, ideally replacing the `play_config` altogether.

PR #562, opened 2023-09-27
Feature/break actions

### LINKED ISSUE #251: Introduce a `Sequential` wrapper around Trials
So far,  some views cannot be captured in the `Trial` setup. The `Trial` setup presupposes that every experiment can be broken down into interchangeable parts, and that there are no views which cannot be created from smaller components.

One of the views for which this does not add up so far is `SongSync`: if we break down its components (recognition yes / no, if yes : countdown + right place yes / no) into separate views with their own calls to `session/experiment/result` on the server, we have less control over timing, as the response time of the server is unpredictable. It's also not straight-forward to use the `onNext` strategy, because it depends on the user's response to the first question to see whether they get to the next round right away, or if they proceed to the countdown with another questionaire.

My idea is to have a `Sequential` wrapper, which can wrap several trials, and control how we move between different possibilities. The advantage of such a mechanism is that we get less idiosyncratic behaviour about preparing views and handling results. The disadvantage is time investment, and that writing the rules for a SongSync view might become more lengthy than the original approach (though we can probably provide a convenience class or function). A potential risk of this approach is that we may refactor, only to notice that there are yet other experiment types which cannot be placed within this framework. I believe both `SongSync` and `Plink` could be rewritten this way though. Perhaps also good to look into the *Musical Pairs* game, and in how far this can be embedded in a `Trial` / `Sequential` setup.

NB: we might not even need a separate class if we format the `next_rounds` as a dictionary rather than an array. A Trial may then set a state, which is linked to the id of the view that should be triggered by a specific response.

### LINKED ISSUE #599: Fix styling changes
Compare current deployed app against the `break-actions` setup and match styling against the old styling.

BeritJanssen:
Close #251 . Close #599 . Instead of introducing a new `Sequential` action, or adding information to the `next_round` array, I found a much less invasive way of accomplishing views with some either/or logic, such as Plink or SongSync: in the config of a `Trial`, pass along conditions under which the current `next_round` array is not handled to the end, but new round data requested from the server instead.

To do:
- [x] check scoring is correct
- [x] check that styles are correct
- [x] correct round numbering


Evert-R:
SongSync with external files isn't working. In the SongSync wrapper the play_method isn't set, so it defaults to buffer. This was previously set in the config of SongSync, because playAudio was called from there. 
In hooked.py this is set as a constant at the top of the file, so maybe we can add this as an argument for song_sync from there, so it can be set from the rules file.

Evert-R (changes requested):
SongSync with external files isn't working. In the SongSync wrapper the play_method isn't set, so it defaults to buffer. This was previously set in the config of SongSync, because playAudio was called from there.
In hooked.py this is set as a constant at the top of the file, so maybe we can add this as an argument for song_sync from there, so it can be set from the rules file.

BeritJanssen:
Ah, good catch! In the meantime, I've started working on #561 - in there, I introduce logic that sets the play method based on the path / duration of the file, so hopefully these kind of oversights can be also prevented in the future.

Evert-R:
Styling bug:
In categorization the playbutton moves down everytime you push a button, except when a feedback smiley is presented.

BeritJanssen:
Can you share a screen capture of this problem? I think this only occurs after passing the training phase, right?

BeritJanssen:
I fixed the spacing of the buttons (the main difference with the current develop branch). Will make a separate issue of the jumping buttons - this is unrelated. It already happens in the current develop branch.

PR #565, opened 2023-10-03
Feature er/webaudio complete

Evert-R:
Draft setup for webaudio.
SongSync still needs to be integrated.

Use play_config in playback.py for the different options:
` - play_method:
          - 'BUFFER': Use webaudio buffers. (recommended for stimuli up to 45s)  
          - 'HTML': Use the HTML tag. (recommended for stimuli longer 45s)
  - external_audio (if set to True, web-audio api will be disabled)`

Defaults are set to:
`play_method: 'BUFFER' `
`external_audio: false`

In this draft the console shows:
Buffers object
Compensated Latency and the type of the response. (output latency isn't available in Safari)
play_method
external_audio


BeritJanssen:
Is there a technical difference between setting `play_method='HTML'` and `external_audio=True`? If not, we might also lose the latter setting, and just document that external files should always use the `HTML` method.

frontend/src/components/Playback/Autoplay.js
```diff
@@ -25,16 +26,23 @@ const AutoPlay = ({instruction, preloadMessage, onPreloadReady, playConfig, sect
         time.current = t;
     };
 
-
     // Handle view logic
     useEffect(() => {
         switch (state.view) {
-            case RECOGNIZE:                
+            case RECOGNIZE:
```
BeritJanssen, line 32: You can get rid of this whole useEffect loop if there is only one state, right? That would be awesome!

frontend/src/components/Playback/Playback.js
```diff
@@ -81,27 +103,57 @@ const Playback = ({
...
+                    latency = webAudio.initWebAudio();
+                }
+                // Store player index
+                setPlayerIndex(index);
+                // Determine if audio should be played
+                if (playConfig.mute) {
+                    setPlayerIndex(-1);
+                    audio.pause();
+                    return;
+                }
```
BeritJanssen, line 144: There is a lot of code repeated in this if / else condition. Once you remove the log statements, perhaps it would also be good to outfactor the differences between html / web audio into a utils file, and then deal with the differences there, so you can call, for instance, the utils method `pauseAudio(playMethod)`, and this method will handle the logic of whether to call pause via the audio element or the webAudio API.

frontend/src/components/Playback/Playback.js
```diff
@@ -216,6 +301,10 @@ const Playback = ({
         }
     };
 
+    if (!sectionsLoaded) {            
+        return <Loading loadingText={'loading'} />;
+    }
+
```
BeritJanssen, line 307: I think this will lose us the option to set the preload message and countdown. Perhaps use the `Preload` component here instead? Also, any text should always come from the backend, so it can be localized.

frontend/src/util/webAudio.js
```diff
@@ -33,3 +72,70 @@ export const changeGain = (level) => {
...
+    const remove = () => {
+        source.removeEventListener(event, _callback);
+    };
+    const _callback = () => {
+        remove();
+        callback();
+    };
+    source.addEventListener(event, _callback);
+    return remove;
+};
```
BeritJanssen, line 141: missing empty line

BeritJanssen (changes requested):
Nice! Apart from (obviously) removing the log statements, I think you can work more along the lines of the `stopPlaying` function in the `Playback` component, and move the if/else logic for the different play modes to smaller functions in more places. I'm a big fan of removing the state switching in the `Autoplay` component. But some things that the `Preload` component provides, i.e., a countdown and an option to show a message ("Get ready" or similar), may then be lost? I was already thinking how the Preload component should become part of the `Playback` component instead, so hoping this can be achieved without too much hassle.

Evert-R:
> Is there a technical difference between setting `play_method='HTML'` and `external_audio=True`? If not, we might also lose the latter setting, and just document that external files should always use the `HTML` method.

Yes, with audio files above 45 seconds you would want to use the `HTML` method, but still use webaudio.
`external_audio` will disable webaudio.

frontend/src/components/Playback/Autoplay.js
```diff
@@ -1,108 +1,78 @@
...
-            default:
-            // nothing
+    useEffect(() => {        
+        let latency = 0;
+        // Play audio at start time            
+        if (!playConfig.mute) {
+            if (playConfig.play_method === 'BUFFER' && !playConfig.external_audio) {
+                console.log('Autoplay buffer');
+                latency = webAudio.getTotalLatency();
+                webAudio.playBufferFrom(section.id, 0, Math.max(0, playConfig.playhead));
```
BeritJanssen, line 28: Won't this always play the file from start? (Second argument is 0)

frontend/src/components/Preload/Preload.js
```diff
@@ -20,17 +22,44 @@ const Preload = ({ instruction, pageTitle, duration, url, onNext }) => {
...
+            // Use Web-audio and preload sections in buffers            
+            sections.map((section, index) => {
+                // Clear buffers if this is the first section
+                if (index === 0) {
+                    webAudio.clearBuffers();
+                }
+                
+                // Load sections in buffer
+                // Return remove listener  
+                return webAudio.loadBuffer(section.id, MEDIA_ROOT + section.url, () => {
```
BeritJanssen, line 39: The comment is not correct, right? What's being returned here is an async request. `useEffect` *should* return its own cleanup, but this `return` is from the `map` method, hence, nothing is returned from the `useEffect` hook itself.

frontend/src/components/Playback/Playback.js
```diff
@@ -75,33 +76,66 @@ const Playback = ({
...
+            if (playConfig.play_method === 'BUFFER' && !playConfig.external_audio) {
+                console.log('Play buffer (local)');                
 
-            // Cancel active events
-            cancelAudioListeners();
+                // Determine latency for current audio device
+                latency = webAudio.getTotalLatency();
+
+                // Store player index
+                setPlayerIndex(index);
```
BeritJanssen, line 92: what is the `playerIndex` used for? I think we need to clean something up here -- not necessarily in this branch though -- because there is a `prevPlayerIndex` and `lastPlayerIndex` ref in this component, and it's not documented what it's used for.

frontend/src/components/Playback/Playback.js
```diff
@@ -115,43 +149,78 @@ const Playback = ({
...
     const onFinishedPlaying = useCallback(() => {
         setPlayerIndex(-1);
-        audio.stop();
+        stopPlaying();
         finishedPlaying && finishedPlaying();
     }, [finishedPlaying]);
 
     // Stop audio on unmount
     useEffect(
         () => () => {
```
BeritJanssen, line 222: perhaps this `useEffect` could be combined with the `cancelAudioListeners` one.

BeritJanssen (commented):
I made some new comments on where documentation might be adjusted / improved.

frontend/src/components/Playback/Autoplay.js
Evert-R: That's indeed a typo.

BeritJanssen (approved):

PR #579, opened 2023-10-08
Fix requesting sections twice in Categorization experiment

### LINKED ISSUE #223: Sections are requested twice
When looking at the Network traffic, it is apparent that sections are often requested twice: once during loading, and again when playback starts. So we should
- investigate when and why this happens
- make sure we don't make unnecessary requests to the backend

albertas-jn:
Closes #223 

@BeritJanssen and @Evert-R 
This is caused by preloading of the first section in `useEffect` of `Playback.js`. The section is loaded again in `playSection()` when the Play button is clicked. I disabled the preloading and the Categorization, BeatAlignment, Eurovision and Tunetwins experiments seem to work fine. Do you foresee any potential problems with other experiments?

### DIFF
--- a/frontend/src/components/Playback/Playback.js
+++ b/frontend/src/components/Playback/Playback.js
@@ -49,9 +49,9 @@ const Playback = ({
     }, [playerIndex]);
 
     // Preload first section
-    useEffect(() => {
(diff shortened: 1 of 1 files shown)


albertas-jn:
Solved by #565

PR #593, opened 2023-10-26
Move click event to input (label double submits)

### LINKED ISSUE #589: "Double submit detected"
Error message appears in the console when using the `Form` component. Perhaps the component doesn't use the `Button` component (which detects clicks and prevents processing further click events)?

Evert-R:
The onclick and keypress events attached to a label will submit twice.
In this PR I attached these events to the input of the buttonArray. 

resolves #589


BeritJanssen (changes requested):
Nice that you found the problem so quickly! The log messages still need to go, though.

PR #612, opened 2023-11-13
Musicgens-scale-update

jaburgoyne:
Revised scale-point descriptions and added new questions to the MusicGens list (as requested at the September consortium meeting).


BeritJanssen:
@jaburgoyne , we can "enforce" experiments to always be shown in English. Would this be sufficient to ensure we're keeping consortium members happy while already integrating the capability to translate?

jaburgoyne:
Yes, absolutely, @BeritJanssen – good solution!

PR #631, opened 2023-12-02
Add buttons to show/hide/select different question groups in experiment admin

albertas-jn:


backend/experiment/static/experiment_admin.css
```diff
@@ -0,0 +1,24 @@
...
+.buttons-column button {
+	display: block;
+	padding: 2px 5px;
+	margin: 5px 0px 0px 0px;
+}
+
+form .button-show-hide {
+ 	width: 45px;
+	padding: 2px 5px;
+	margin: 0px 10px 0px 0px;
```
BeritJanssen, line 23: Instead of absolute pixel values, better use relative values such as vh / vw. While we would expect most users not to use the admin interface with mobile devices, during field work, it would be nice if the admin interface scaled.

BeritJanssen (changes requested):
As discussed, perhaps we could introduce a loading spinner while the default questions are fetched, though this can also be left for later.
In the overview of questions categories, I think it would help to use bold face for them, or in another way emphasize them. Visually, the titles are less prominent than the buttons right now.
AdminInterface

backend/experiment/static/experiment_admin.css
BeritJanssen: We discussed, Django Admin actually uses absolute values all over, so it may actually lead to unexpected results to use relative values for this custom addtion.

backend/experiment/static/experiment_admin.css
```diff
@@ -0,0 +1,24 @@
+.buttons-row {
+	margin: 0px 0px 0px 10px;
+}
+
```
BeritJanssen, line 4: Still missing css to emphasize the question titles here.

backend/experiment/static/experiment_admin.js
```diff
@@ -0,0 +1,150 @@
...
+
+	let loader = document.createElement("div")
+	loader.className = "loader"
+	loader.style.display = "none"
+	loader.innerText = "Loading..."
+	buttonsColumn.append(loader)
+	
+	groups.forEach( (group) => {
+
+		group.style.fontWeight = "bold"
```
albertas-jn, line 36: It's here

BeritJanssen (approved):

PR #637, opened 2023-12-05
Config: Configure test workflow to run on pushes to all branches

drikusroor:
Before, the `test` workflow would only run when pushing to `develop` or `main`, or on the `pull_request` trigger. Now, the `test` workflow will be triggered on a push to _any_ branch as per the varying preferences in the team.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,9 +2,6 @@ name: Test suite
 
 on:
   push:
-    branches:
-      - main
-      - develop
   pull_request:
 
 jobs:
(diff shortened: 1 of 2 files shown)


albertas-jn (approved):
This looks OK.
Also, coming back to #613, I think it would be better to move all test files from `scripts` to `tests`.

drikusroor:
> Also, coming back to #613, I think it would be better to move all test files from `scripts` to `tests`.

I actually disagree to this. The tests are technically stored somewhere else. The scripts folder might be a good place for other scripts in the future as well.

PR #639, opened 2023-12-07
Docs: Add instructions for installing Python dependencies

drikusroor:
This PR documents how to add python dependencies to the Django backend. Instead of reading markdown, you might want to look at it here: https://github.com/Amsterdam-Music-Lab/MUSCLE/tree/docs/add-python-deps/backend#install-packages

* Does the added entry in `backend/README.md` accurately and correctly describe the process of adding and installing new Python dependencies?
* Should it be documented in `backend/README.md`, or perhaps in the `README.md` in the root of the repository or perhaps even in the [Wiki](https://github.com/Amsterdam-Music-Lab/MUSCLE/wiki)?

## Update

- I've moved the documentation to the wiki: https://github.com/Amsterdam-Music-Lab/MUSCLE/wiki/X.-Installing-python-packages
- I've added a script that automatically runs the two `pip-compile` commands

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -88,6 +88,32 @@ from .views import MyView
 
 Check the client documentation for instructions for adding the view to the client.
 
+## Install packages
+
+If you want to install new packages, you can do so by adding the dependencies to:
+* `requirements.in/base.txt` for packages that are needed in both *development* **and** *production*
+* `requirements.in/dev.txt` for packages that are only needed in *development*
+* `requirements.in/prod.txt` for packages that are only needed in *production*
+
+Then, run the following `pip-compile` commands to generate the `requirements.txt` files:
+
+```bash
+# Install all packages for development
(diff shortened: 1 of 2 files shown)


BeritJanssen:
Nice, and necessary to have this documentation! NB for some reason, the changes for "troubleshooting" don't show in the overview of "Files changed". So perhaps double-check everything is committed? Don't understand what could've caused this. I would tend to put this in the wiki, but I realize that this kind of information may also be nicer to be kept with the repo itself, so I also wouldn't be opposed to having a folder `documentation` at this repo's top level, and move all the technical information there, with descriptive file names.

BeritJanssen (changes requested):
Nice! As stated in the comment, I'd prefer to add this information to the wiki, or -- probably easier for other developers -- add a documentation folder at the repo's root where this info goes. If you want to go that route, could you make an issue to add the other `X` info from the wiki there?

PR #640, opened 2023-12-07
Test: Add more frontend unit tests

drikusroor:


frontend/src/hooks/useSingleToArray.test.js
```diff
@@ -0,0 +1,35 @@
+import { renderHook } from '@testing-library/react-hooks';
+import useSingleToArray from './useSingleToArray';
+
+describe('useSingleToArray', () => {
```
BeritJanssen, line 4: I think the `useSingleToArray` function may actually be superficial now, as we enforce sections being sent as an array on the backend.

frontend/src/util/label.test.js
```diff
@@ -0,0 +1,52 @@
+import { render } from '@testing-library/react';
+import { getPlayerLabel, LABEL_NUMERIC, LABEL_ALPHABETIC, LABEL_ROMAN, LABEL_CUSTOM, renderLabel } from "./label";
+
+describe('getPlayerLabel', () => {
+
+    it('returns numeric label correctly', () => {
+        expect(getPlayerLabel(0, LABEL_NUMERIC)).toBe(1);
```
BeritJanssen, line 7: Working on moving the `getPlayerLabel` logic to the backend at the moment.

BeritJanssen (approved):
Awesome!

frontend/src/hooks/useSingleToArray.test.js
drikusroor: I checked and it doesn't seem to be used anymore in the whole frontend application and shall therefore remove it and its tests from the codebase.

afbeelding

frontend/src/util/label.test.js
drikusroor: I've deprecated the function to prevent ourselves from using it in the future. Also, do you have the specific issue (if exists) that describes moving that logic to the backend?

PR #643, opened 2023-12-08
CI: Add linting jobs

drikusroor:
This PR adds linting jobs for both the frontend (`eslint`) and backend (`flake8`). The question right now is whether our linting settings/preferences are configured correctly and to our liking. Additionally, the linting job for the backend "fails", although I allowed it to fail. Maybe we can alter the `flake8` configuration so that it won't fail anymore, or we can allow it to fail and slowly fix all issues.


BeritJanssen:
I think allowing to fail and fix issues over time is the preferable option.

BeritJanssen (approved):

PR #644, opened 2023-12-08
Fix Eurovision2020 and Kuiper2020 expected_result in next_heard_before_action()

### LINKED ISSUE #632: Hooked on Christmas: bonus rounds are always wrong
The Hooked on Christmas now runs through without error, but in the bonus round, every answer is taken to be wrong, no matter if the song was actually played before or not.

albertas-jn:
Closes #632


drikusroor:
To prevent a bug like this from happening again, would it be wise to add a unit test for this? (I don't know how easy that is at this stage for this experiment)

BeritJanssen:
This bug is caused by underdocumented code, written before any of us stepped into the project. We should make sure to get all the details down asap. Indeed, documenting in the form of tests is a good solution.

BeritJanssen (approved):
Ah, good catch!

PR #646, opened 2023-12-11
Migrate jsondata serialize export

### LINKED ISSUE #537: Add section list to default JSON exports
We've expanded our JSON exports to include good information about sessions and participants as separate trees. Given this structure, it would be convenient also to include a tree listing information about *sections*. Although this will duplicate the playlist CSV file in many cases, for debugging and future analysis after an experiment may have been taken down, it would be useful to have a record of what metadata the MUSCLE infrastructure had attached to a particular section_id for a particular experiment.

In short, an ideal JSON export should dump the complete data structure for sessions, participants, *and* sections.

Evert-R:
This PR migrates the `json_data` field of the `Session` and `Result` models from `CharField` to `JSONField`
It's recommended to backup the data base before running this PR.

For compatibility the `load_json_data` and `save_json_data` function of both models have been modified to work with the new JSONField, but `json_data` can now be used directly as a `dict` as well.

The experiment `Export JSON` function in the admin will now offer a zip file for download containing all related experiment data as raw django model objects in  6 JSON files, using the native django serializer.
`participants.json`, `profiles.json`, `results.json`, `sections.json`, `sessions.json`, `songs.json`

resolves #537


backend/result/migrations/0006_remove_result_old_data.py
BeritJanssen, line 1: It would be nice if the migration files 0004-0006 could have more descriptive names.

backend/session/migrations/0003_auto_20231123_1339.py
```diff
@@ -0,0 +1,25 @@
...
+
+def forwards_func(apps, schema_editor):
+    Session = apps.get_model('session', 'Session')
+    for session in Session.objects.all():
+        if session.json_data:
+            for key,value in json.loads(session.json_data).items():
+                session.json_temp[key] = value
+                session.save()
+                
+def reverse_func(apps, schema_editor):
```
BeritJanssen, line 14: I believe you'll also need to use the `apps.get_model` mechanism for the reverse data migration.

backend/session/migrations/0004_auto_20231124_1146.py
BeritJanssen, line 1: More descriptive names for 0003 / 0004 would be nice.

backend/experiment/admin.py
```diff
@@ -34,14 +42,45 @@ class ExperimentAdmin(InlineActionsModelAdminMixin, admin.ModelAdmin):
...
+        all_songs = Song.objects.none()
+        all_sections = Section.objects.none()
+        all_participants = Participant.objects.none()
+        all_profiles = Result.objects.none()
+
+        # Collect data
+        all_sessions = obj.export_sessions().order_by('pk')
+
+        for session in all_sessions:
+            all_results |= session.export_results()
```
BeritJanssen, line 58: I'm not familiar enough with the `|=` operator to judge: but won't we overwrite the list of `all_results` for every session, hence only export the last session's data? Or does the `|=` operator actually merge these?

backend/experiment/tests/test_admin_experiment.py
```diff
@@ -66,4 +66,4 @@ def test_admin_export(self):
         experiment = Experiment.objects.first()
         response = this_experiment_admin.export(request, experiment)
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(response['content-type'], 'application/json')
+        self.assertEqual(response['content-type'], 'application/x-zip-compressed')
```
BeritJanssen, line 69: I find this test way too limited. I would want to know that for a toy experiment with several sessions, including linked results, sections and participants, we get the relevant data.

BeritJanssen (changes requested):
Wonderful to have done with the akward `json.dumps` and `json.loads`! Can you rename the `auto...` migration files so it's apparent from their name what they do? And please add a more thorough test of the admin export? That one's going to take some work, but I don't think testing that we get a zip file is enough here, we want to know that the contents of the zip file match the contents of a test database.

backend/experiment/admin.py
Evert-R: I had to search for this as wel, but `|` seems to be the way to concatenate django queries. So `|=` will concatenate the new query to the existing.

backend/session/migrations/0003_auto_20231123_1339.py
Evert-R: Good catch !

PR #648, opened 2023-12-11
Feature/player config

### LINKED ISSUE #561: Reorganize player config

BeritJanssen:
Close #561 : instead of a lose config, set arguments to Playback, and subclass player types.


albertas-jn (approved):
I hope this will be well tested before merging.

albertas-jn (changes requested):
Some of the tests fail.

PR #649, opened 2023-12-11
Chore(deps): Migrate to React 18

### LINKED ISSUE #442: Update React
We're one major version behind.

drikusroor:
Resolves #442 

This PR aims to migrate React from version 17 to version 18, which is the most recent version of React. [See here for what is new in React 18](https://react.dev/blog/2022/03/29/react-v18).

Everything seems to work fine after the migration and some additional tweaks. However, after using React's [`StrictMode`](https://react.dev/reference/react/StrictMode) component, I found that `findDOMNODE`, which comes from the `react-transition-group` dependency, is deprecated. I've tried if updating it would do anything but alas. It's not a urgent issue I think, but perhaps we need to address it in the future.


BeritJanssen:
Thanks for doing this! Will the app still build from this branch? I don't see the `findDOMNode` function used anywhere, and the components imported from "react-transition-group" - `TransitionGroup` and `CSSTransition` - were introduced before my time. Does the `StrictMode` wrapper for dev only mean we suppress some warnings during build time?

drikusroor:
> Thanks for doing this! Will the app still build from this branch? I don't see the `findDOMNode` function used anywhere, and the components imported from "react-transition-group" - `TransitionGroup` and `CSSTransition` - were introduced before my time. Does the `StrictMode` wrapper for dev only mean we suppress some warnings during build time?

`StrictMode` warns us about incorrect patterns or deprecated functions, which is only useful on our local develop environments, so I've turned it off for production. It's technically a component so if it would have been turned on, warnings can show up in the console on the acc or prod website.

BeritJanssen (approved):

Issue #609, opened 2023-11-08
Create Six Card Tutorial for TuneTwins Congo Branch


davidjohnbaker1:
Create module of TuneTwins where scientist is able to select running a tutorial, then a pared down version of TuneTwins (#605) plays with only 6 cards. This is done to give people practice before moving on to the 16 board.

BeritJanssen:
The number of pairs can easily be adjusted. At the moment, this would have to happen through a separate rules file, though we may also consider whether to use the `number of rounds` setting in the admin interface as `number of pairs`. Might be confusing for future maintenance though.

BeritJanssen:
@davidjohnbaker1 , I believe this was off the table? I'll close the issue, feel free to reopen if you decide this is needed after all.

davidjohnbaker1:
Yes, this can be closed.

PR #651, opened 2023-12-11
stripped down tutorial intro no CSS

davidjohnbaker1:
First attempt at seeing if I can help out more on the small Congo related tasks.

Attempted #609 , with the following:

- Copied matching pairs to make matching pairs tutorial file
- Added in call to rules in init 
- removed unnecessary code other that start and do it again (looking at this level of minimalism for congo work)

Not sure if

- data will save, no need for it to, but have not totally figured out save structure

New Issues

- Will need to make a new issue to fix the CSS so it's a 2x3 grid, but didn't want to make a hack solution to this on the front end that just creates code duplication, will file this right away for future. If you have suggestions on best ways to do it (was going to just make Matching Pairs.js/scss duplicate tutorial set of files, then change all the calls from the new matching_pairs_tutorial.py file, but that didn't seem like the most elegant want to do it. 

Please let me know if these kinds of small fixes with my limited webdev skills are ultimately going to help out or just make more work for everyone. 

Also did this by cloning off main (as I got errors when I ran off of development).

### DIFF
--- a/backend/experiment/rules/matching_pairs_tutorial.py
+++ b/backend/experiment/rules/matching_pairs_tutorial.py
@@ -1,0 +1,123 @@
+import random
+
+from django.utils.translation import gettext_lazy as _
+from django.template.loader import render_to_string
+
+from .base import Base
+from experiment.actions import Consent, Explainer, Final, Playlist, StartSession, Step, Trial
+from experiment.actions.playback import Playback
+from experiment.questions.demographics import EXTRA_DEMOGRAPHICS
+from experiment.questions.utils import question_by_key
+from result.utils import prepare_result
+
+from section.models import Section
+
+class MatchingPairsTutorial(Base):
+    ID = 'MATCHING_PAIRS_TUTORIAL'
+    num_pairs = 3
+    contact_email = 'aml.tunetwins@gmail.com'
+
+    def __init__(self):
+        self.questions = [
+            #question_by_key('dgf_gender_identity'),
+            #question_by_key('dgf_generation'),
+            #question_by_key('dgf_musical_experience', EXTRA_DEMOGRAPHICS),
+            #question_by_key('dgf_country_of_origin'),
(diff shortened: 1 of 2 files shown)


backend/experiment/rules/matching_pairs_tutorial.py
BeritJanssen, line 1: This whole file can be reduced to three lines:
```
from .matching_pairs import MatchingPairs

class MatchingPairsTutorial(MatchingPairs):
    ID = 'MATCHING_PAIRS_TUTORAL'
    num_pairs = 3
```

No need to repeat any of the other code then, and it's possible to change the questions from the admin interface.

BeritJanssen (changes requested):
Using inheritance, we can make the rules file for the MatchingPairsTutorial *much* shorter. Every method from the parent class will be reused, and you only only override the `ID` and `num_pairs class` variables.

BeritJanssen:
Thanks for your reasoning to make a new issue. It would indeed not be desirable to make a new frontend component specifically to deal with less cards in a 2x3 grid.

davidjohnbaker1:
Ah, more more elegant! 

Though for the tutorial, we do not want the intro text or all the stuff at the end, just a basic loop that goes :

ready to play -- six card game -- play again? 

This will never be used online, just by lia in the congo. 

I've amended the commit to reflect that, but defer to you on if what i've pushed is the best way to deal with this. 

Also, I was not able to get this code to run only with a call to 

`from .matching_pairs import MatchingPairs`

but did try to leave only the required library calls.

drikusroor:
@davidjohnbaker1 @BeritJanssen I see this PR has been in a "Changes requested" state for over a month now. Is there anything currently blocking it from going forward?

BeritJanssen:
As far as I'm aware, the "tutorial with fewer cards" feature was abandoned for now, so I believe we can close this?

davidjohnbaker1:
Yes, this was abandoned. We mis-communicated about this, I will close the issue for now.

Issue #652, opened 2023-12-11
Fix CSS for Six Card Tutorial


davidjohnbaker1:
Following merge of #609 , will need to change CSS to have 2x3 grid (2 rows, 3 columns)

drikusroor:
Hi @davidjohnbaker1, does only your own specific experiment (The Six Card tutorial) need a 2x3 grid? I assume it shouldn't cause the already existing Matching Pairs experiment to break, right?

PR #657, opened 2023-12-12
Added: Add 2x3 grid support for Matching Pairs component if it has equal or less than 6 sections

drikusroor:
Related to #652 

## How to test this?

Creating an MatchingPairs-like experiment with six or less sections should result into the `MatchingPairs` component to be displayed with only two columns.

A quicker route may be to look a the component in Storybook. ~You can do this by running the following command: `docker compose run --rm client yarn run storybook`, although it (connecting to Storybook running inside of a Docker container) does not seem to work on my pc (anymore?) at the time of writing this...~ 😓

**Update:**

The correct Storybook command is `docker compose run --rm -p 6006:6006 client yarn run storybook` (which opens the correct port).


davidjohnbaker1:
In the comment here, you note 2 x 3 as two columns, three rows, but I should have been more clear that it should be three columns, two rows (have not tested this yet).

drikusroor:
> In the comment here, you note 2 x 3 as two columns, three rows, but I should have been more clear that it should be three columns, two rows (have not tested this yet).

Ah, good to know! 😄 I've updated #652 accordingly. Shouldn't take much time to update the code to support 3 columns instead of 3. 👍

drikusroor:
It should show 3 columns now if there are 6 or less sections. You can check it by pulling the branch and running Storybook with the docker command from the PR description ;-)

davidjohnbaker1:
Trying to check this, but running into trouble with the `storybook` part of it. Is this a new dependency? I updated my node and yarn and also ran an install of both after that, but still says the command is not found. If this is a common web tool I am missing, let me know.

drikusroor:
> Trying to check this, but running into trouble with the `storybook` part of it. Is this a new dependency? I updated my node and yarn and also ran an install of both after that, but still says the command is not found. If this is a common web tool I am missing, let me know.

Storybook is a newish dependency that has been merged to the `develop` branch already. However, to see the 3 columns, you need to be on this (`feat/matching-pairs-columns`) branch. If the `storybook` command doesn't work for you, there could be several reasons, the most likely being that your Docker container is not up-to-date.

If you want to run it through Docker:

```sh
git fetch
git checkout feat-matching-pairs-columns
git pull

# the storybook command is located in frontend/package.json, 
# which get copied in the build. See also frontend/Dockerfile
docker compose build client
docker compose run --rm -p 6006:6006 client yarn run storybook
```

If you are just using node (18+) + yarn:

```sh
git fetch
git checkout feat-matching-pairs-columns
git pull
cd frontend

yarn run storybook
```

davidjohnbaker1:
Yes, this looks great! I won't be the one to close or merge this, but when I run the MP with 3 pairs on this branch, it looks just like what I'd imagine.

BeritJanssen (approved):
Nice! Perhaps in the long run we can add more layout options and rules. Thanks for adding this one.

PR #659, opened 2023-12-13
Fix problem with "sync" playback in SongSync

### LINKED ISSUE #662: "sync" phase of a Hooked round doesn't resume where it left off
In the current `develop` branch, the original `SongSync` view was split into three separate Trials:
- one for the "recognition" phase (play tune, ask user if they recognize it)
- one for the "silence" phase (4 seconds of silence)
- one for the "sync" phase (play tune from where recognition time left off + 4 seconds of silence (+ optionally a random additional  shift of 1-15 seconds)

The "sync" stage does not correctly handle the state from the "regognition" phase: i.e., it plays from the start of the file + 4 seconds + optional additional shift, not from the moment the user clicked "yes" in the "recognition" phase.

BeritJanssen:
Close #662. The current solution saves `decisionTime` in `window.sessionStorage` every time the player clicks a form during a `Trial`. For every playback view, we can now set whether we want to `resume_play` from where a previous player left off.

Will experiment with whether this could also be achieved with a useState hook at the `Experiment` level, but this approach works for the time being.

### DIFF
--- a/backend/experiment/actions/playback.py
+++ b/backend/experiment/actions/playback.py
@@ -24,6 +24,7 @@ class Playback(BaseAction):
             - show_animation: whether to show an animation during playback 
             - (multiplayer) label_style: player index number style: NUMERIC, ALPHABETIC, ROMAN or empty (no label)
             - play_once: the sound can only be played once
+            - resume_play: if the playback should resume from where a previous view left off
     '''
 
     TYPE_AUTOPLAY = 'AUTOPLAY'
@@ -45,6 +46,7 @@ class Playback(BaseAction):
             'show_animation': False,
             'mute': False,
             'play_once': False,
+            'resume_play': False
         }
         if play_config:
             self.play_config.update(play_config)
--- a/backend/experiment/actions/wrappers.py
+++ b/backend/experiment/actions/wrappers.py
@@ -96,7 +96,8 @@ def song_sync(session, section, title, response_time=15, play_method='BUFFER'):
                         play_config={
             'ready_time': 0,
             'playhead': silence_time + (random.randint(100, 150) / 10 if not continuation_correctness else 0),
-            'show_animation': True
+            'show_animation': True,
+            'resume_play': True
         }),
         config=trial_config,
         title=title
--- a/frontend/src/components/Playback/Playback.js
+++ b/frontend/src/components/Playback/Playback.js
@@ -23,7 +23,7 @@ const Playback = ({
     sections,
     instruction,
     onPreloadReady,
-    preloadMessage,
+    preloadMessage = '',
     autoAdvance,
     responseTime,
     playConfig = {},
@@ -104,10 +104,7 @@ const Playback = ({
                         pauseAudio(playConfig);
                         return;
                     }
-                    // if the current player has resume_play set to true,
-                    // retrieve previous player's decisionTime from sessionStorage
-                    const playheadShift = playConfig.resume_play ? 
-                        parseFloat(window.sessionStorage.getItem('decisionTime')) : 0;
+                    const playheadShift = getPlayheadShift();
                     let latency = playAudio(playConfig, sections[index], playheadShift);
 
                     // Cancel active events
@@ -134,6 +131,14 @@ const Playback = ({
             },
             [playAudio, pauseAudio, sections, activeAudioEndedListener, cancelAudioListeners, startedPlaying, onAudioEnded]
     );
+
+    const getPlayheadShift = () => {
+        /* if the current Playback view has resume_play set to true,
+        retrieve previous Playback view's decisionTime from sessionStorage
+        */
+        return playConfig.resume_play ? 
+        parseFloat(window.sessionStorage.getItem('decisionTime')) : 0;
+    }
 
     // Local logic for onfinished playing
     const onFinishedPlaying = useCallback(() => {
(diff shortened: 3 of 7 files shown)


drikusroor:
> The current solution saves `decisionTime` in `window.sessionStorage` every time the player clicks a form during a `Trial`. For every playback view, we can now set whether we want to `resume_play` from where a previous player left off.

With "The current solution", do you mean the proposed solution in this PR or the current situation in `develop/main`? I don't intend to be nitpicky about this but would like to be on the same page. ;)

> Will experiment with whether this could also be achieved with a useState hook at the `Experiment` level, but this approach works for the time being.

Right now you store the value on the `"decisionTime"` key in the session storage. Therefore, the value is not experiment-specific. What happens if a participants decides to do another experiment, will it have an effect somehow?

frontend/src/components/Playback/Playback.js
```diff
@@ -105,7 +104,11 @@ const Playback = ({
                         pauseAudio(playConfig);
                         return;
                     }
-                    let latency = playAudio(playConfig, sections[index]);
+                    // if the current player has resume_play set to true,
+                    // retrieve previous player's decisionTime from sessionStorage
+                    const playheadShift = playConfig.resume_play ? 
+                        parseFloat(window.sessionStorage.getItem('decisionTime')) : 0;
```
drikusroor, line 110: I think it would be better to make a dedicated helper class or function for fetching and storing values from and to the session storage. It could be in this component, or perhaps in the `util` folder if we think we're going to use it more often than once.

Added bonus is that we can spy on that helper method to see if it gets called in our unit tests.

frontend/src/components/Trial/Trial.js
```diff
@@ -45,19 +43,21 @@ const Trial = ({
...
             const form = feedback_form ? feedback_form.form : [{}];
 
             if (result.type === "time_passed") {
                 form.map((formElement) => (formElement.value = "TIMEOUT"));
             }
 
             if (feedback_form) {
+                const decisionTime = getTimeSince(startTime.current);
+                // keep decisionTime in sessionStorage to be used by subsequent renders
+                window.sessionStorage.setItem('decisionTime', decisionTime);
```
drikusroor, line 55: Same comment as before

frontend/src/util/audioControl.js
```diff
@@ -1,15 +1,16 @@
 import * as audio from "./audio";
 import * as webAudio from "./webAudio";
 
-export const playAudio = (playConfig, section) => {    
+export const playAudio = (playConfig, section, playheadShift=0) => {
```
drikusroor, line 4:
```suggestion
export const playAudio = (playConfig, section, playheadShift = 0) => {    
```

frontend/src/components/Playback/Playback.js
```diff
@@ -105,7 +104,11 @@ const Playback = ({
...
+                    let latency = playAudio(playConfig, sections[index], playheadShift);
```
drikusroor, line 111: Additionally, I think it might be nice to combine this whole bit in a separate function named something like `startOrContinuePlayAudio`, e.g. something that explains what is happening. This whole bit means it continues the audio from where the participant left off, if it actually left off, right?

backend/experiment/actions/playback.py
```diff
@@ -45,6 +46,7 @@ def __init__(self, sections, player_type='AUTOPLAY', preload_message='', instruc
             'show_animation': False,
             'mute': False,
             'play_once': False,
+            'resume_play': False
```
drikusroor, line 49: Why is `resume_play` `False` here?

backend/experiment/actions/wrappers.py
```diff
@@ -96,7 +96,8 @@ def song_sync(session, section, title, response_time=15, play_method='BUFFER'):
                         play_config={
             'ready_time': 0,
             'playhead': silence_time + (random.randint(100, 150) / 10 if not continuation_correctness else 0),
-            'show_animation': True
+            'show_animation': True,
+            'resume_play': True
```
drikusroor, line 100: And why is it `True` here? :D

drikusroor (approved):
I haven't tested it manually but it looks like it should work as expected. I do think some logic can be abstracted away to helper functions or even shared helper functions outside of the component.

Additionally, the PR mentions it fixes a problem, but I don't see what the problem is, who has it, and why it is a problem exactly. To me it seems more like a new useful feature. Some context would therefore be nice.

Lastly, in the PR description you mentioned about experimenting with this at the Experiment level with a `useState` hook. Perhaps you could also use this opportunity to experiment with a more global state management solution, be it `useContext`, `redux`, or `zustand` (the last one being my current favorite, just saying 😅 )

I'll leave it up to you what you want to do with my comments as I'm not up to speed with the importance and urgency of this PR :-)

backend/experiment/actions/playback.py
BeritJanssen: That's the default.

backend/experiment/actions/wrappers.py
BeritJanssen: We only set it to `True` in any view that is supposed to "resume play", which is the case in this particular stage of the Hooked experiment, where a player gets to hear a continuation after a silence time and has to indicate whether the track continued at the expected, or unexpected moment.

BeritJanssen:
Thanks for the comments, and also for suggestions to manage the state in a more efficient way. Since there is *some* urgency with fixing the bug (which, indeed, we didn't even log an issue yesterday, I was so taken aback that it happened). That being said, I agree the logic for determining the position into the file from where playback should resume should be factored out, and that a unit test will actually help us to spot problems with this code in the future.

albertas-jn (approved):

Issue #598, opened 2023-11-07
Fix React warnings


BeritJanssen:
When starting up the React dev server, the compiler will warn about problems in the code - double instead of triple equation signs and missing dependencies for `useEffect` hooks mostly. I always try to fix them if I see them, but with changes they sometimes creep back in...

drikusroor:
Could you elaborate on which React warnings we are talking about here? 😄

PR #660, opened 2023-12-13
Draft: Fixed: React lint warnings

drikusroor:
#598 

@BeritJanssen the remaining linting errors are all `react-hooks/exhaustive-deps`, e.g. "missing" or "incorrect" dependencies. Now, `eslint` proposes fixes here, but I'm not completely aware of what happens and should happen exactly so I'm afraid to break stuff, and I don't want to risk that over some linting warnings. Perhaps you can look into these or we can look at them together in the future?


frontend/src/components/Final/Final.js
```diff
@@ -46,10 +45,6 @@ const Final= ({ experiment, participant, session, score, final_text, action_text
         };
     }, [score, showScore]);
 
-    const finalize = () => {
-        finalizeSession({ session, participant });
-    }
-
```
BeritJanssen, line 52: Was this removed by esLint? The `finalizeSession` logic certainly shouldn't go. It makes sure the backend receives the signal to write to the database that the session is finished, and its end time. (So we can easily find sessions in which participants completed the whole experiment.)

BeritJanssen:
Thanks for configuring the linter. I think that indeed we should review the missing dependencies for the hooks manually. #648 should fix some of those warnings already.

BeritJanssen (changes requested):
I think we can merge some of these adjustments by the linter already, and take care of the missing / incorrect dependencies in the react hooks later. Perhaps as a team we should review our linter settings, since I'm not sure everyone has one configured.

But please undo the change removing `finalizeSession` from `API.js` and `Final.js`. Indeed, I see that `finalize()` is never called, while it should probably be placed into the `useEffect` hook as a cleanup function.

frontend/src/components/Final/Final.js
drikusroor: Well, `eslint` warned me about it because it was an unused function and therefore an unused import. The `finalize` function never called anywhere. I don't know if that's always been the case?

drikusroor:
> I think we can merge some of these adjustments by the linter already, and take care of the missing / incorrect dependencies in the react hooks later. Perhaps as a team we should review our linter settings, since I'm not sure everyone has one configured.
> 
> But please undo the change removing `finalizeSession` from `API.js` and `Final.js`. Indeed, I see that `finalize()` is never called, while it should probably be placed into the `useEffect` hook as a cleanup function.

I've looked it up and you have updated the call to `finalize` on June 20 here and made the `div` into a `a` element: https://github.com/Amsterdam-Music-Lab/MUSCLE/commit/1cc4ac45c1253761beefc5da76e2b4184e07107a#diff-dd2727c4ba071122a678bb23e7b9b2833fd33ab5995321f1df7d3dc4f4523d33

I think then it got removed after merging `develop` into a feature branch while resolving conflicts and so on:

https://github.com/Amsterdam-Music-Lab/MUSCLE/commit/1cc4ac45c1253761beefc5da76e2b4184e07107a#diff-dd2727c4ba071122a678bb23e7b9b2833fd33ab5995321f1df7d3dc4f4523d33

Apart from that, I don't know if it ever behaved like it should have? I can imagine when you click a link that also has a `onClick` handler, race conditions might cause the `onClick` never to get executed or finished as it already navigates towards a new page.

drikusroor:
Merging this and addressed the finalize session not being called as a seperate issue in #676 !

Issue #293, opened 2022-12-06
Add theming


jaburgoyne:
The lowest-hanging fruit seems to be to allow other labs to change the fonts (display and body text), the default colour scheme (cf. the AML style guide), and the favicons

drikusroor:
As discussed last Tuesday, I think we concluded it would be nice if we could configure a theme through the Django admin interface. Preferably the:

1. Font
2. Logo
3. Background image
4. Design token colors (set primary color, secondary color, success color, warning color, etc.)

A Dark theme would be 2nd priority.

@jaburgoyne Could you update the PR title & description to something more concise, with clearer boundaries and a well-defined scope? Perhaps by using a User Story Format (example below)?

> As a MUSCLE admin user I would like to be able to configure the font, logo, background image and main design token colors (primary, success, positive, negative, secondary, info) so I can easily customize the MUSCLE application's design for my organization.

jaburgoyne:
As a lab leader implementing the MUSCLE infrastructure at my own laboratory, I would like to be able to configure the fonts (heading, body, and perhaps code), logo, background image, and utility-class colours (primary, secondary, etc.) so that my lab's experiments have a consistent local look.

jaburgoyne:
Lower priority: As an experiment designer using text in non-Latin scripts, I may want to be able to configure the fonts (heading, body, and perhaps code) to support internationalisation for a specific experiment only.

jaburgoyne:
Still lower priority: As an experiment designer, I may want to be able to configure the background image and utility-class colours specifically for an individual experiment.

jaburgoyne:
Lowest priority: As a lab manager running an experiment on behalf of an external party, I may want to be able to configure the logo for an individual experiment.

jaburgoyne:
The utility classes for colours that are implemented in Bootstrap (the world I'm most familiar with) would be the following:

- [ ] fg
- [ ] bg
- [ ] primary
- [ ] secondary
- [ ] success
- [ ] info
- [ ] warning
- [ ] danger
- [ ] light
- [ ] dark

We don't necessarily need to support all of these (and could also support 'normal' colour names in addition), but this may still help to contextualise the issue.

BeritJanssen:
The backend `admin_interface` app now provides the option to override colours. Perhaps we need to rename them in terms of the bootstrap classes?

jaburgoyne:
Great idea! Let's do it

PR #666, opened 2023-12-17
Draft: POC: Configure theme colors in backend

drikusroor:
This PR contains a POC (proof of concept) for we _could_ approach the feature of a configurable theme in Django's backend. Basically it's a new model (`ThemeConfig`), with admin form pages (`/admin/theme/themeconfigs`), and an output page (`/theme/theme.css`).

## The `ThemeConfig` model:

```python
class ThemeConfig(models.Model):
    css_variables = models.TextField()  # Store all hard-coded variables as text
    additional_variables = models.TextField(blank=True)
    global_css = models.TextField(blank=True)
```

- The `css_variables` column contains css variables (key + value, e.g. `--teal: #1b949d;`) stored as json.
- I've added `additional_variables` so that people can easily add custom values if needed. Maybe we don't need this column.
- With `global_css`, the user can override anything in the css of the frontend application by (if it wants to) just doing stuff like
```css
body {
  background: red;
}
```

## The form

The form allows you to edit the pre-defined css variables (taken from the original `variables.scss` from the frontend.

First, it allows you to set the colors of the "color" css variables (teal, green, etc.). Then, you can use those colors in the "design token" css variables (primary, secondary, etc.), or pick a custom color for a design token.

The `additional_variables` and `global_css` are just free form textarea.

There is some validation on the `css_variables` that checks whether there are valid hex colors and valid variable names. The free form textareas don't have strict validation.

afbeelding

## The output

The output theme file, which you can load through `/theme/theme.css`, has the following structure:

```css
/* Theme configuration */

:root {

  /* Color & Design Token Configuration */
  --teal: #1b949d;
  --yellow: #cdbf1f;
  --pink: #ff3eed;
  --red: #ff0009;
  --blue: #0007de;
  --green: #36f424;
  --indigo: #00fcff;
  --gray: #aaaaaa;
  --gray_900: #4b4b4b;
  --black: #000000;
  --primary: var(--teal);
  --success: var(--green);
  --positive: var(--green);
  --negative: var(--red);
  --secondary: var(--yellow);
  --info: var(--gray);

  /* Additional Variables */
  --aqua: #00ffff;
}

/* Global CSS */
{ background-color: var(--aqua); }
```

We can then import this stylesheet into the React application and use the css variables that are initialized and assigned. As I was working on this, I realized that the scss configuration that we currently have is not fully compatible with using css variables as we use some scss functions under the hood in build time (`darken`, `lighten`, making colors transparent), which we cannot do run-time. So we have to think about providing these colors to the frontend. There are several options:

- Use a python library like `colour` to auto-generate darker, lighter and semi-transparent versions of all the colors in the theme file. This will then generate new colors under the hood for `--teal-light` or `--negative-dark`, etc. If we generate these during the submit of a theme instead of during the fetch of `/theme/theme.css`, we _shouldn't_ run into performance issues. We can also think about caching this theme file in memory or something. It's not big and won't change often probably.
- Use javascript functions at the initialization that generate the darker, lighter, transparenter versions that we need and inject them into the html. Basically the frontend version of the previous suggestion.
- Use css filters and opacity to simulate the same effect of making colors darker and lighter and more transparent. This is pretty easy, but the result might not be the nicest.


BeritJanssen:
Nice! I assume this is aimed towards #293 ? In my vision, we'd just have different stylesheets live next to each other in the frontend (and control usage through env variables), or ask other groups to fork and override. An advantage of that approach would be that the source of truth of the color scheme is always to be found in the repository.

Setting colours through the database would certainly be nice, but I wonder how much to "grow" this feature, and what its limits are. We also want other groups using the system to be able to choose different fonts, font sizes, images, favicons... Would it for instance be feasible that some user uploaded their own font and defined usage of it, and then collect that all into one custom css before frontend build? And can we cater for all potential elements of the current layout that future users may want to override? In my opinion, it would be "all or nothing" - if not every style element can be overridden in this way, it would probably be confusing to be able to do it with *some*.

backend/theme/static/theme_admin.js
```diff
@@ -0,0 +1,103 @@
...
+            // if the value is not empty, add it to the cssVariables array
+            if (value !== '') {
+                cssVariables.push(`--${name}: ${value};`);
+            }
+        });
+
+        // inject the cssVariables into the document
+        const cssVariablesString = cssVariables.join(' ');
+        const style = document.createElement('style');
+        style.innerHTML = `:root { ${cssVariablesString} }`;
```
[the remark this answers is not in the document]
drikusroor: Wow, this is neat. Good bot!

backend/experiment/migrations/0019_alter_experimentseries_options.py
BeritJanssen, line 1: You need to redo the migrations if you pick this branch up again.

backend/theme/migrations/0002_auto_20240216_1142.py
BeritJanssen, line 1: Please give the migrations more descriptive names. Perhaps even go back to _initial (`manage.py migrate theme 0001`) and "rewrite" the migration history so there's less `AlterField` operations.

drikusroor:
Closing this POC PR as the issue is split in two issues now:
- Configurable theme (logo, background, font) per experiment (#805)
- Configurable colors/design tokens per experiment (issue to be created still)

PR #668, opened 2023-12-18
Feature/tele tunes

BeritJanssen:
Add a rules file and introduce more fine-grained support over response times in different phases of the Hooked game. The TeleTunes clips last 30 seconds instead of 45 seconds (TV theme songs being usually shorter than the average pop song), so it was necessary to reduce the response time so as not to "run out" of audio before the end of a round.


albertas-jn (approved):

backend/experiment/actions/wrappers.py
```diff
@@ -81,6 +87,7 @@ def song_sync(session, section, title, response_time=15, play_method='BUFFER'):
         title=title
     )
     continuation_correctness = random.randint(0, 1) == 1
+    trial_config['resonse_time'] = sync_time
```
drikusroor, line 90: @BeritJanssen typo :-)

PR #672, opened 2023-12-19
make non essential elements optional

### LINKED ISSUE #605: Remove non-essential elements of Congo Branch of TuneTwins
In the Congo branch, the functionality of the TuneTwins games needs to be removed to only essential functioning.

This includes

* Remove the equalizer animation 
* Remove the matching result animations
* Change the text feedback to be less prominent by only showing text in small corner of screen, in grey, replace Lucky match with "0", Correct match with "+1", incorrect match with "-1" , no global score is shown 
* Game begins only with entering text field with participant ID, then goes directly to game, which only plays 1x, no loop, items still randomized within game, but condition determines game as which game is covered in #604

Evert-R:
This PR enables you to configure elements of matching pairs from the top if the rules file.

These are the defaults we used to until now:

`show_animations = True    
show_total_score = True    
show_score_message = True   
show_turn_feedback = False
histogram_bars = 5` 

(set histogram bars to zero for no histogram)

resolves #605


davidjohnbaker1:
This looks good from my end! I am sure Henkjan will probably have small requests here or there on it once we see it live, but this functionality is exactly what we asked for. Thanks so much!

davidjohnbaker1 (approved):
Looks good to me!

BeritJanssen (changes requested):
Though I think it's need that this PR gives such fine-grained control over the feedback and animations, I believe for now one control for show_animations, and one for show_score_feedback will do. Thinking ahead of #724 , this one can be HIDDEN / top / bottom_right. Will keep the options in the frontend, but adjust the backend options when merging in develop. I'd like to avoid a plethora of settings which may be hard to understand for future developers.

PR #673, opened 2023-12-19
Fixed: Fix typo

drikusroor:


BeritJanssen:
I pushed this fix directly to develop. 😕 PR would've been better, but it's fixed, so this can be closed. Thanks for catching that! 💯

Issue #628, opened 2023-12-01
[TMS] Sound not working


jaburgoyne:
Vanderbilt has just reported that *no* fragments are playing sound in That's My Song. I cannot reproduce the issue myself but have not tested fully with VPNs, etc. Their location is Nashville, Tennessee, and they access the experiment using iPads.

They would like to be able to run a session again today: is it possible to have a look?

tanguy-rdm:
Hi all !

Thanks a lot for being so responsive in trying to help! Indeed, I reviewed your troubleshooting steps and those seem to be a great starting point !!

On Vanderbilt side, the team realized that the bug might be associated with an incompatibility between the game and the iOS version of the iPAD used to run the game. From what we know, sound was failing when our iOS was around 16.6 (i.e., prior to update). Facing the sound issues, we decided to update our iPads to the most recent available version (17.1.2.) and now the game seems to work just fine !

With this in mind, I have two remarks:
(1) Do you think it is plausible that the game might be only compatible with specific iOS versions? 
(2) If this is the case, would there be an option for the game's programmers to adjust this and extend the compatibility to a wider range of iOS versions? I am asking this specifically because, in the long run, we (as a lab) might not be able to control which iOS version participants will use to run the game. 

Again thanks for the help !

BeritJanssen:
Thanks for the remark about iOS versions. Indeed, I just found this post which seems to clarify the issue, and why the update fixes this:
https://devcodef1.com/news/1068842/fixing-ios-17-web-audio-issue
We are using webAudio, and there's been a faulty implementation for that in iOS. I'll investigate in how far we can implement the workaround suggested in the post to assure participants with older iOS versions don't face issues.

Evert-R:
I have made a pull request with the solution offered in that post.
The PR can be discarded if it doesn't fix the ios17 issue.

Evert-R:
This soulution is now running on the acceptance server:
https://acc.amsterdammusiclab.nl/thats_my_song

BeritJanssen:
The solution for closing and opening the audioContext does not seem to fix the problem across systems. Alternatively, we could do the following:
- determine the user's system (using `window.navigator.userAgent`)
- if the system contains "OS {majorversion_minorversion} like Mac OS X" and `majorversion<17.0`, disable webAudio

jaburgoyne:
[...]

PR #674, opened 2023-12-19
Close and recreate audiocontext for every play

Evert-R:
This PR closes and recreates the audiocontext of webaudio after each played sound.
https://devcodef1.com/news/1068842/fixing-ios-17-web-audio-issue

Hopefully this resolves issue #628
https://github.com/Amsterdam-Music-Lab/MUSCLE/issues/628

If not, this PR can be discarded, for efficiency reasons. 
(I cannot test this myself on IOS17)

### DIFF
--- a/frontend/src/util/webAudio.js
+++ b/frontend/src/util/webAudio.js
@@ -107,6 +107,10 @@ export const stopBuffer = () => {
 
 // play buffer by section.id
 export const playBuffer = (id) => {
+    // temporary solution for ios17
+    // https://devcodef1.com/news/1068842/fixing-ios-17-web-audio-issue
+    audioContext.close();
+    audioContext = new (window.AudioContext || window.webkitAudioContext)();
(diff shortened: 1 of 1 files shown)


BeritJanssen (approved):
This should work!

frontend/src/util/webAudio.js
```diff
@@ -107,6 +107,11 @@ export const stopBuffer = () => {
 
 // play buffer by section.id
 export const playBuffer = (id) => {
+    // temporary solution for ios17
+    // https://devcodef1.com/news/1068842/fixing-ios-17-web-audio-issue
+    audioContext.close();
```
drikusroor, line 112: Will this throw an error if `audioContext` isn't (yet) defined, or is that impossible to happen?

BeritJanssen:
It turns out that the error persists on some iOS versions even when recreating the audioContext every time, so I think we shouldn't merge this.

PR #678, opened 2023-12-20
Feature/teletunes questions

BeritJanssen:
I thought that assigning questions through the admin would work for all experiments now, but it turns out it only works for those which use the `experiment.rules.base.get_questionnaire` function. Most Hooked experiments get a single question after each round, and use the `experiment.rules.base.get_single_question` function instead. Since we don't want to fully randomize the questions here, but first get demographic, then musicgens questions, I reasoned it would be faster for now to hard-code the questions "old style", and add an issue for the `get_single_question` function ( #679 ), as we still need to figure a way to do "randomization in blocks" through the admin interface also.

The branch also adds a test for the playhead shift in song_sync rounds, checking that indeed, the playhead advances by `silence_time` if the track should continue where it left off, and otherwise `silcence_time` + a random jitter between min and max values.

### DIFF
--- a/backend/experiment/actions/utils.py
+++ b/backend/experiment/actions/utils.py
@@ -1,4 +1,5 @@
 from os.path import join
+import random
 
 from django.conf import settings
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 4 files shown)


drikusroor (approved):
LGTM. And yes, let's hard-code for now and fix #679 later. This one is supposed to go live soon right?

PR #680, opened 2023-12-20
check correctness of get_preferred_songs function

BeritJanssen:
This branch adds a test to check that the information about the "most liked" songs presented to a participant in the musical preferences experiment are indeed those with the *lowest* score in the database. As tests should, this still showed some flaws in the logic, which - as far as the test can tell - are now fixed.

### DIFF
--- a/backend/experiment/rules/musical_preferences.py
+++ b/backend/experiment/rules/musical_preferences.py
@@ -284,11 +284,11 @@ class MusicalPreferences(Base):
         return info
 
     def get_preferred_songs(self, result_set, n=5):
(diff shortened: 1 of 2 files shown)


drikusroor (approved):
I think I approve, but I'm not sure if I completely understand. How does this test the correctness? Are you first saving the songs in random/inversed score order and do you then check if the lowest scored song is the first song in the list when you fetch it?

BeritJanssen:
The results are created in the order of sections (ordered alphabetically), with decreasing scores. You are right that the test might either document this, or (better) explicitly order the sections before result creation.

PR #688, opened 2023-12-23
Config: Migrate from CRA to Vite

drikusroor:
This pull request includes updates to the Vite configuration file and environment variables. It also updates the environment variables used in the application.


drikusroor:
> Why?

A very good question! Apparently I have forgotten to provide a proper explanation in the PR description 😅 So yes, I would hereby like to propose to think about migrating from CRA (Create React App) to Vite. CRA - which runs on Webpack under the hood - is a [deprecated / dead / no longer maintained project](https://github.com/facebook/create-react-app/issues/13072) and the React team [no longer recommends it](https://react.dev/learn/start-a-new-react-project) (CRA used to be on that page). They do recommend things like Next.js or Remix, which we could also consider, but those projects require a completely different setup (with server-side rendering). For our situation - serving a build of static assets - [Vite is probably the best option](https://www.reddit.com/r/reactjs/comments/12x6k9x/it_looks_like_createreactapp_is_dead_what_should/). I have used Vite in React projects before and I have also migrated CRA projects to Vite before. Apart from the fact that CRA is no longer maintained, Vite also has benefits:
- It's way faster (hence the name, Vite is fast in French)
- It has typescript built-in, making it easier to migrate to typescript if we want to

For the rest not much changes. The application works more or less the same. There are some files and folders that need to be in a different location and the environment variables are no longer to be prefixed with `REACT_APP_`, but with `VITE_`.

.env-github-actions
```diff
@@ -7,12 +7,14 @@ AML_SECRET_KEY=amlsecretkey
...
+VITE_STRICT=false
+VITE_API_ROOT=http://localhost:8000
+VITE_EXPERIMENT_SLUG=fe
+VITE_AML_HOME=https://www.amsterdammusiclab.nl
+VITE_HTML_PAGE_TITLE=Amsterdam Music Lab Experiment
+VITE_HTML_OG_DESCRIPTION=Listening experiments from the Amsterdam Music Lab. Test your musical knowledge and skills in engaging citizen-science experiments.
+VITE_HTML_OG_IMAGE=https://app.amsterdammusiclab.nl/images/og-hooked.jpg
+VITE_HTML_OG_TITLE=Amsterdam Music Lab Experiment
+VITE_HTML_OG_URL=https://app.amsterdammusiclab.nl
+VITE_HTML_BODY_CLASS=
```
BeritJanssen, line 20: I think renaming all these variables to `VITE` makes the purpose of them less clear. The `REACT_APP` here does not refer so much to the compilation procedure as to the role of the variables, which are about controlling the behaviour of the REACT frontend. Perhaps we could change these to `CLIENT` or `FRONTEND` instead, so that they are more descriptive.

frontend/package.json
```diff
@@ -76,21 +75,25 @@
...
     "@storybook/blocks": "7.6.6",
+    "@storybook/builder-vite": "^7.6.6",
     "@storybook/preset-create-react-app": "7.6.6",
     "@storybook/react": "7.6.6",
     "@storybook/react-webpack5": "7.6.6",
     "@storybook/testing-library": "0.2.2",
-    "@testing-library/jest-dom": "^6.1.5",
     "@testing-library/react": "^14.1.2",
     "@testing-library/user-event": "^14.5.1",
+    "@vitest/coverage-v8": "^1.2.1",
```
BeritJanssen, line 85: This would be good to move to devDependencies, along with other testing dependencies.

frontend/src/util/timer.js
```diff
@@ -20,16 +20,17 @@ export const Timer = ({
...
 
         // callback after interval
         if (time - lastTime > interval) {
             lastTime = time;
 
             // tick
-            onTick && onTick(time, delta);
+            if (onTick) {
+                onTick(time, delta);
+            }
```
BeritJanssen, line 33: I thought that above was shorthand for below?

frontend/vite-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="vite/client" />
```
BeritJanssen, line 1: Why do we need this file, which seems to only contain a comment?

BeritJanssen (changes requested):
Nice! Still have to see how it plays out in practice. Please do change the `VITE_...` variable names to something more descriptive, see my comments. Users need to at least have an idea of what these variables are for.

.env-github-actions
drikusroor: > I think renaming all these variables to `VITE` makes the purpose of them less clear. The `REACT_APP` here does not refer so much to the compilation procedure as to the role of the variables, which are about controlling the behaviour of the REACT frontend. Perhaps we could change these to `CLIENT` or `FRONTEND` instead, so that they are more descriptive.

Actually, the `REACT_APP` prefix is a [requirement for adding environment variables](https://create-react-app.dev/docs/adding-custom-environment-variables/) when using CRA (Create React App). [The same thing is true for VITE](https://vitejs.dev/guide/env-and-mode#env-files).

A possible way to avoid this is to use the variable names we want to use in the `.env` file, and then have the `docker-compose.yaml` file map them like this:

```
# .env file

CLIENT_API_ROOT=https://example.com
```

```
# docker-compose.yaml

client:
        environment:
          - VITE_API_ROOT=${CLIENT_API_ROOT}
```

Would you prefer such a solution or just use `VITE_`?

frontend/package.json
drikusroor: I think it is there already ;-)

frontend/vite-env.d.ts
drikusroor: I think it's so that we can add type definitions for environment variables. See also [here](https://vitejs.dev/guide/env-and-mode#intellisense-for-typescript)

frontend/src/util/timer.js
drikusroor: It is! Reverted.

BeritJanssen (changes requested):
Can you please still change the variable names here? Either leave it as `REACT` (since we're still serving / building React here), or change to `FRONTEND`. The last setting in this list (`STRICT`) is arguably a React setting, right?

drikusroor:
> Can you please still change the variable names here? Either leave it as `REACT` (since we're still serving / building React here), or change to `FRONTEND`. The last setting in this list (`STRICT`) is arguably a React setting, right?

That's not possible. CRA requires `REACT_APP` as prefix, but Vite requires `VITE_` as prefix for environment variables.

BeritJanssen:
All the variable names were chosen and set by ourselves. So if it says "process.env.REACT_APP..." in the code somewhere, that comes from this environment variable. This doesn't have anything to do with CRA, as far as I'm aware.

drikusroor:
> ~All the variable names were chosen and set by ourselves. So if it says "process.env.REACT_APP..." in the code somewhere, that comes from this environment variable. This doesn't have anything to do with CRA, as far as I'm aware.~ My bad: turns out that indeed CRA auto-ignores all environment variables not prefixed with `REACT_APP`, and then Vite must have similar behaviour, but for `VITE`. Would our internal variables perhaps still deserve a more descriptive name? So that we have in the `docker-compose`: VITE_SOMETHING=${FRONTEND_SOMETHING}? Or do you think this might be even more confusing?

Yeah, we can definitely do that, especially since we have to update our environment variables anyway 👍

PR #689, opened 2023-12-26
Fixed: finalizeSession bug in Final component

### LINKED ISSUE #676: Finalize session does not get called
**Describe the bug**
The so called `finalizeSession` endpoint does not get called at the end of a session in the frontend (anymore). Calling this endpoint tells the backend that the session is over. It used to be called, but somewhere during development the call got removed. See also [this PR comment](https://github.com/Amsterdam-Music-Lab/MUSCLE/pull/660#issuecomment-1862573060)

**To Reproduce**
1. Start any experiment
2. Keep the developers tools open in your browser on the network tab
3. Finish it
4. Check the network tab and confirm that `/session/{id}/finalize` does not get called

**Expected behavior**
The finalize session endpoint _does_ get called when the session is over

**Screenshots**
`n/a`

`n/a`

**Smartphone (please complete the following information):**
`n/a`

**Additional context**
Related to #598 and #660

drikusroor:
This pull request fixes a bug in the Final component where the finalizeSession function was not being called correctly. The bug caused the finalizeSession function to not be called when the Final component mounted. This pull request adds the missing call to finalizeSession in the useEffect hook.

Resolves #676

### DIFF
--- a/frontend/src/components/Final/Final.test.js
+++ b/frontend/src/components/Final/Final.test.js
@@ -92,9 +92,10 @@ describe('Final Component', () => {
 
         expect(profileLink).toBeInTheDocument();
 
+        expect(history.location.pathname).toBe('/');
+
         fireEvent.click(profileLink)
 
-        expect(history.length).toBe(2);
         expect(history.location.pathname).toBe('/profile');
 
     });
--- a/frontend/src/components/Final/Final.test.js
+++ b/frontend/src/components/Final/Final.test.js
@@ -1,0 +1,115 @@
+import React from 'react';
+import { render, screen, fireEvent, waitFor } from '@testing-library/react';
+import { BrowserRouter, Router } from 'react-router-dom';
+import { createMemoryHistory } from 'history'
+
+import Final from './Final'; // Adjust the import path as necessary
+
+
+jest.mock('../../API', () => ({
+    finalizeSession: jest.fn(),
+}));
+
+jest.mock('../../config', () => ({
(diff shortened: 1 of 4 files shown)


frontend/src/components/Final/Final.test.js
```diff
@@ -0,0 +1,115 @@
...
+
+        const history = createMemoryHistory();
+
+        const mockActionTexts = {
+            all_experiments: 'All Experiments',
+            profile: 'Profile',
+        };
+
+        render(
+            <Router history={history}>
```
BeritJanssen, line 83: Why do you wrap the `Final` component in `Router` with history here?

frontend/src/components/Final/Final.test.js
```diff
@@ -0,0 +1,115 @@
...
+        );
+
+        const profileLink = screen.getByTestId('profile-link');
+
+        expect(profileLink).toBeInTheDocument();
+
+        fireEvent.click(profileLink)
+
+        expect(history.length).toBe(2);
+        expect(history.location.pathname).toBe('/profile');
```
BeritJanssen, line 98: Ah, I see: you check whether the profile link correctly forwards to the desired location. Very nice! The test stumbles over the import from `history`, however - does this need to be added to the dependencies in package.json?

BeritJanssen (changes requested):
The test stumbles over not being able to import the `history` library. Could also be due to a version mismatch?

PR #690, opened 2023-12-27
Draft: Config: Add ESLint configuration for frontend

drikusroor:
This draft pull request adds a new ESLint configuration for the frontend code. It includes configuring VS Code auto formatting based on ESLint rules and adding the .eslintrc.js file to the Dockerfile. Additionally, it fixes linting issues in the codebase.

You should see this as the entry point for a discussion about how we want to setup proper linting (for JS & React)

Some questions about format and linting preferences I would have for you already:
- Do you prefer indentation of 2 or 4 spaces?
- How strict do you want the linting to be?
- Can we leave unused variables or undefined variables in the code?
- Should the last item of an array have a dangling comma?
- Double or single quotes?
- Should props be typed and if so, how strict should we be on it?


drikusroor:
I'm closing this PR due to the lack of response and because some form (a bit more minimal) of eslint will eventually be added in #688

PR #691, opened 2023-12-27
Added: Add upload and export functionality to PlaylistAdmin

### LINKED ISSUE #591: Upload / download playlist csvs
Instead of copying / pasting csv files, it may be interesting to add buttons to upload / download .csv files to / from the database.

drikusroor:
This pull request adds the ability to upload CSV files and export playlists to CSV and JSON formats in the PlaylistAdmin section.

Resolves #591

### DIFF
--- a/backend/section/admin.py
+++ b/backend/section/admin.py
@@ -26,6 +26,7 @@ class SectionAdmin(admin.ModelAdmin):
     # Prevent large inner join
     list_select_related = ()
 
+
 admin.site.register(Section, SectionAdmin)
 
 # @admin.register(Playlist)
--- a/backend/section/admin.py
+++ b/backend/section/admin.py
@@ -11,7 +11,7 @@ from django.utils.translation import gettext_lazy as _
 import audioread
 
 from .models import Section, Playlist, Song
-from .forms import AddSections
+from .forms import AddSections, PlaylistAdminForm
 
 
 class SectionAdmin(admin.ModelAdmin):
@@ -33,8 +33,8 @@ admin.site.register(Section, SectionAdmin)
 
 class SongAdmin(admin.ModelAdmin):
     list_per_page = 50
-    list_display = ('artist', 'name')    
-    search_fields = ['artist', 'name']   
+    list_display = ('artist', 'name')
+    search_fields = ['artist', 'name']
 
     # Prevent large inner join
     list_select_related = ()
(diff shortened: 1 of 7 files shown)


BeritJanssen (changes requested):
Nice! I tested it and noticed the following:
- the csv export has two leading columns and an extra column after the filename, e.g.:
2020,2020,default,1_tempo,0.0,15.856326530612245,BAT/1_tempo.wav,[],0,0,0

This means the file cannot be uploaded / pasted into the csv field again.

It would also be nice to have the download button closer to the text field. Next to it perhaps? Or directly below?
Screenshot 2024-01-16 at 17 47 08

It's probably better to get rid of the JSON export here - I don't think it would be used by anyone. The desired functionality is to be able to import and export playlist csvs.

drikusroor:
> Nice! I tested it and noticed the following:
> 
> * the csv export has two leading columns and an extra column after the filename, e.g.:
>   2020,2020,default,1_tempo,0.0,15.856326530612245,BAT/1_tempo.wav,[],0,0,0
> 
> This means the file cannot be uploaded / pasted into the csv field again.

The first two columns are the ids and the pks. They can probably be removed I think? The other field used to be the play count. However, in the original input description (when importing csv data) it speaks about `restricted_to_nl`, which is a field that doesn't exist anymore. When I look at the Section model, I see the fields `play_count` and `code` after the `filename`. Which one should be exported?

> It would also be nice to have the download button closer to the text field. Next to it perhaps? Or directly below? Screenshot 2024-01-16 at 17 47 08

I tried to do this and found it quite tedious. Do you know an easy way to inject that button there? 🤔 

> It's probably better to get rid of the JSON export here - I don't think it would be used by anyone. The desired functionality is to be able to import and export playlist csvs.

👍 Removed it

BeritJanssen:
Ah, looks like we also need to update the import from csv. No, we don't want to export section ids here, but export exactly the same format csv that the admin interface expects in the TextField (see the small helper text under the TextField in the screenshot). I wasn't aware that `restricted_to_nl` was removed. The last columns (according to the help line under the csv input field) should be "restricted_to_nl", "tag" and "group". We should then also check the `process_csv` function again.

I think we might indeed need to write a custom html form in order to achieve the move of the button, now that I think of it. 🤔 The button is visible enough where it is, so it's probably not worth it.

BeritJanssen:
I still went after what happened to the `restrict_to_nl` field on `Session`. It is actually still there, but now as a field of `Song`, as a `JSONField`, with the reasoning we could add in lists of country codes in which playback was *allowed*. @jaburgoyne actually indicated that we don't need this restriction at all anymore, so I'll file an issue to remove `Song.restricted`, which will have consequences both on this feature, and on `playlist.update_sections`.

PR #692, opened 2023-12-28
Added: Add FloatingActionButton and UserFeedback components

### LINKED ISSUE #264: Add Feedback button
Add feedback button with 2/3 questions and text field - comparable to the feedback link on  all UvA sites. Could be a generic feature for all AML experiments?
uva

drikusroor:
This pull request adds the FloatingActionButton and UserFeedback components to the project. It also includes various style updates and unit tests.

At the experiment's `feedback_info` you can supply a parameter to turn it on or off for a given experiment by setting `show_float_button` to `True`.

For example:

```py
def feedback_info(self):
        info = super().feedback_info()
        info['show_float_button'] = True
        return info
```

It currently looks like this:

https://github.com/Amsterdam-Music-Lab/MUSCLE/assets/8208970/1e199b58-207f-4740-b16b-77559a4b8435

## Flow
- User sees floating feedback button in the center right on it and clicks it
- Floating Action Button opens with an animation
- User enters feedback in text input
- User clicks submit
- If successful, a "Thank you..." message is shown
- User can close the Floating Action Button again by clicking the "X" close icon button, or by clicking somewhere else in the screen

## Questions for the reviewer(s)
* Should we enable or disable it by default for all experiments?
* Should it look _exactly_ like the existing feedback button on the UvA site in the issue?
* If not, is there anything else you would like to change to the appearance or the behavior of the components? 

Resolves #264

### DIFF
--- a/backend/experiment/rules/gold_msi.py
+++ b/backend/experiment/rules/gold_msi.py
@@ -43,5 +43,4 @@ class GoldMSI(Base):
 
     def feedback_info(self):
         info = super().feedback_info()
-        info['show_float_button'] = True
         return info
--- a/backend/experiment/rules/gold_msi.py
+++ b/backend/experiment/rules/gold_msi.py
@@ -40,3 +40,8 @@ class GoldMSI(Base):
             return questions
         else:
             return final_action_with_optional_button(session, '', request_session)
+
+    def feedback_info(self):
+        info = super().feedback_info()
+        info['show_float_button'] = True
+        return info
--- a/frontend/src/components/FloatingActionButton/FloatingActionButton.js
+++ b/frontend/src/components/FloatingActionButton/FloatingActionButton.js
@@ -21,27 +21,39 @@ const FloatingActionButton = ({
     }
 
     return (<>
-        <div className={
-            classNames("floating-action-button",
-                getPositionClassNames(position),
-                expanded && 'floating-action-button--expanded',
-                className
-            )}>
+        <div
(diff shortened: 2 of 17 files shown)


hhoning (approved):
looks great to me!

backend/experiment/rules/gold_msi.py
```diff
@@ -40,3 +40,8 @@ def next_round(self, session, request_session=None):
             return questions
         else:
             return final_action_with_optional_button(session, '', request_session)
+
+    def feedback_info(self):
+        info = super().feedback_info()
+        info['show_float_button'] = True
+        return info
```
BeritJanssen, line 47: Hmm, we have existing experiment series that have the `gold_msi` rules included. It would be weird to include the floating button there but not in the other experiments. So perhaps remove this from this specific experiment.

frontend/src/components/FloatingActionButton/FloatingActionButton.js
```diff
@@ -0,0 +1,67 @@
...
+
+    const [expanded, setExpanded] = React.useState(false);
+
+    /**
+     * @param {string} position
+     * @returns {string}
+     */
+    const getPositionClassNames = (position) => {
+        const [vertical, horizontal] = position.split('-');
+        return `floating-action-button--${vertical} ${horizontal ? `floating-action-button--${horizontal}` : ''}`;
```
BeritJanssen, line 20: Can you document what this function does?

Issue #683, opened 2023-12-21
Reorganize color schemes in frontend


BeritJanssen:
Currently, we don't have an easy mechanism to set styles of playback buttons. It would be good to add different colour schemes easily.

Originally, colours were all set individually through the backend, which led to rather lengthy list of hex strings, which were difficult to read and reuse. This led to my idea to only send style classes (what's now "neutral", "boolean", etc.) and apply them through a whole `Trial`, so it would affect all the buttons in the associated form and playback. This had to be revisited when realizing that Trial can hold multiple questions, which may need different styling (i.e., the Musical Preferenes experiment). So the style is now applied *per question*, and the style classes are not send or set for the playback buttons at all. What makes the Playbuttons slightly more complex is that they use mixins for styling, as the shadow around them is also a shade of their primary colour.

This issue may be more of a "design" one: how to make sure colours can be set with a small effort, in a file that's easy to find, and so that the choices we have now set for different colour schemes can be overridden easily by other teams using the infrastructure?

drikusroor:
I've been thinking about and experimenting with this a bit in the last couple of days, and I've com to the conclusion that I think we first need an overview of the possible color schemes and variations for the components that we have or that we want to have.

But based on my interpretation of what you are describing I've had something in mind:

* We (the MUSCLE team) decide which color tokens (primary, secondary, warning, danger, etc.) and color schemes (boolean, neutral, boolean-inverted, neutral-inverted, etc.) exist. Users that clone and self-host this platform can optionally alter the colors in the color tokens and color schemes.
* The self-hosting user can select a default color scheme for the whole platform. (optional)
* The user who designs / programs an experiment can select a default color scheme for the experiment.
* The user who designs / programs an experiment can select a color scheme for every question / trial of an experiment, by supplying the color scheme class as a question / trial configuration parameter.
* For trials that use multiple components, I was thinking of using some form of nested configuration with a fallback to the parent class. In that way we can give the user full control over the color scheme to use per component.
* The trial component needs to take this into account and be able to pass nested color scheme configurations to its children.
* All components need to be able to handle the color schemes that we (the MUSCLE team) have defined and have a prop for this. Some components will probably need their own interpretation of a color scheme instead of simply relying on a globally defined button style mixin.

An example of how to approach this in pseudo-code:

## Backend / Experiment configuration

Global color scheme: `boolean`

```
Experiment A
  Question:
    text: What is your age?
    colorScheme:
[...]

PR #696, opened 2024-01-04
POC: Reorganize color schemes

drikusroor:
Related #683


backend/experiment/rules/gold_msi.py
```diff
@@ -27,7 +26,14 @@ def __init__(self):
         self.questions = MSI_F3_MUSICAL_TRAINING + demographics
 
     def first_round(self, experiment):
-        consent = Consent()
+        consent = Consent(
+            style=FrontendStyle(
+                FrontendStyle.BOOLEAN,
+                submit_button=FrontendStyle(
```
BeritJanssen, line 32: What is the reasoning here to assign the `SUCCESS` style here for the submit button? I'd assume it's already implicated in the `BOOLEAN` style at a higher level?

backend/experiment/actions/consent.py
```diff
@@ -20,8 +21,9 @@ class Consent(BaseAction):  # pylint: disable=too-few-public-methods
                 contentiones, vix ex maiorum denique! Lorem ipsum dolor sit \
                 amet, nec te atqui scribentur. Diam molestie posidonium te sit, \
                 ea sea expetenda suscipiantur contentiones."
-    
-    def __init__(self, text=default_text, title='Informed consent', confirm='I agree', deny='Stop'):
+
+    def __init__(self, text=default_text, title='Informed consent', confirm='I agree', deny='Stop', style=FrontendStyle()):
+        super(Consent, self).__init__(style)
```
BeritJanssen, line 26: You can just write `super().__init__(style)` here.

backend/experiment/actions/base_action.py
```diff
@@ -1,9 +1,16 @@
+from .frontend_style import FrontendStyle
+
 class BaseAction(object):
     ID = 'BASE'
+    style = FrontendStyle()
 
-    def __init__(self):
+    def __init__(self, style = FrontendStyle()):
```
BeritJanssen, line 7: If we pass the style here as an argument, there's no need to also set it as a class variable in line 5.

backend/experiment/actions/base_action.py
```diff
@@ -1,9 +1,16 @@
...
         pass
 
     def action(self):
-        self.view = self.ID
-        return self.__dict__
+        action_dict = self.__dict__
+        action_dict['view'] = self.ID
+        action_dict['style'] = self.style.to_dict()
+
+        return action_dict
```
BeritJanssen, line 16: This could become problematic, as subclasses of `BaseAction` may implement their own `action` function.  The same goes for `__init__` - `super` is not always called by the subclasses.

backend/experiment/tests/test_frontend_style.py
```diff
@@ -0,0 +1,64 @@
+import unittest
+
+from experiment.actions.frontend_style import FrontendStyle
+
+class TestFrontendStyle(unittest.TestCase):
+
+    def test_init_with_valid_root_style(self):
+        style = FrontendStyle(FrontendStyle.PRIMARY)
```
BeritJanssen, line 8: The initialization of `FrontendStyle` is somewhat awkward, I think - it now looks as if it's both a class and an enum.

BeritJanssen (changes requested):
I like it! But I'm not sure I understand the reasoning for introducing the `nested_styles` in the `FrontendStyle` class. I thought that every `action` may apply one style to its root element, and that controls the colours of the children. Was this introduced specifically because `Consent` does things differently? I also cannot foresee the use for `get_style` and `apply_style` from the current POC, as they are only used in the tests. Let's discuss IRL when we can.

backend/experiment/rules/gold_msi.py
drikusroor: That's just in case you would want to override it. The two buttons in the `Consent` form aren't styled using the button array component afaik.

backend/experiment/actions/base_action.py
```diff
@@ -1,9 +1,16 @@
+from .frontend_style import FrontendStyle
+
 class BaseAction(object):
     ID = 'BASE'
+    style = FrontendStyle()
```
drikusroor, line 5:
```suggestion
```

backend/experiment/actions/consent.py
drikusroor:
```suggestion
        super().__init__(style)
```

backend/experiment/tests/test_frontend_style.py
drikusroor: Yeah, I was struggling to find the best way to do it :-( Would you have any suggestions? Does python have enums?

backend/experiment/actions/base_action.py
drikusroor: Hmm, what would be a good solution to this then? Ideally I would like to have all actions serialise their style so that the frontend knows what to expect. Is there another way to enforce that?

backend/experiment/tests/test_frontend_style.py
BeritJanssen: Yes, it does! [https://docs.python.org/3/library/enum.html](https://docs.python.org/3/library/enum.html)

backend/experiment/rules/gold_msi.py
BeritJanssen: You're right, so I think we should refactor such that Consent reuses the button array component, and we can also set styles in the same way.

backend/experiment/actions/base_action.py
BeritJanssen: You'd probably need to call `super()` either on the `__init__` (and then do `style.to_dict()` in the init function) or the `action` function of the subclasses to be sure.

backend/experiment/tests/test_frontend_style.py
BeritJanssen: I think it would be neater to have a separate Enum with all the style options, and the class which is initialized with one of these options.

backend/experiment/actions/base_action.py
drikusroor: I think adding the `style.to_dict()` method in a custom implementation of the `action` method is the way to go. If it's ommitted, it will just mean no styles will be sent to the frontend, right? That shouldn't normally be a problem since every component has a (or not yet) default style.

backend/experiment/rules/gold_msi.py
drikusroor: I think that would be a good follow up ticket as it seems more like a new feature/refactor and I would like to keep that separate. However, I will remove the nested styles from the consent then in `gold_msi.py`.

BeritJanssen (changes requested):
I would still get rid of the changes to `Consent`, because I cannot foresee any other style than the boolean one being needed here. Also the `**nested_styles` option for `FrontendStyle` adds complexity that I don't think we need at this moment.

backend/experiment/actions/frontend_style.py
```diff
@@ -0,0 +1,66 @@
...
+    def is_valid(value):
+        return value in EFrontendStyle.__members__.values()
+
+
+class FrontendStyle:
+
+    VALID_STYLES = EFrontendStyle.__members__.values()
+
+    """
+    Initialize the FrontendStyle with a root style, and optionally nested styles.
```
BeritJanssen, line 27: doc string is outdated.

BeritJanssen (approved):

backend/experiment/actions/frontend_style.py
drikusroor:
```suggestion
    Initialize the FrontendStyle with a root style.
```

PR #705, opened 2024-01-09
only activate React StrictMode when Docker variable is True

BeritJanssen:
This fix is necessary when working with request-heavy experiments, such as MatchingPairs locally: we should work on a way to reduce the number of requests on rerender, but for the time being, `StrictMode` will be disabled in most cases.


drikusroor (approved):
LGTM. The frontend linting job is failing but I think that's because of Github issues: https://www.githubstatus.com/

Issue #655, opened 2023-12-12
Reduce "prop drilling"


BeritJanssen:
The current frontend setup passes data around from the `Experiment` component to deeply nested children. This is a hurdle both for future maintainers (looking up dependencies through multiple parents is no joy), and for reusability.

For state that doesn't change much, e.g., participant and session, we may move the state logic to a [useContext hook](https://react.dev/reference/react/useContext). Perhaps this will also allow us to make API calls from children components, instead of passing around callbacks.

The `onNext` function in the `Experiment` component (which retrieves the next action from the `next_round` array, or makes a call to the backend for a new round if that array is empty) still needs some consideration, however. Perhaps `useContext` would also be fine here, since we typically *do* rerender anyway every time we get a new action, but there are efficiency warnings all over the place about `useContext` which make me wonder whether we should use it as a catch-all solution to maintain state over components.

Making this issue to collect ideas, which can hopefully be turned into more concrete issues.

BeritJanssen:
What about [useSyncExternalStore](https://react.dev/reference/react/useSyncExternalStore) for participant / session data? It seems that this is exactly what we want: creating an Observable that multiple components can subscribe to to get the latest participant and session information. We do want to avoid that new components subscribing trigger a new API call, however.

PR #706, opened 2024-01-09
Feature/zustand store

BeritJanssen:
This branch adds the `zustand` library and uses it to handle the state of the participant. Addresses #655. The same solution might be adopted for handling the state of `session`, which is now still passed down over mulitple children.


frontend/src/API.js
```diff
@@ -37,8 +37,15 @@ export const URLS = {
...
-    useGet(API_BASE_URL + URLS.participant.current + urlQueryString);
+export const getParticipant = async (urlQueryString) => {
+    try {
+        const response = await axios.get(API_BASE_URL + URLS.participant.current + urlQueryString);
+        return response.data;
+    } catch (err) {
+        console.error(err);
+        return null;
+    }
+}
```
drikusroor, line 48: Perhaps it's better to leave the error-handling at the place where you can decide what to do with the error. (see next comment)

frontend/src/components/App/App.js
```diff
@@ -1,18 +1,38 @@
...
+    
+    useEffect(() => {
+        if (queryParams && !(new URLSearchParams(queryParams).has("participant_id"))) {
+            console.error("Unknown URL parameter, use ?participant_id=")
+            return;
+        }
+        getParticipant(queryParams).then(data => {
+            setParticipant(data);
+        })
+    }, [])
```
drikusroor, line 34: If you handle the error here, you could choose to show an error message if you want and prevent the experiment from being shown/loaded.

frontend/src/components/StoreProfile/StoreProfile.js
```diff
@@ -28,7 +29,7 @@ const StoreProfile = ({ history }) => {
         }
     };
 
-    if (loadingParticipant) {
+    if (!participant) {
         return <Loading />;
     }
```
drikusroor, line 34: Related to the previous comments: What happens if fetching the participant fails and participant stays `null` forever? It will probably show the loading indicator forever as well, causing the user to think it's still loading.

drikusroor (approved):
While I do think the error handling could be done differently, especially with regards to the user experience, I do think it's an important step forwards into adding a global state management configuration. I'll leave it up to you what to do with my comments and approve it already.

Later, if we want to add more stores, we could think about using "[slices](https://github.com/pmndrs/zustand/blob/main/docs/guides/slices-pattern.md)", which makes combining different (entity) stores into one global store easier. I've experimented with this in one of my personal projects. Feel free to [check it out here](https://github.com/drikusroor/popi/blob/c1a0a2c221a4a2c86ef36a5be666a832f9aff2db/web/src/store/index.ts).

frontend/src/API.js
BeritJanssen: I tried to make the code accessible elsewhere in the application -- but indeed, we don't need to call this in multiple locations, that's the whole point of using the store.

PR #717, opened 2024-01-17
Config (Traefik): Configure a Traefik reverse proxy Docker container to use muscle.local and backend.muscle.local domains

drikusroor:
Not sure if we want this, but during yesterday's demo session I heard some people talking about not remembering the urls and port numbers to use for the application and the backend. This PR adds a Traefik container that reroutes `http://localhost:3000` and `http://localhost:8000` to `http://muscle.local` and `http://backend.muscle.local`, which is easier to remember (and of course you can also just type `muscle.local` and `backend.muscle.local`).

Additionally, it adds a idempotent script (`./scripts/add-hosts-entries`) to add the hostnames to the hostfile in `/etc/hosts` file, which is required Traefik to work.


BeritJanssen (commented):
Nice! Perhaps just "muscle" (instead of "muscle.local") might even do? Just for my understanding: where does the actual mapping of the routes to the relevant localhost ports happen? Is it in the add-hosts-entries script? I worry that the port mappings are not explicit now, so it might be a bit hocus-pocus to future developers.

drikusroor:
> Nice! Perhaps just "muscle" (instead of "muscle.local") might even do? Just for my understanding: where does the actual mapping of the routes to the relevant localhost ports happen? Is it in the add-hosts-entries script? I worry that the port mappings are not explicit now, so it might be a bit hocus-pocus to future developers.

That happens in the Traefik container (see the `docker-compose.yml` configuration). Traefik functions as a reverse proxy here. The `add-hosts-entries` script only maps the domains to `127.0.0.1`. As Traefik runs on port `80` it will be hit when you send requests to `127.0.0.1` and then proxy it internally: 
```
muscle.local => 127.0.0.1(:80) (Traefik) => localhost:8000
```

The only thing I'm still thinking about is how to execute the `add-hosts-entries` script together with `docker compose up (-d --build)` so that the hosts will always be up to date with the Traefik configuration.

docker-compose.yaml
```diff
@@ -55,6 +60,10 @@ services:
           - 8000:8000
         command: bash -c "python manage.py migrate && python manage.py bootstrap && python manage.py runserver 0.0.0.0:8000"
     client:
+        labels:
+          - "traefik.enable=true"
```
BeritJanssen, line 64: Does this line mean that traefik will proxy the container's port?

docker-compose.yaml
```diff
@@ -55,6 +60,10 @@ services:
...
+          - "traefik.http.routers.client.rule=Host(`muscle.local`)"
+          - "traefik.http.routers.client.entrypoints=web"
```
BeritJanssen, line 66: What does this setting do? What is the "web" entrypoint?

BeritJanssen:
Perhaps add some comments to the docker-compose then to explain what the `traefik` settings do?

docker-compose.yaml
drikusroor: Yes, the labels from line 64-66 will tell traffic that this service to proxy `muscle.local` to this service (which happens to run on port `3000`)

docker-compose.yaml
drikusroor: `web` in this context means insecure http traffic, which is fine in a development environment context. If this were to be a production configuration we might want to use `websecure` instead of `web`.

drikusroor:
> Nice! Perhaps just "muscle" (instead of "muscle.local") might even do? Just for my understanding: where does the actual mapping of the routes to the relevant localhost ports happen? Is it in the add-hosts-entries script? I worry that the port mappings are not explicit now, so it might be a bit hocus-pocus to future developers.

I don't think the port mappings are that important. Traefik communicates with Docker and simply proxies to that specific Docker service. What might be interesting to check is to start up the newly added Traefik container (I just merged the PR) and navigate to `http://localhost:8080`, which is the backend interface of Traefik that shows you how the different services, load balancers, etc. are currently configured.

Also, I've added some explanation to the Wiki of how to start up the Development environment: https://github.com/Amsterdam-Music-Lab/MUSCLE/wiki/2.-Start-the-application

PR #722, opened 2024-01-19
Podmanify MUSCLE applications to enable automatic CI/CD

### LINKED ISSUE #718: Podman-ify the application so that we automatically deploy to the new (test) server
Currently, it's tedious to deploy the MUSCLE project. We have to ssh into one of the servers, fetch the latest branch, run docker commands, etc. Ideally, we should have some form of CI/CD with automated deployment, reducing our workload and shortening the feedback cycle for testing.

There has been a plan to convert our Docker setup to use Podman (which is similar and compatible). Podman then runs on the server, and by using a self-hosted GitHub Action runner we can run the required Podman commands from our Github workflow.

**Requirements**:

* [x] Update (or for now add) workflow to run Podman commands to *build* the Podman images that will eventually run on the new server
* [x] Update the Dockerfiles where needed
* [x] Update (or for now add) workflow to run Podman commands to *run* (and therefore deploy) the Podman containers on the new server
* [x] ~Move the Storybook jobs to its own workflow that still runs on Github's cloud runner?~ Update: it already is separate!
* [x] Change DNS settings so that [tst.amsterdammusiclab.nl](https://tst.amsterdammusiclab.nl) will give us the frontend and `/admin` will give us the Django backend dashboard. *The DNS change has already been requested
* [x] Configure it so that the application will be deployed after pushes to `develop` (or by a manual trigger) to the test environment using the test environment variables configured in Github.
* [ ] Verify that everything works and if not, fix things where necessary.

* I think we have talked of just building the frontend in GitHub Actions using a cloud runner with a basic yarn image. But having everything configured in Podman is more consistent, I guess.

**Additional context**
- We can draw inspiration from the AML blog that [has already been podmanified](https://github.com/Amsterdam-Music-Lab/amsterdam-music-lab/pull/18).
- We aren't Podman connaisseurs, so please ask Auke for advice if you are stuck
- See also:
  - #570 
  - #592
  - Work in progress: #722

drikusroor:
This PR adds a Github workflow that runs on a self-hosted Github Runner on the test environment. The setup uses Podman and podman-compose (instead of directly using Docker and docker-compose) through a newly created `docker-compose-deploy.yml` configuration. The workflow is configured to run on push to `develop` or through a manual `workflow_dispatch`. This means that the test environment should and will always be in sync with the `develop` branch and we don't have to manually deploy anymore. We can later decide wether we want to re-use this workflow for the acceptance and production environment with separate jobs that depend on different conditions or that we want to create separate workflows for that.

Resolves #718


docker-compose-deploy.yml
```diff
@@ -0,0 +1,108 @@
...
+      - DJANGO_SUPERUSER_EMAIL=${DJANGO_SUPERUSER_EMAIL}
+      - DJANGO_SUPERUSER_PASSWORD=${DJANGO_SUPERUSER_PASSWORD}
+      - SENTRY_DSN=${SENTRY_DSN}
+      - SQL_DATABASE=${SQL_DATABASE}
+      - SQL_USER=${SQL_USER}
+      - SQL_PASSWORD=${SQL_PASSWORD}
+      - SQL_HOST=${SQL_HOST}
+    ports:
+      - 8000:8000
+    command: bash -c "python manage.py migrate && python manage.py bootstrap && python manage.py collectstatic --noinput && gunicorn aml.wsgi:application --bind 0.0.0.0:8000"
```
BeritJanssen, line 59: Just an idea, we can also do this later: we might start up gunicorn with multiple workers, which may help us with high server loads.

frontend/Dockerfile
```diff
@@ -1,9 +1,22 @@
-FROM node:18-alpine
+# Builder image
+FROM node:18-alpine as builder
```
BeritJanssen, line 2: Don't we need docker-io here, too?

BeritJanssen (approved):
LGTM! Nice to have the production setup right in this repo, that's going to make it easier for developers from other labs to set it up on servers, too.

frontend/Dockerfile
drikusroor: Yeah, that would make sense, thanks! 👍 

```suggestion
FROM docker.io/node:18-alpine as builder
```

PR #725, opened 2024-01-22
Remove StartSession component

### LINKED ISSUE #714: Move `createSession` request to Experiment component?
There is a `StartSession` component whose only responsibility is to fire a request to create a session, and pass it through to its parent (`Experiment`). @jaburgoyne Is there any scenario in which we would want to *not* create a session? It seems silly to explicitly set `StartSession` through the backend in the `first_round`, if we cannot ever proceed to the `next_round` calls, which depend on the session to be created. I would propose to get rid of that component and just fire the `createSession` call from the `Experiment` component. This avoids:
- unnecessary repetition of `StartSession` in all rules files
- the definition of a component without any html (other than error handling, which can be done at a higher level just as well)
- problems with passing through data between `Experiment` and `StartSession`

BeritJanssen:
Close #714 .


frontend/src/components/Experiment/Experiment.js
```diff
@@ -65,12 +65,26 @@ const Experiment = ({ match }) => {
...
+            return;
+        }
+        try { 
+            createSession({
+                experiment,
+                participant,
+                playlist,
+            }).then(data => {
+                setSession(data.session);
+                setSessionStore(data.session);
```
drikusroor, line 79: Why do we need these both still?

frontend/src/components/Experiment/Experiment.js
```diff
@@ -99,7 +110,7 @@ const Experiment = ({ match }) => {
         } else {
             // Try to get next_round data from server
             const round = await getNextRound({
-                session: session.current
+                session: session
```
drikusroor, line 113:
```suggestion
                session
```

drikusroor (approved):
Nice, I love it when a PR removes more code than it adds. I'm only wondering why we need to combine a local session state and a store session state. Or is this something we intend to de-duplicate later on?

frontend/src/components/Experiment/Experiment.js
BeritJanssen: This is for "historical reasons" - I thought to leave adapting the child components of `Experiment`, some of which still expect a `session` prop, for later.

PR #726, opened 2024-01-22
Feature/intermediate score

### LINKED ISSUE #484: Save the data from unfinished boards Matching-pairs
Currently, only possible to save the data of the finished board on the server. It would be helpful if we could also save data from unfinished boards, so that we could include more data in our analysis. This issue is to mark this wish so that it can be realised sometime in the future.

BeritJanssen:
Close #484 : move scoring to backend. The drawback of this approach is that with slow connection speeds, players will have to wait before they get to see feedback on their move. Might also do double bookkeeping instead -- once in the frontend component, once in the backend -- but that seems a bit awkward.


backend/experiment/rules/matching_pairs.py
```diff
@@ -117,14 +118,42 @@ def get_matching_pairs_trial(self, session):
...
+        else:
+            if 'seen' in second_card:
+                score = -10
+                given_response = 'misremembered'
+            else:
+                score = 0
+                given_response = 'no match'
+        prepare_result('move', session, json_data=result_data,
+                       score=score, given_response=given_response)
+        return score
```
drikusroor, line 157: Perhaps a unit test can be added for this method that simulates the requests made from the frontend.

backend/experiment/rules/matching_pairs.py
```diff
@@ -117,14 +118,42 @@ def get_matching_pairs_trial(self, session):
             title='Tune twins',
             playback=playback,
             feedback_form=None,
-            result_id=prepare_result('matching_pairs', session),
+            result_id=prepare_result('matching_pairs_board', session),
```
drikusroor, line 121: Why is this renamed?

backend/result/views.py
```diff
@@ -81,3 +74,24 @@ def get_result(request, question):
...
+        session = Session.objects.get(
+            pk=session_id, participant__id=participant.id)
+    except Session.DoesNotExist:
+        return HttpResponseServerError("No session found")
+
+    # Prevent creating results when session is finished
+    if session.is_finished():
+        return HttpResponseServerError("Session has already finished")
+
+    return session
```
drikusroor, line 97: Can't we make a middleware out of this and put an attribute above the `intermediate_score` method? Perhaps out of scope for now, but perhaps something worth considering.

frontend/src/API.js
```diff
@@ -116,6 +117,30 @@ export const scoreResult = async ({
...
+    try {
+        const vars = {
+            session_id: session.id,
+            json_data: JSON.stringify(result),
+            csrfmiddlewaretoken: participant.csrf_token
+        };
+
+        const response = await axios.post(
+            API_BASE_URL + URLS.result.intermediateScore,
+            qs.stringify(vars)
```
drikusroor, line 134: What is `qs`? Is it used to stringify json? Can't the built-in `JSON` library already do that?

frontend/src/API.js
```diff
@@ -116,6 +117,30 @@ export const scoreResult = async ({
...
+        );
+        return response.data;
+    } catch (err) {
+        console.error(err);
+        return null;
+    }
```
drikusroor, line 140: Wouldn't it be better to leave the error handling to the component in question? Perhaps `MatchingPairs` wants to handle an error differently than another component.

frontend/src/components/Playback/MatchingPairs.js
```diff
@@ -100,20 +98,8 @@ const MatchingPairs = ({
...
-                        score.current = 10;
-                        setTurnFeedback('0')
-                    }
-                } else {
-                    if (currentCard.seen) { score.current = -10; }
-                    else { score.current = 0; }
-                    setTurnFeedback('-1')
-                };
+                const imScore = await scoreIntermediateResult({session, participant, result: {currentCard, lastCard}});
+                score.current = imScore.score;
```
drikusroor, line 102: if the http request of `scoreIntermediateResult` fails, it returns a `null` value. When that happens you cannot do `score.current = imScore.score;`. It might block the application, or not?

drikusroor (changes requested):
In general a good step forwards to saving the score after every step, but I do think the intermediate score method (in the backend) of matching pairs deserves a test and that the error handling in calling the intermediate score method in the frontend could be done better.

backend/experiment/rules/matching_pairs.py
BeritJanssen: This name was changed to indicate that this result is about the whole board. I'm actually wondering whether to get rid of this result and its handling in `calculate_score` altogether, as it's not strictly necessary to log this when clearing the board.

backend/result/views.py
BeritJanssen: Perhaps a decorator will do?

frontend/src/API.js
BeritJanssen: This is using the same pattern as other functions in `API.js`.

BeritJanssen: Replacing `qs.stringify` by `JSON.stringify` doesn't work at any rate, tbh I also don't know why `qs.stringify` is needed here. See [this thread](https://stackoverflow.com/questions/15850866/are-json-and-node-js-querystring-the-same) for more info. Apparently Django expects a different form of serialization than provided by JSON.

frontend/src/components/Playback/MatchingPairs.js
BeritJanssen: Good point!

backend/result/views.py
drikusroor: > Perhaps a decorator will do?

Yes, that's the word I was looking for!

backend/experiment/rules/matching_pairs.py
BeritJanssen: I removed the creation of this result, and the `calculate_score` function altogether.

frontend/src/API.js
BeritJanssen: I followed the pattern of most API functions right now: return null and log the error to the console. It's the responsibility of components to display an error to the user. Throwing the error here gives less useful output to a participant. We should display a message that something went wrong, and developers can see the full error message via Sentry.

BeritJanssen:
I added a test and introduced a global store for errors, which are now all displayed through the App component.

backend/result/views.py
BeritJanssen: I decided against turning this into a decorator, as I'm not sure this would cut down the number of lines beyond a flat function call.

drikusroor (approved):
LGTM!

Issue #732, opened 2024-01-23
Import modules in `experiment.rules` automatically


BeritJanssen:
**Is your feature request related to a problem? Please describe.**
Right now, a lot of bookkeeping is needed in order to add a new rules file: next to the file itself, we also need to update `experiment.rules.__init__.py`.

**Describe the solution you'd like**
We might walk over the submodules of `rules` and add them to the `__all__` object, and use that over the custom dict in the selection of rules in `experiment.admin`, see: https://stackoverflow.com/questions/3365740/how-to-import-all-submodules

## Important ⚠️

Please remember to remove the rules registration logic from #735 ❗️

albertas-jn:
I think the main issue here is `EXPERIMENT_RULES` in `experiment/rules/__init__.py`. We would still need to manually edit this dictionary,  because experiment ID strings are not the same as the names of the corresponding classes. So I am not sure if automatic importing of rules modules is worth doing.

BeritJanssen:
Hm, perhaps this is not worth it then indeed. Obviously, we should get rid of the ugly `EXPERIMENT_RULES` construct (see also #330 ), and write a method in `__init__.py` which automatically imports the `__repr__` or `__init__` methods from the rules into a dict. I think that at least should be doable.

PR #735, opened 2024-01-25
Added: Add a Django management command to create new experiments

drikusroor:
This PR adds a Django management command that allows a developer to create new experiments.

* It first asks the user for the experiment name
* Based on that name it will create a new file with the experiment name in snake case (e.g. `experiment/rules/snake_case.py`) based on a template in `experiment/management/commands/templates/experiment.py`
* It will register the experiment to `experiment/rules/__init__.py` (this might have to be removed later, see also #732 )
* Lastly, it will add a basic unit test in `experiment/rules/tests/test_snake_case.py`, based on the template in `experiment/management/commands/templates/test_experiment.py`

Additionally, this pull request adds a script that allows executing Django management commands through Docker, making it easier to execute this specific command.

## Demonstration

https://github.com/Amsterdam-Music-Lab/MUSCLE/assets/8208970/860d6aed-c3bd-48cb-b60c-b640ce71ceb5

### DIFF
--- a/backend/experiment/management/commands/createexperiment.py
+++ b/backend/experiment/management/commands/createexperiment.py
@@ -29,6 +29,8 @@ class Command(BaseCommand):
             if overwrite.lower() != 'y':
                 self.stdout.write(self.style.WARNING(f"File {filename} was not created"))
                 return
+            else:
+                self.stdout.write(self.style.WARNING(f"File {filename} will be overwritten"))
 
         # Create the file by copying ./experiment/management/commands/templates/experiment.py
         with open(filename, 'w') as f:
--- a/backend/experiment/management/commands/createexperiment.py
+++ b/backend/experiment/management/commands/createexperiment.py
@@ -9,6 +9,16 @@ class Command(BaseCommand):
         # Ask for the experiment name
         experiment_name = input("What is the name of your experiment? (ex. Musical Preferences): ")
 
+        # Create the experiment rule class
+        self.create_experiment_rule_class(experiment_name)
+
+        # Add the new experiment to ./experiment/rules/__init__.py
+        self.register_experiment_rule(experiment_name, './experiment/rules/__init__.py')
+
+        # Create a basic test file for the experiment
+        self.create_test_file(experiment_name)
+
+    def create_experiment_rule_class(self, experiment_name):
         # Get the experiment name in different cases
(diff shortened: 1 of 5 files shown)


backend/experiment/management/commands/createexperiment.py
```diff
@@ -0,0 +1,122 @@
...
+        filename = f"./experiment/rules/{experiment_name_snake_case}.py"
+
+        # Check if the file already exists
+        if os.path.isfile(filename):
+            # Warn the user that the file already exists and ask if they want to overwrite it
+            self.stdout.write(self.style.WARNING(f"File {filename} already exists"))
+            overwrite = input("Do you want to overwrite it? (y/n): ")
+
+            # If the user does not want to overwrite the file, exit the command
+            if overwrite.lower() != 'y':
```
BeritJanssen, line 36: I worry a bit that giving an option to overwrite may break existing experiments in the database. I think it's safer not to provide that option.

backend/experiment/management/commands/createexperiment.py
```diff
@@ -0,0 +1,122 @@
...
+
+        with open(file_path, 'r') as file:
+            lines = file.readlines()
+
+        # Find the line to insert the new import, maintaining alphabetical order
+        import_index = next(i for i, line in enumerate(lines) if line.startswith('from .') and line > new_import)
+        lines.insert(import_index, new_import)
+
+        # Find the line to insert the new dictionary entry, maintaining alphabetical order within the EXPERIMENT_RULES
+        dict_start_index = lines.index("EXPERIMENT_RULES = {\n") + 1  # Start after the opening brace
```
BeritJanssen, line 71: NB: I was hoping to supersede the `EXPERIMENT_RULES` dict with the solution proposed in #732

BeritJanssen (changes requested):
Nice idea! I don't think it's wise to let experimenters overwrite existing rules files though. Other than that, this looks great! Oh yeah, a small note: your code formatter doesn't seem to insert empty lines at the end of files. I understand some other code editors may be thrown by missing empty lines (hence Github flags these as well).

backend/experiment/management/commands/createexperiment.py
drikusroor: I can make an issue to remove this method after #732 is merged? Or should I remove it here first?

BeritJanssen (approved):

PR #738, opened 2024-01-26
Revert: Remove restricted field from section

### LINKED ISSUE #729: Remove `Song.restricted` field
Right now, we have an option to restrict playback of a section, defined on its associated song, to specific countries. We can actually remove this field altogether, as it's usually allowed to play copy-righted songs for academic purposes. This would mean the following:
- [ ] remove the field `restricted` from `Song` model and add a migration
- [ ] update the `playlist.update_sections` function and its test to not expect that csv field anymore
- [ ] update the hint under the Playlist csv TextField in the admin interface
- [ ] update the csv upload / download functionality accordingly
- [ ] remove all utility functions which make use of the `restricted` notion

drikusroor:
Resolves #729


BeritJanssen (approved):

BeritJanssen:
NB: before merging, check with @jaburgoyne whether we can definitely remove the "restricted_to_country_x" option for section playback.

backend/experiment/migrations/0019_alter_experimentseries_options.py
BeritJanssen, line 1: It turns out I'm working on the experiment.migrations in parallel. It would be easier for me if you could remove this migration from this branch, I'll add it in mine then.

drikusroor: I can't merge this PR without a migration anyway. The PRs are blocking each other. So either we merge this one and you recreate your migration, or you merge your feature and then I recreate the migration here and merge the PR. Which do you prefer?

BeritJanssen: This is an old migration, for changing the plural of ExperimentSeries. I thought you could just kick this migration in experiments out, without any major hickups. Or do the tests fail? In that case, indeed, merge here, and I'll just have to overhaul the migration on the other branch.

drikusroor: Ah I see.. I thought it was related to the `restricted` field. My bad :-) I've removed the file!

PR #739, opened 2024-01-29
Chore: Merge main into develop (after 0.8.0)

drikusroor:


BeritJanssen:
I was confused at first why this would be necessary. In git-flow, hotfix branches are always merged to develop *and* main, so then you'd never merge main into develop.

BeritJanssen (commented):
I'm confused. I see commits in this branch which aren't reflected in the "files changed" - could it be the hotfix commits have already be integrated in develop?

drikusroor:
Yes, this PR is probably not necessary. I made a hotfix last friday but I haven't properly used git-flow like we should

BeritJanssen:
We should still increase the version number on develop though, and the template update for a bug report has also not been picked up in develop yet. So I don't think this is completely superfluous.

PR #744, opened 2024-01-30
Bugfix er/iosversion fix

Evert-R:
This PR checks for the user's ios version and switches to 'EXTERNAL' if the version < 17.
It also fixes a bug that only one sound got played in 'EXTERNAL' mode in matching pairs.

### DIFF
--- a/frontend/src/util/webAudio.js
+++ b/frontend/src/util/webAudio.js
@@ -25,21 +25,21 @@ export const initWebAudioListener = () => {
 };
 
 // init HTML audio element in webaudio context and connect track to destination (output)
-export const initWebAudio = () => {    
-    if (track === undefined) {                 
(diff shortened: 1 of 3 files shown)


frontend/src/util/webAudio.js
```diff
@@ -136,3 +136,17 @@ export const listenOnce = (event, callback) => {
     source.addEventListener(event, _callback);
     return remove;
 };
+
+// function to check wether a users device is webaudio compatible
+export const compatibleDevice = () => {
```
BeritJanssen, line 141: Perhaps this is a nice function to write a test for?

BeritJanssen (approved):
Nice! Not a condition before merging, but certainly a nice addition would be a test for the `compatibleDevice` function.

PR #745, opened 2024-01-30
Fix Playbutton style

### LINKED ISSUE #578: [Categorization] Playback button lost its styling
The background of the playback button is missing.

### LINKED ISSUE #614: Increase size of PlayButton
To avoid this issue, which is caused by `Preload` (which is briefly rendered) taking more space than the `PlayButton`:

https://github.com/Amsterdam-Music-Lab/MUSCLE/assets/11174072/2e1e66ca-d87b-45bc-8989-b094fffaef42

albertas-jn:
Closes #578
Closes #614


frontend/src/components/PlayButton/PlayButton.scss
```diff
@@ -52,3 +52,7 @@ $playIconSize: 30px;
         }
     }
 }
+
+.playbutton-spacer {
+    height: 87px; // Difference in height between PlayButton and Preload
```
BeritJanssen, line 57: Does this work on different devices?

frontend/src/components/PlayButton/PlayButton.js
```diff
@@ -6,8 +6,9 @@ const PlayButton = ({ playSection, isPlaying, className="" }) => {
     const [clicked, setClicked] = useState(false);
 
     return (
+        <>
         <div
-            className={classNames("aha__play-button border-outside", "btn", {
+            className={classNames("aha__play-button btn-blue border-outside", "btn", {
```
BeritJanssen, line 11: Doesn't this mean that *all* PlayButtons will be blue from now on? Okay for now as a fix, but we should (re)introduce the option to override the play button color.

BeritJanssen (changes requested):
Nice! As for my two questions: if the answer to "all PlayButtons are blue" is *yes*, I'd still find this okay as an ad-hoc mix. With the spacer, can you confirm that this is stable across devices (i.e., different screen sizes)?

BeritJanssen (approved):

PR #749, opened 2024-01-31
Chore: Update package.json files with license, homepage, and description

### LINKED ISSUE #740: Add license field to `package.json`
I noticed we suppress a warning in `package.json` about a missing license field. Since we do have a license, we should add that to the `package.json`, too.

drikusroor:
Resolves #740


frontend/package.json
```diff
@@ -2,7 +2,9 @@
   "name": "aml-frontend",
   "version": "0.1.0",
   "private": true,
-  "homepage": "/",
+  "homepage": "app.amsterdammusiclab.nl",
+  "description": "The React Frontend for the MUSCLE platform",
+  "license": "MIT",
```
BeritJanssen, line 7: We probably don't need any information but the description in the frontend `package.json`. The version here might also be removed. But that might again raise all kinds of warnings, I suppose? Oh yes, `private: true` can be `false`, both here and in the main `package.json`. We don't have any plans to release (parts of) MUSCLE via npm, but also no need to block release.

BeritJanssen (approved):
Nice! If there's a way to not duplicate information (or, in case of version numbers, conflicting information) in the main and frontend `package.json`, that would be nice, but not a strict requirement before merging.

package.json
```diff
@@ -2,6 +2,9 @@
     "name": "aml-experiments",
     "version": "1.7.0",
     "private": true,
```
drikusroor, line 4:
```suggestion
    "private": false,
```

frontend/package.json
```diff
@@ -2,7 +2,9 @@
   "name": "aml-frontend",
   "version": "0.1.0",
```
drikusroor, line 3:
```suggestion
```

package.json
```diff
@@ -2,6 +2,9 @@
     "name": "aml-experiments",
     "version": "1.7.0",
```
drikusroor, line 3:
```suggestion
```

frontend/package.json
```diff
@@ -2,7 +2,9 @@
...
   "private": true,
-  "homepage": "/",
+  "homepage": "app.amsterdammusiclab.nl",
```
drikusroor, line 5:
```suggestion
```

PR #757, opened 2024-02-06
add Trial to template

BeritJanssen:
I was thinking along those lines for the template for `createexperiment`: add a `Trial` which demonstrates a typical setup. Still to be tested though, but just in general, @albertas-jn , do you think this could work? If you can already add documentation about the different methods in a typical rules file (i.e., first_round, next_round) to the wiki, that would certainly be helpful!

@drikusroor I removed the `feedback_info` method, since I figured we probably wouldn't *need* to override this - or is there a reason you did in the original template? I also got rid of the `questions` array in the `__init__`, because we can now select questions through the admin interface, and I'd rather that users choose that method over hard-coding them in the rules file.

### DIFF
--- a/backend/experiment/management/commands/templates/experiment.py
+++ b/backend/experiment/management/commands/templates/experiment.py
@@ -27,7 +27,6 @@ class NewExperiment(Base):
                             'isced-2', 'isced-5'])
         ]
 
-
     def first_round(self, experiment):
         # 1. Informed consent (optional)
         rendered = render_to_string('consent/consent.html')
--- a/backend/experiment/management/commands/templates/experiment.py
+++ b/backend/experiment/management/commands/templates/experiment.py
@@ -8,10 +8,10 @@ from experiment.questions.demographics import EXTRA_DEMOGRAPHICS
 from experiment.questions.utils import question_by_key
 from experiment.rules.base import Base
 from result.utils import prepare_result
-from section.models import Playlist
 
 
 class NewExperiment(Base):
+    ''' An experiment type that could be used to test musical preferences '''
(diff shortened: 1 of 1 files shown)


drikusroor:
> @drikusroor I removed the `feedback_info` method, since I figured we probably wouldn't _need_ to override this - or is there a reason you did in the original template? I also got rid of the `questions` array in the `__init__`, because we can now select questions through the admin interface, and I'd rather that users choose that method over hard-coding them in the rules file.

That's okay, I actually didn't put _that_ much thought into the actual content of the template, but more in making the management command work. I figured we would and will refine it every now and then based on (new) insights and features. So please feel free to update where you deem necessary!

albertas-jn:
@BeritJanssen you can't remove the `questions` array in the `__init__`, because this is how admin determines default questions for the rule file.

Also, I am thinking that perhaps the template that I have written before in #302 would be a better way to go? Because it actually defines something that is running and playing sounds. With the command `createexperiment`, you can only change the name, but still need to add code yourself. If people will need to add code, then they will also be able to copy a template file and edit it themselves.

BeritJanssen:
@albertas-jn , indeed, I was also wondering about whether to make it a "useful" experiment. I now brought it back so that it could be run with any playlist. I agree that `running a command` vs. `copy and rename this file` is similar in difficulty for the user. What I like about the `command` approach is that it takes care of quite a lot of details when making a new experiment, i.e., actually renaming the file, and also renaming the class & ID, as well as adding the class to `__init__.py` so it can be discovered (even though I'm hoping we can perhaps smooth that particular step in #732 .

PR #764, opened 2024-02-13
Feature er/add consent field

Evert-R:
This PR introduces a `consent` field on `experiment`

This field will override any consent text loaded from the rules file. So every instance of the experiment can have a different text.
The action in `consent.py` has been modified to be able to use both HTML (default) and MARKDOWN text. 
Template files will now be loaded from the action file instead of the rules file using the `url` argument. 
To render the markup language this package had been installed: [Django markup](https://pypi.org/project/django-markup/)
The render format can be set using `render_format` on the `Consent` action. HTML is default.
Django template tags are also rendered from the consent field. So if a translation is available django will also translate the text in the consent field. 

All rules files have been modified for this new setup.


backend/experiment/actions/consent.py
```diff
@@ -22,8 +32,27 @@ class Consent(BaseAction):  # pylint: disable=too-few-public-methods
...
+        # from field: experiment.consent (prio-1)
+        if text!='':
+            dry_text = text
+        # from template file (prio-2)    
+        elif url!='':
+            dry_text = render_to_string(url)
+            url_length = len(url)
+            if url[(url_length-2):url_length] == 'md':
+                render_format = 'MARKDOWN'
+        # use efault text (prio-3)
```
BeritJanssen, line 46: I don't understand this comment.

backend/experiment/management/commands/templates/experiment.py
```diff
@@ -27,10 +27,14 @@ def __init__(self):
...
-        rendered = render_to_string('consent/consent_new_experiment.html')
-        consent = Consent(rendered, title=_(
-            'Informed consent'), confirm=_('I agree'), deny=_('Stop'))
+        # 1. Informed consent from file or admin (admin has priority)        
+        consent = Consent(
+            experiment.consent, 
+            title=_('Informed consent'), 
+            confirm=_('I agree'), 
+            deny=_('Stop'), 
+            url='consent/consent_new_experiment.html'
```
BeritJanssen, line 36: Please change this to `consent/consent.html`, as `consent_new_experiment.html` is not auto-created, and also shouldn't be.

backend/experiment/migrations/0019_add_consent_field.py
```diff
@@ -0,0 +1,22 @@
...
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='experimentseries',
+            options={'verbose_name_plural': 'Experiment Series'},
+        ),
+        migrations.AddField(
+            model_name='experiment',
+            name='consent',
+            field=models.TextField(blank=True, default=''),
```
BeritJanssen, line 20: I wonder whether a FileUpload field would be better?

backend/requirements/dev.txt
```diff
@@ -71,13 +92,15 @@ pip-tools==7.3.0
 platformdirs==2.4.0
     # via pylint
 psycopg2==2.9.1
-    # via -r requirements.in/base.txt
+    # via -r requirements.in\base.txt
```
BeritJanssen, line 95: I wonder whether a backslash will cause problems on unix systems... But if this runs in the containers,  it's probably fine.

BeritJanssen (approved):
Nice! As stated in the comment, I wonder whether an upload field is better? Perhaps something to discuss. Also, I think that for the user upload, it might be better to go markdown only: this reduces the chance of having missing closing tags, etc. We also prevent `<script>` tags that way - even though we probably don't need to worry about admin users to try anything like that.

backend/experiment/actions/consent.py
```diff
@@ -22,8 +32,27 @@ class Consent(BaseAction):  # pylint: disable=too-few-public-methods
...
+    def __init__(self, text, title='Informed consent', confirm='I agree', deny='Stop', url='', render_format='HTML'):
+        # Determine which text to use
+        # from field: experiment.consent (prio-1)
+        if text!='':
+            dry_text = text
+        # from template file (prio-2)    
+        elif url!='':
+            dry_text = render_to_string(url)
+            url_length = len(url)
+            if url[(url_length-2):url_length] == 'md':
```
BeritJanssen, line 44: Where is the 'md' set? Is this autodetected? And if, where?

backend/experiment/migrations/0019_add_consent_field.py
Evert-R: I think a file field will be better. I made a new branch for this and will set this one to draft for the time being.

BeritJanssen:
This one can be closed now, right?

PR #766, opened 2024-02-13
Allow user to create experiment in db after creating experiment rule set

drikusroor:
Per a suggestion by @hhoning 

This pull request adds a new feature that allows the user to create an experiment in the database after creating an experiment rule set, which you often want to do anyway.

## Screenshot

image

### DIFF
--- a/backend/experiment/management/commands/createexperiment.py
+++ b/backend/experiment/management/commands/createexperiment.py
@@ -28,12 +28,7 @@ class Command(BaseCommand):
         create_experiment = input("Do you want to create a new experiment with the same name? (yes/no): ")
 
         if create_experiment.lower() == 'yes' or create_experiment.lower() == 'y':
-            self.stdout.write(self.style.WARNING("Creating a new experiment..."))
-            Experiment.objects.create(
-                name=experiment_rules_set_name, 
-                rules=experiment_rules_set_name,
-                slug=experiment_rules_set_name.lower().replace(' ', '_')
-            )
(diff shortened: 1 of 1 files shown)


BeritJanssen:
Good idea! Again, I would've liked an issue first to discuss with the team. Note that if you do this, the experiment still won't run because we cannot auto-attach a playlist, too.

drikusroor:
> Good idea! Again, I would've liked an issue first to discuss with the team. Note that if you do this, the experiment still won't run because we cannot auto-attach a playlist, too.

I could expand it with a warning to the user that it still has to create and attach a playlist, or I could also ask the user if the script should automatically create an (empty) playlist. What do you think?

BeritJanssen:
I'm not sure about the added value of this tbh. Let's discuss this in the next Scrum meeting.

hhoning:
My original suggestion came from the idea that CreateExperiment, with a certain nam, should show up with that name in the Experiment tab in the Admin (since you just generated that). But it does not. It in fact ends up a level lower in the Ruleset menu, doesn't it? So rename CreateExperiment into CreateRuleset would solve it :-)

BeritJanssen:
Should we close this branch, and instead rename the command to `create_ruleset`? I think we *do* want users to jump through the hoops of going into the admin interface & setting up a playlist as well. *Or* we check in a mini playlist into this repo, which will be auto-loaded and linked as well. But I do worry that if you "automagically" set everything up, the threshold to using the admin interface is going to actually get higher.

PR #768, opened 2024-02-14
Add a manage-podman script

chfin:
Add ./scripts/manage-podman, a copy of ./scripts/manage that uses podman-compose instead of docker-compose.
This script can be used to run "manage" commands inside a container when using podman instead of docker.

### DIFF
--- a/scripts/manage-podman
+++ b/scripts/manage-podman
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+# This script executes Django management commands through Docker.
+podman-compose run --rm server python manage.py $@


drikusroor:
Does this solve the toughbook situation or is this because we're moving to Podman anyways?

chfin:
Not related to the thoughbook issue. I've used podman instead of docker during the training sessions and there were some thoughts about supporting podman as a rootless alternative to docker both for users and potentially for production. Not sure what's the current status of that. But so far, podman works perfectly as a drop-in replacement, once it's installed and configured correctly.

drikusroor:
> Not related to the thoughbook issue. I've used podman instead of docker during the training sessions and there were some thoughts about supporting podman as a rootless alternative to docker both for users and potentially for production. Not sure what's the current status of that. But so far, podman works perfectly as a drop-in replacement, once it's installed and configured correctly.

We are [currently in the process](https://github.com/Amsterdam-Music-Lab/MUSCLE/pull/722) of migrating the MUSCLE project to Podman on the servers. I think it does make sense to then also start using Podman locally during development to keep things consistent, but I'm not sure if that discussion has been started yet. If that will be the case, it would probably better to replace all references to `docker-compose` with `podman-compose`, making this PR obsolete. Perhaps you could also modify the existing manage script to use `podman-compose` and we can merge it after #722 has been merged?

chfin:
I'm not sure how easy it is to use podman on Mac and Windows. Looks like it's build for Linux and uses a [VM on Mac and WLS on Windows](https://podman.io/docs/installation), so it might not quite as straightforward to use for development. Maybe it's better to allow a user to use either docker or podman, depending on what is more convenient, or would that be too much effort to maintain?

BeritJanssen:
Yes, let's add support for Podman, but as an extra option next to Docker.

BeritJanssen (approved):

drikusroor:
> Yes, let's add support for Podman, but as an extra option next to Docker.

Does that mean we'll have to make Podman versions of the tother test/storybook/lint scripts too? Wouldn't it be nicer to update all the existing scripts with a parameter like `-p` that runs the same command using Podman?

chfin:
Did you decide against this approach in the end?

drikusroor:
> Did you decide against this approach in the end?

We thought it might be better to nudge people into aliasing `podman` and `podman-compose` if they would like to use Podman, see also #803

PR #771, opened 2024-02-14
Refactor stores to use Zustand's slice pattern

### LINKED ISSUE #752: Proposal: Improve the Zustand store by using the slice pattern and a context provider
**Problems & solutions**
The Zustand stores currently work fine as they are. We have multiple stores for multiple entities. However, there is one direct issue and one potential future issue that I'd like to address.

The first issue is that some storybook stories (`MatchingPairs` for example) do not work anymore as they rely on the store. It would be nice if we could add a [context provider](https://github.com/pmndrs/zustand/tree/main?tab=readme-ov-file#react-context) to the store that we could use in the stories to supply them with a store and its values and methods.

The second potential issue is that we might need more logic that relies on multiple store entities. Currently, the stores are separate, which means they cannot access each others' data. A nice way to combine the stores into one global store, while still keeping the entities/logic separate, is to use the [slice pattern](https://github.com/pmndrs/zustand/blob/main/docs/guides/slices-pattern.md). A benefit of this is that we can move important shared functions to the store instead of having them in components. An added benefit is that a component that relies on multiple stores/slices now only has to import a `useStore` function instead of three different stores (`useErrorStore`, `useParticipantStore`, and `useSessionStore`), like in the `Playlist` component for example.

drikusroor:
This pull request refactors the existing stores to use Zustand's slice pattern.

Resolves #752

### DIFF
--- a/frontend/src/stories/MatchingPairs.stories.js
+++ b/frontend/src/stories/MatchingPairs.stories.js
@@ -1,6 +1,22 @@
+import useBoundStore from 'util/stores';
 import MatchingPairs, { SCORE_FEEDBACK_DISPLAY } from '../components/Playback/MatchingPairs';
 
 import audio from './assets/audio.wav';
+
+
+const Decorator = (Story) => {
+    const setSession = useBoundStore(state => state.setSession);
+    const setParticipant = useBoundStore(state => state.setParticipant);
+    setSession({ id: 1 });
+    setParticipant({ id: 1, csrf_token: '123' });
+
+    return (
+        <div id="root" style={{ width: '100%', height: '100%', backgroundColor: '#ddd', padding: '1rem' }}>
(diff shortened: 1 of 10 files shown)


frontend/src/stories/MatchingPairs.stories.js
```diff
@@ -1,7 +1,23 @@
+import useBoundStore from 'util/stores';
 import MatchingPairs, { SCORE_FEEDBACK_DISPLAY } from '../components/Playback/MatchingPairs';
 
 import audio from './assets/audio.wav';
 
+
+const Decorator = (Story) => {
```
BeritJanssen, line 7: Maybe call this `StoreDecorator` to make it more desriptive?

BeritJanssen (approved):

PR #774, opened 2024-02-19
Feature er/matching pairs light

### LINKED ISSUE #730: Create a stripped down MatchingPairs rules file
Currently, `matching_pairs.py` shows an Explainer, Consent form, and questions. For field work, this is all unnecessary, as explanation, consent and questionnaire will happen outside of MUSCLE. So we need a `matching_pairs` rules file that leaves everything out, and moves the score feedback to a less prominent place. This can be achieved by passing `bottom_right` as a score display option.

Evert-R:
This PR adds a stripped down rules file for Matching Pairs. `MatchingPairsLite`
I needed to ad a `start` button to have a user click on the page so the audio engine gets initialized.
resolves #730


BeritJanssen (changes requested):
I think we can bring this file down to half its length by subclassing `MatchingPairs` instead. I don't expect we need any social media info in the Final round here?

BeritJanssen:
@Evert-R I tried a throw at it, it seems to work to inherit everything but `first_round` and `next_round`, plus class variables. Or am I overlooking something?

PR #776, opened 2024-02-20
Fix/experiment logic

BeritJanssen:
After experiments with #774 , it turns out that an empty `first_round` array is not desirable, as we need at least one user interaction (button click) with the document in order to load or play audio. This means the frontend `Experiment` component does not need a dependency on `continueToNextRound` in its `useEffect` hook anymore. That simplifies the setup considerably, and will hopefully prevent the multiple requests to `next_round` we've been seeing.

The only persisting problem is that the unit test gets stuck in `useEffect`: I think the mocked version of the custom  `useExperiment`, or more precisely the underlying `useGet` hook, is the culprit here.  Any ideas how to properly mock this - without triggering `useEffect` again and again - are very appreciated.

### DIFF
--- a/backend/experiment/management/commands/templates/experiment.py
+++ b/backend/experiment/management/commands/templates/experiment.py
@@ -28,15 +28,20 @@ class NewExperiment(Base):
         ]
 
     def first_round(self, experiment):
+        ''' Provide the first rounds of the experiment, 
+        before session creation
+        The first_round must return at least one Info or Explainer action
+        Consent and Playlist are often desired, but optional
+        '''
         # 1. Informed consent (optional)
         rendered = render_to_string('consent/consent.html')
         consent = Consent(rendered, title=_(
             'Informed consent'), confirm=_('I agree'), deny=_('Stop'))
         
-        # 2. Choose playlist (only relevant if there are multiple playlists the participant can choose from)
+        # 2. Choose playlist (optional, only relevant if there are multiple playlists the participant can choose from)
         playlist = Playlist(experiment.playlists.all())
 
-        # 3. Explainer (optional)
+        # 3. Explainer
         explainer = Explainer(
             instruction='Welcome to this new experiment',
             steps=[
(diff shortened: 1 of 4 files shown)


drikusroor:
> The only persisting problem is that the unit test gets stuck in useEffect: I think the mocked version of the custom useExperiment, or more precisely the underlying useGet hook, is the culprit here. Any ideas how to properly mock this - without triggering useEffect again and again - are very appreciated.

Is there a specific reason for wanting to mock `useGet`? Maybe it would be easier to mock axios http requests (specify the return value per url + request method combination). Doing that, `useGet` and `useExperiment` would also be covered in the unit tests indirectly, which is a nice bonus.

backend/experiment/management/commands/templates/experiment.py
```diff
@@ -28,15 +28,20 @@ def __init__(self):
         ]
 
     def first_round(self, experiment):
+        ''' Provide the first rounds of the experiment, 
+        before session creation
+        The first_round must return at least one Info or Explainer action
+        Consent and Playlist are often desired, but optional
```
drikusroor, line 34: Should we perhaps validate this somehow (idk if it's even possible)?

frontend/src/components/Experiment/Experiment.js
```diff
@@ -87,7 +87,7 @@ const Experiment = ({ match }) => {
             );
             setState(undefined);
         }
-    }, [checkSession, updateActions, setError, setState])
+    };
```
drikusroor, line 90: Nice, I like this cleanup of hooks :-)

frontend/src/components/Experiment/Experiment.test.js
```diff
@@ -10,45 +10,35 @@ jest.mock("../../util/stores");
...
 
 jest.mock("../../API", () => ({
-    useExperiment: () => [experimentObj, false],
+    useExperiment: () => {
+
+        return [experimentObj, false]
+    },
     createSession: () => Promise.resolve(sessionObj),
     getNextRound: () => Promise.resolve(nextRoundObj)
 }));
```
drikusroor, line 25: As mentioned in my comment, perhaps it would be better and easier to mock Axios requests and return the data that would normally initialise the experiment in the correct / desired way.

Something like this:

```js
import axios from 'axios';

jest.mock('axios');

axios.get.mockResolvedValue({ data: {} }); // Adjust as needed
```

BeritJanssen: I had been thinking along the same lines in the meantime. :-) Will try that.

PR #778, opened 2024-02-20
Fix: Increase file size limit for uploads in nginx configuration

### LINKED ISSUE #777: [TEST] We cannot upload files larger than X mb in the test environment right now
**Describe the bug**
When trying to upload an audio file (7.1 mb) in the [add sections page of a playlist](https://tst.amsterdammusiclab.nl/server/admin/section/playlist/), I got the following error: 

> ## 413 Request Entity Too Large
> 
> nginx/1.25.3

**To Reproduce**
1. Go to https://tst.amsterdammusiclab.nl/server/admin/section/playlist/
2. Click on add sections of a playlist (test playlist)
3. Select a reasonably larger (>= 7 mb) audio file
4. Click Add sections
5. See error

**Expected behavior**
- Audio file is uploaded successfully and added as a new section

**Screenshots**
image

 - OS: macOS
 - Browser Safari

**Additional context**

The solution is *probably* to increase the file size limit of nginx in `nginx/default.conf`. Something like this:

```conf
http {
    ...
    client_max_body_size 100M;
    ...
}
```

drikusroor:
This pull request adds the `client_max_body_size` directive to the nginx configuration and sets it to `100mb`, allowing for larger file uploads.

@BeritJanssen did or do we have a file size limit configured already somewhere in Django?

Resolves #777

### DIFF
--- a/nginx/custom-nginx.conf
+++ b/nginx/custom-nginx.conf
@@ -9,6 +9,8 @@ server {
 
     # Proxy pass to the Django app
     location /server/ {
+        client_max_body_size 100M;
+
         proxy_pass http://server:8000;
         proxy_set_header Host $host;
         proxy_set_header X-Real-IP $remote_addr;


drikusroor:
I've considered timeouts that might occur when uploading files. The default value in nginx is 60 seconds, which should be enough I think. If not, it should be very easy to increase it:

```conf
server {
    listen 80;

    client_max_body_size 100M;
    client_body_timeout 120s;
    client_header_timeout 120s;
    keepalive_timeout 120s;
```

PR #785, opened 2024-02-20
Feature er/add consent file

Evert-R:
This PR introduces a `consent` file field on experiment

This field will override any consent text loaded from the rules file. So every instance of the experiment can have a different text.
The action in `consent.py` has been modified to be able to use both HTML and MARKDOWN text.
Template files will now be loaded from the action script instead of the rules file using the url argument.
To render the markup language this package had been installed: [Django markup](https://pypi.org/project/django-markup/)
The render format is detected from the filename extention.

All rules files have been modified for this new setup.


backend/experiment/forms.py
```diff
@@ -145,6 +145,9 @@ class Meta:
         model = Experiment
         fields = ['name', 'slug', 'active', 'rules',
                   'rounds', 'bonus_points', 'playlists', 'experiment_series']
+        help_texts = {'consent': 'Enter the text to ask consent for using the experiment data for this instance of the experiment.<br> \
+                      This field will override any text loaded from the rules file.<br> \
+                      Allowed tags: HTML (default), MARKDOWN (Optional setting in the rules file)'}
```
BeritJanssen, line 150: Do you mean "allowed formats" here? I find this help text slightly confusing. What does "optional setting in the rules file" mean?

BeritJanssen (approved):

BeritJanssen:
Can you still fix the linting issue?

PR #792, opened 2024-02-23
Fixed: Set correct default slug in .env.dist to gold-msi

drikusroor:
Right now, whenever you start from scratch, the default Goldsmith MSI experiment that is created has the slug `gold-msi`. However, the default slug in `.env.dist` is set to `gmsi`. This would cause a new setup to move to a non-existent experiment when you would navigate to `http://localhost:3000` (which then redirects you to `http://localhost:3000/gmsi` This PR fixes this inconsistency.

Meanwhile, the default slug in `.env.github-actions` is set to `fe`. Should that be changed as well? Or perhaps the whole file can be removed as we now set that default slug through environment variables in Github.


BeritJanssen (approved):

BeritJanssen:
As for the question about `.env.github-actions`: if *all* the environment variables are now set through Github, it is probably indeed superfluous to keep track of them in the .env file.

PR #798, opened 2024-02-26
CI: Add restart options to Docker services

drikusroor:
This pull request adds the `restart` option to the Docker services in the `docker-compose.yml` file. The `restart` option ensures that the services automatically restart if they fail or if the Docker daemon restarts. This improves the reliability and availability of the application as the new test & acc environment's containers are currently down after a github runner update.

### DIFF
--- a/docker-compose-deploy.yml
+++ b/docker-compose-deploy.yml
@@ -21,10 +21,13 @@ services:
       interval: 5s
       timeout: 5s
       retries: 5
+    restart: always
   ip2country:
     image: extrawurst/ip2country:latest
     ports:
       - 8854:5000
+    restart: always
+
   server:
     build:
       context: ./backend
@@ -57,6 +60,8 @@ services:
     ports:
       - 8000:8000
     command: bash -c "python manage.py migrate && python manage.py bootstrap && python manage.py collectstatic --noinput && gunicorn aml.wsgi:application --bind 0.0.0.0:8000"
+    restart: always
+
   client-builder:
     build:
       context: ./frontend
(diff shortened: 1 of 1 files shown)


docker-compose-deploy.yml
```diff
@@ -57,6 +60,8 @@ services:
     ports:
       - 8000:8000
     command: bash -c "python manage.py migrate && python manage.py bootstrap && python manage.py collectstatic --noinput && gunicorn aml.wsgi:application --bind 0.0.0.0:8000"
+    restart: always
```
BeritJanssen, line 63: I would expect that the depends_on conditions would also restart db / ip2country. Can you check if this is the case? Just wondering whether it could be potentially problematic if the db and ip2country service got a signal from the `server` to restart, while also getting their own signal to restart.

docker-compose-deploy.yml
```diff
@@ -86,8 +91,9 @@ services:
       - REACT_APP_HTML_BODY_CLASS=${REACT_APP_HTML_BODY_CLASS}
       - REACT_APP_SENTRY_DSN=${REACT_APP_SENTRY_DSN}
       - REACT_APP_STRICT=${REACT_APP_STRICT}
+    restart: "no"
```
BeritJanssen, line 94: Well done! When I first included restart rules in the former docker-deploy, I let the frontend restart itself, too. 😆

BeritJanssen (approved):
Please do check if the restart on db / ip2country is really necessary, or if it's not sufficient to have the rule in place for server, which depends on the other two services (see my comment).

PR #806, opened 2024-02-27
Add CongoSameDiff experiment type

drikusroor:
This pull request adds the CongoSameDiff experiment type to the project. It includes the necessary code changes and tests.


backend/experiment/rules/congosamediff.py
```diff
@@ -0,0 +1,142 @@
...
+        '''
+
+        # 1. Playlist
+        playlist = Playlist(experiment.playlists.all())
+
+        # 2. Explainer
+        explainer = Explainer(
+            instruction='Welcome to this Same Diff experiment',
+            steps=[],
+            step_numbers=True
```
BeritJanssen, line 35: You can leave `step_numbers` out here, as you don't provide any steps.

drikusroor:
```suggestion
```

backend/experiment/models.py
```diff
@@ -213,6 +212,7 @@ def export_table(self, session_keys: List[str], result_keys: List[str], export_o
 
     def get_rules(self):
         """Get instance of rules class to be used for this session"""
+        from experiment.rules import EXPERIMENT_RULES
```
BeritJanssen, line 215: why this move? Was there a circular import problem?

backend/experiment/rules/tests/test_congosamediff.py
```diff
@@ -0,0 +1,115 @@
...
+        experiment = Experiment(id=1, name='CongoSameDiff', slug='congosamediff', rounds=4)
+        actions = congo_same_diff.first_round(experiment)
+        assert len(actions) >= 1
+        assert isinstance(actions[0], PlaylistAction)
+        assert isinstance(actions[1], Explainer)
+
+    def test_next_round_final_round(self):
+        congo_same_diff, session = create_context('final')
+
+        session.get_next_round = lambda: 6
```
BeritJanssen, line 53: Clever! Never occurred to me that this was a possibility to test what happens after a number of rounds have passed in unit tests.

backend/experiment/rules/tests/test_congosamediff.py
```diff
@@ -0,0 +1,115 @@
...
+        congo_same_diff, session = create_context('final')
+
+        final_action = congo_same_diff.get_final_round(session)
+
+        assert isinstance(final_action, Final)
+        assert final_action.final_text == 'Thank you for participating!'
+
+
+def create_context(name):
+    playlist = PlaylistModel.objects.create(name=f'CongoSameDiff{name}')
```
BeritJanssen, line 95: Why create different playlist and experiment objects every time? I cannot see why the same playlist & experiment couldn't be reused over all tests.

backend/experiment/models.py
drikusroor: Yes :-( I'm importing and using the `Experiment` class in the CongoSameDiff ruleset, which caused the circular import issues. Do you know any other way to solve it?

backend/session/models.py
```diff
@@ -137,6 +137,9 @@ def section_from_any_song(self, filter_by={}):
         To ensure appropriate IP restrictions, most rules should use this
         method instead of operating on the playlist directly.
         """
+        if self.playlist is None:
```
BeritJanssen, line 140: There are a couple of experiments which "cheat" a bit in that they make a query over the whole database to pull in specific sections. While I agree with enforcing a playlist on the session, we should check that acceptance / production databases don't contain experiments without associated playlists.

BeritJanssen (approved):
Nice! See my remark about that experiments may not have a playlist associated (I don't think that's the case, but the playlist *can* be null).

backend/session/models.py
drikusroor: Maybe we can think about some properties on the base ruleset like 'playlist_required = False' False by default but can be set to True in this specific ruleset. And we could do a check with that property in mind. What do you think? I think we already had an issue about validation of experiments which is kind of related.

backend/experiment/rules/tests/test_congosamediff.py
drikusroor: I think I had some problems with them interfering with each other. Don't remember exactly but I will check & see what happens if I put it back into the setupTestData method

backend/experiment/models.py
BeritJanssen: No, I was just curious why we didn't have this problem before. The answer is: we didn't add type hints before.

backend/experiment/rules/tests/test_congosamediff.py
BeritJanssen: Ah, that could also be solved by using Django's `setUp` method instead of the (class) `setUpTestData` method, see [documentation](https://docs.djangoproject.com/en/5.0/topics/testing/overview/#writing-tests). I forgot why we used the `setUpTestData` method in the first place, perhaps to speed up tests. `setUp` should destroy the objects after every test, and reconstruct them for the next.

backend/session/models.py
BeritJanssen: I think we can document & enforce that experiments should have playlists associated, no need to create exceptions on the rule level as far as I can see. At some point, `experiment.playlist` was set to `blank=True`, which was mainly a result of mild to severe annoyance when filling in an experiment admin form, only to be bounced back because the playlist wasn't filled. Perhaps we could actually improve this bit: instead, auto-fill the first playlist in the database, and warn the user that they should revisit the playlist(s) associated with the experiment.

backend/experiment/rules/tests/test_congosamediff.py
drikusroor: In the end I didn't really need the `create_context` method and it worked with just using the `setupTestData` method.

backend/session/models.py
drikusroor: Perhaps it would be nice to discuss this in a separate issue. I do think it would be good to warn the (admin) user that a Playlist doesn't exist or is empty before launching the experiment. I don't know what's the best place and way to do that. I do know that it took me some debugging before I realised that this was the reason my experiment wasn't working correctly as the frontend just shows a generic error message (could not load first round or something). I believe in throwing exceptions early and often. Some experiments need playlists with sections (sometimes with sections with certain tags or groups, or a minimum amount), so I would assume the admin user would be notified of this in the process of creating or updating an exeriment. At the same time, a playlist might get deleted, and in that scenario I would expect the API to throw an error that gets logged somewhere (frontend, backend, Sentry).

PR #810, opened 2024-02-28
Add theme support to Experiment model and components

### LINKED ISSUE #805: Configure custom logo, background and font per experiment
Split off from #666 and #293

Right now, there is no way to configure a custom background, logo and font per experiment. It would be nice to be able to do that, either to give experiments a specific branding or to be able to use specific fonts for languages that are not supported with the default font (Chinese?).

A configurable theme that has the following fields:
- name (required)
- description (optional)
- logo
- background
- font (google font)

Along with some previews for the logo, background and font.

Additionally, one should be able to select a theme from the experiment. If nothing is selected, default to the original theme that (e.g. no theme).

**Additional context**
- Colors / design tokens are split off to a new issue as well (to be created still)

drikusroor:
This pull request adds theme support per Experiment. It gives the admin user the ability to create and select a theme for an experiment, in which it can configure a logo image, background image, heading font, and body font).

Right now, this means the user can configure images and fonts by inputting image and google font urls. Later, we can expand this functionality by provoding upload features.

Resolves #805

### Theme configuration add/edit

image

image

### Themes admin overview

image

### Select theme in Experiment

image


BeritJanssen:
Should / could we also set the favicon through the admin interface?

drikusroor:
> Should / could we also set the favicon through the admin interface?

That sounds like a good idea. I prefer not to have scope creep in this PR so I've created issue #819 for it.

backend/theme/migrations/0001_initial.py
```diff
@@ -0,0 +1,26 @@
...
+        migrations.CreateModel(
+            name='ThemeConfig',
+            fields=[
+                ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('name', models.CharField(default='Default', max_length=255, unique=True)),
+                ('description', models.TextField(blank=True, null=True)),
+                ('font_url', models.CharField(blank=True, max_length=255, null=True)),
+                ('logo_url', models.CharField(blank=True, max_length=255, null=True)),
+                ('background_url', models.CharField(blank=True, max_length=255, null=True)),
+                ('active', models.BooleanField(default=True)),
```
BeritJanssen, line 23: What is the `active` field for?

backend/theme/tests.py
BeritJanssen, line 1: This file can probably go, right?

backend/theme/views.py
BeritJanssen, line 1: This file, too, can probably be removed?

BeritJanssen (commented):
I like it! The only thing that's not clear to me: what would happen if a user set `example.com/not-a-font.jpg` in the font form fields? Silent fail?

drikusroor:
> I like it! The only thing that's not clear to me: what would happen if a user set `example.com/not-a-font.jpg` in the font form fields? Silent fail?

It's possible to enter an incorrect font or image url that will silently fail in the frontend. However, thanks to the preview in the Django interface that shouldn't happen as you can see when it fails.

PR #812, opened 2024-02-28
Remove docker-compose-manage.yaml and add gettext package to DockerfileDevelop

### LINKED ISSUE #802: Cleanup: Remove docker-compose-manage.yml
Right now, the codebase contains the `docker-compose-manage.yml`, which appears to be or have been used to execute Django management commands. However, we now have the `scripts/manage` script for that, making this file obsolete.

We remove this `docker-compose.manage.yml` file and the `backend/DockerfileManage` file it is dependent on, but keep the dependencies from `backend/DockerfileManage` that are not present in `backend/Dockerfile` and `backend/DockerfileDevelop`, where needed.

The alternative is to keep these docker manage files but remove the `scripts/manage` script. I personally don't like this approach because I think it's better if Django's management commands are executed in the exact same setup as the one the actual backend application is running in. Less chance for weird bugs because two environments are not 100% the same.

**Additional context**
Drikus:
> Do any of you still use docker-compose-manage.yaml or can we remove [that file](https://github.com/Amsterdam-Music-Lab/MUSCLE/blob/develop/docker-compose-manage.yaml)?
Berit:
> This file has its own backend Dockerfile. The only difference is that this one installs gettext, which is used to generate and compile the translation files in backend/locale. Originally, the ffmpeg dependency was also exclusive to DockerfileManage, then it made more sense to keep them separate. Now I wonder whether we might not just as well move gettext over to DockerfileDevelop.

drikusroor:
This pull request removes the docker-compose-manage.yaml file and its related DockerfileManage, and adds the gettext package to the DockerfileDevelop. This change improves the development environment by removing unnecessary files and adding a required package.

Resolves #802


BeritJanssen (approved):

BeritJanssen:
I believe some documentation in the wiki still refers to the manage image, should adjust this, too.

PR #813, opened 2024-02-29
Fixed(CI): Fix podman workflow on PRs

drikusroor:
Prevent the podman workflow from always running (and thereby _deploying_ to ACC) on pull requests, which was still the case up to now.


BeritJanssen (approved):
See also https://github.com/Amsterdam-Music-Lab/MUSCLE/pull/814.

PR #815, opened 2024-03-02
Added: Enforce experiment slugs to be lowercase and disallow reserved keywords

### LINKED ISSUE #773: Introduce validation for `experiment.slug`
**Describe the bug**
This is more a bug waiting to happen:

**To Reproduce**
Create an experiment from the admin interface. Give it the slug "admin". Now, when you go to "localhost:3000/admin", you won't get to see your experiment, but instead the admin interface.

**Expected behavior**
* You shouldn't be able to create slugs which conflict with other routes of the interface.
* One additional aspect we might tackle when handling this issue: migrate the field from a `CharField` to a  [SlugField](https://docs.djangoproject.com/en/4.2/ref/models/fields/#slugfield). That means that only letters, numbers, underscores or hyphens are allowed, and will prevent somebody to accidentally type special characters into their slug.
* As discussed in the team meeting of February 27, 2024, we also want to enforce slugs to be lowercase.

drikusroor:
This pull request converts the experiment slug column from CharField to SlugField and adds a validator to enforce lowercase slugs and disallow reserved words.

Resolves #773


BeritJanssen (commented):
Without testing this locally: does the validation actually already give hints as you type? Or would we need to add some comment about allowed characters, and that we enforce lowercase, next to the slug form field in the admin interface?

drikusroor:
> Without testing this locally: does the validation actually already give hints as you type? Or would we need to add some comment about allowed characters, and that we enforce lowercase, next to the slug form field in the admin interface?

Yes, this works out of the box with Django :-)

image
image

BeritJanssen:
Noice!

BeritJanssen (approved):

PR #820, opened 2024-03-05
Hotfix/tunetwins play again

BeritJanssen:
Just realized that TuneTwins is broken in its current version: you cannot start the game if you click "start again" now. This hotfix is supposed to be merged into develop and main.


backend/experiment/rules/matching_pairs.py
```diff
@@ -80,14 +80,14 @@ def next_round(self, session):
                 title='Score',
                 final_text='Can you score higher than your friends and family? Share and let them try!',
                 button={
-                    'text': 'Play again',
+                    'text': 'Play again', 'link': '{}/{}'.format(
+                        settings.CORS_ORIGIN_WHITELIST[0], session.experiment.slug)
```
drikusroor, line 84: What does this mean and why does it need a cors origin whitelist setting entry?

drikusroor (approved):

backend/experiment/rules/matching_pairs.py
BeritJanssen: This just so happens to be a place in the settings where we've saved the url of the frontend. Perhaps it would indeed be clearer to make a dedicated setting variable instead.

PR #836, opened 2024-03-06
Added: Add Image model and admin functionality

### LINKED ISSUE #825: Add image model with upload function
## Requirements

- In Django, add a model with the following properties:
  - src (FileUpload field, I assume)
  - alt (text input)
  - href (url/link input)
  - rel (text input or multi select with the [official rel values](https://www.w3schools.com/tags/att_a_rel.asp))
  - target (select with options null/None or _blank. we could add more options but probably not necessary. see also the [target attribute docs](https://www.w3schools.com/tags/att_a_target.asp))
  - tag (or tags, perhaps separated by commas/colons)

## Context (goal behind goal)

This image model will (eventually) be used for several features in the nearby future such as:
- Experiment collection footer images #826 
- Theme images #822
- Experiment image #831
- Visual Matching Pairs
- Dynamic OG image in #834

drikusroor:
This pull request adds the Image model to the backend/image app, along with the necessary admin functionality. It also includes tests for the ImageAdmin and ImageModel.

Resolves #825


backend/image/migrations/0001_initial.py
```diff
@@ -0,0 +1,31 @@
...
+            name='Image',
+            fields=[
+                ('id', models.BigAutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('file', models.ImageField(upload_to='images/')),
+                ('title', models.CharField(max_length=255)),
+                ('description', models.TextField()),
+                ('alt', models.CharField(max_length=255)),
+                ('href', models.URLField()),
+                ('rel', models.CharField(max_length=255)),
+                ('target', models.CharField(max_length=255)),
```
Evert-R, line 25: I think it would be safer/easier to override the `target`  field in the form with a choicefield, as there are only a few valid options.

backend/image/models.py
```diff
@@ -0,0 +1,17 @@
+from django.contrib.postgres.fields import ArrayField
+from django.db import models
+
+
+class Image(models.Model):
+    file = models.ImageField(upload_to='images/')
```
Evert-R, line 6: Perhaps it's better to add a subfolder here for better overview, for example:  `uploads/%Y/%m/%d/`
This will also allow multiple files with the same name, while keeping the original filename in place.

Evert-R (approved):

backend/image/migrations/0001_initial.py
drikusroor: Good point! I actually thought about that before ... and then I forgot about it again 😄

backend/image/models.py
drikusroor: Also a good point. I've updated the upload path.

PR #837, opened 2024-03-07
Feature er/update section

### LINKED ISSUE #709: More conservative handling of playlist changes
Currently, sections will be deleted if its filename is replaced. This means that if we change the place where a file is hosted, results from sessions before this change will have their `section` field set to `null`. We should overhaul the logic of `section.models`, starting around l. 124.

Evert-R:
This PR introduces a `url_prefix` field on `Section`

This allows us to point a playlist to an external server during a live experiment without altering the playlist.
When `url_prefix` is left empty the local server will be used as usual. 
When filled with a valid url (that means starting with `http://` or `https://`) `get_section` will point to `<url_prefix>/<section.filename>`
The last `/` will be added to the `url_prefix` if it's not there.

closes #709


backend/experiment/actions/playback.py
```diff
@@ -49,6 +49,8 @@ def __init__(self,
                          for s in sections]
         if str(sections[0].filename).startswith('http'):
```
BeritJanssen, line 50: I think this if condition is obsolete? The filename itself will not start with `http` anymore, right? Or did you leave this for backwards compatibility with playlists where the section csv has been altered?

backend/section/forms.py
```diff
@@ -13,16 +13,32 @@ class AddSections(forms.Form):
...
+                               )
 
     class Meta:
         model = Playlist
         fields = '__all__'
+        help_texts = {
+            'url_prefix': 
+                'URL for hosting the audio files on an external server.<br> \
+                Make sure the folder structure matches \
+                that of the local server.<br>\
```
BeritJanssen, line 36: I think this docstring may be really confusing in case users choose to host from a different server from the start. Perhaps you can better put "Make sure the path of the audio file is valid"?

backend/section/models.py
```diff
@@ -28,6 +32,8 @@ def save(self, *args, **kwargs):
         """Update playlist csv field on every save"""
         if self.process_csv is False:
             self.csv = self.update_admin_csv()
+        if self.url_prefix != '' and self.url_prefix[-1] != '/':
```
BeritJanssen, line 35: more readable: "self.url_prefix.endswith('/')

backend/section/views.py
```diff
@@ -26,6 +26,10 @@ def get_section(request, section_id, code):
         if str(section.filename).startswith('http'):
             # external link, redirect
             return redirect(str(section.filename))
+        
+        if section.playlist.url_prefix != '':
```
BeritJanssen, line 30:
```suggestion
        if section.playlist.url_prefix:
```

BeritJanssen (approved):
Nice! I do wonder though whether we should perhaps just edit the playlists with filenames pointing to remote servers, so we can get rid of the `if filename.startswith('http')` conditions in `backend/experiment/actions/playback` and `backend/section/views`, as these will be confusing to future maintainers. Can also be filed as a separate issue though.

backend/experiment/actions/playback.py
Evert-R: Yes, I left this in for backwards compatibility, but we should get rid of this.

Evert-R:
> so we can get rid of the `if filename.startswith('http')` conditions in `backend/experiment/actions/playback` and `backend/section/views`, as these will be confusing to future maintainers. Can also be filed as a separate issue though.

Yes, we should get rid of those. I will make an issue for editing the playlists.

PR #843, opened 2024-03-12
Fix: refactor MatchingPairs

### LINKED ISSUE #796: Refactor VisualMatchingPairs
### Discussed in https://github.com/Amsterdam-Music-Lab/MUSCLE/discussions/775

<div type='discussions-op-text'>

<sup>Originally posted by **BeritJanssen** February 19, 2024</sup>
@drikusroor , I've been wondering if we can refactor `VisualMatchingPairs` so it can actually be run through the `MatchingPairs` component instead, bringing down the number of components, and also the amount of verbatim repeated code in the frontend. Differences between the visual and audio version
- different Preload (this is controlled through the `play_method`, so that's already taken care of)
...

Can you specify more differences? Or provide comments for any difficulties you might foresee? I'm asking because in this case we could also bring back the rules files to just the one ("Fieldwork") one, of which we could make different experiments in the admin interface, one of which would be coupled to a "Playlist" of image files.</div>

### LINKED ISSUE #817: Congo matching pairs loads a wrong number of tiles
**Describe the bug**
the game loads the wrong number of cards

**To Reproduce**
It is random behavior showing on Toughbook, the game could load with correct number of cards(16), but also loads with 8 or 32 cards. 

**Expected behavior**
Every time we load the game, it should show the expected number of cards which is 16.

**Screenshots**
IMG_0387

 - OS: Windows
 - Browser: Firefox 123.0 / Edge 122.0.2365.59

**Additional context**
the 16 cards are compressed by two sets of 8 stimuli, meaning 8 matches/pairs. In the original condition of the game, the 2 sets of cards are identical, but in the degradation conditions, the first set of 8 cards are original stimuli, and the second set of 8 cards are degraded stimuli. Now, the degradation board will also load the same sets of stimuli occasionally, which means there are no degradation stimuli loaded, but only two sets of original stimuli.

### LINKED ISSUE #818: Congo matching pairs loads wrong set of stimuli
**Describe the bug**
The 16 cards of a matching pairs board are compressed by two sets of 8 stimuli, meaning 8 matches/pairs. In the original condition of the game, the 2 sets of cards are identical, but in the degradation conditions, the first set of 8 cards are original stimuli, and the second set of 8 cards are degraded stimuli. Now, the degradation board will also load the same sets of stimuli occasionally, which means there are no degradation stimuli loaded, but only two sets of original stimuli. This behavior happens randomly when refresh the browser with a degradation game link, sometimes it loads both original and degraded stimuli, sometimes it is two sets of original stimuli, it does not seem to have something to do with Cache. When the cache is cleaned, this bug still appears. 
 

**To Reproduce**
I don't know if it happens on Mac, with the tough book, if you go to any degraded link, and load the game, it can happen anytime. 

**Expected behavior**
The board will load the correct sets of stimuli in the degradation condition, meaning loading 1 set of original stimuli and 1 set of corresponding degraded stimuli to form the game board. 

**Screenshots**

-OS: Windows
-Browser: Firefox 123.0 / Edge 122.0.2365.59

Context 
- see #817

BeritJanssen:
Close #796, close #818, close #817 . This PR merges the `VisualMatchingPairs` and `MatchingPairs` component. All tests but 'it ends the game when all pairs are matched' pass in isolation - it seems there are some side effects of the tests using `mockAxios`, which somehow cannot be cleared by `jest.resetAllMocks()` in the `beforeEach` function. This is why three tests are skipped at the moment, after 4+ hours on this I decided to "jest get on with it". Also worked towards applying recommendations from [this article](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library) to the tests.


frontend/src/components/MatchingPairs/MatchingPairs.test.js
```diff
@@ -0,0 +1,156 @@
...
+        expect(cards[2]).toBeDisabled();
+        
+        fireEvent.click(cards[1]);
+        fireEvent.click(cards[3]);
+        fireEvent.click(screen.getByTestId('overlay'));
+        await screen.findByText('Score: 120');
+        expect(cards[1]).toBeDisabled();
+        expect(cards[3]).toBeDisabled();
+        expect(submitResult).toHaveBeenCalled();
+    });
```
drikusroor, line 118: Did these tests pose problems?

frontend/src/components/MatchingPairs/MatchingPairs.test.js
```diff
@@ -0,0 +1,156 @@
...
+    beforeEach(() => {
+        jest.resetAllMocks();
+    });
+    
+    const mockSections = [
+        { id: 1, content: 'Card 1', url: '/cat-01.jpg', inactive: false, turned: false, noevents: false, seen: false, group: 1 },
+        { id: 2, content: 'Card 2', url: '/cat-02.jpg', inactive: false, turned: false, noevents: false, seen: false, group: 2 },
+        { id: 3, content: 'Card 1', url: '/cat-01.jpg', inactive: false, turned: false, noevents: false, seen: false, group: 1 },
+        { id: 4, content: 'Card 2', url: '/cat-02.jpg', inactive: false, turned: false, noevents: false, seen: false, group: 2 },
+    ];
```
drikusroor, line 31: I see these tests mainly test the visual variant of the game. Should we have tests for the audio variant too?

backend/experiment/actions/utils.py
```diff
@@ -96,3 +96,12 @@ def get_last_n_turnpoints(session, num_turnpoints):
...
     return silence_time + (random.uniform(min_jitter, max_jitter) if not continuation_correctness else 0)
+
+
+def is_audio_file(section):
+    from section.validators import audio_file_validator
+    try:
+        audio_file_validator(str(section.filename))
+        return True
+    except:
+        return False
```
drikusroor, line 107: I would expect a (Django) validator to be used in a form and not as a utility function that is used in the Playback class. You could also make it as easy as this and you wouldn't need to use a try-catch.

```py
audio_extensions = ['.wav', '.mp3', '.aac', '.ogg', '.flac']

def is_audio_file(section):
    filename = section.filename
    return any(filename.lower().endswith(ext) for ext in audio_extensions)
```

frontend/src/components/MatchingPairs/MatchingPairs.js
```diff
@@ -18,75 +18,66 @@ const MatchingPairs = ({
...
+                    setFeedbackClass('');
+                    setFeedbackText('');
             }
-
-
-            // add third click event to finish the turn
-            document.getElementById('root').addEventListener('click', finishTurn);
+            setFeedbackClass(fbclass);
+            turnedCards[0].matchClass = turnedCards[1].matchClass = fbclass;
+            turnedCards[1].seen = turnedCards[1].seen = true;
```
drikusroor, line 79: I see you are setting properties directly on objects in an array that are defined within this function. Does that actually do something? The values of the properties are probably lost after the execution of the function / after a re-render, or not?

drikusroor (approved):
LGTM! have you tested both variants locally?

frontend/src/components/MatchingPairs/MatchingPairs.js
BeritJanssen: Good point, I should check that. From manual testing, it seems like this works. The matchClass is something that will only be used for an animation, so loosing that information wouldn't be problematic. `seen`, however, would.

frontend/src/components/MatchingPairs/MatchingPairs.test.js
BeritJanssen: Yes, see description of the PR - `mockAxios` doesn't seem to reset properly between tests.

frontend/src/components/MatchingPairs/MatchingPairs.test.js
BeritJanssen: I don't think it matters what the sections look like, as we aren't testing the actual audio playback anyway.

backend/experiment/actions/utils.py
BeritJanssen: I thought to reuse the validator as it's already there, but you are right, it's a bit weird, and also doesn't number the number of lines added. I also should still be writing a unit test for the part of the playback that determines the `play_method`.

PR #858, opened 2024-03-14
Add validator

Evert-R:
This PR adds the migration for the validator of `url_prefix' on the `Section` model


drikusroor (approved):
Looks good to me 🚀

PR #863, opened 2024-03-15
Fixed: Remove incorrect import

drikusroor:
This import would cause the admin dashboard to throw an error, thereby making it impossible to update experiment series.


BeritJanssen:
This import is already removed on develop, as far as I can see?

drikusroor:
> This import is already removed on develop, as far as I can see.

As this commit has already been deployed to production last Friday, should we merge this PR and mark it as a patch (`v1.9.3`) release?

BeritJanssen:
Sounds good to me!

PR #867, opened 2024-03-18
feat: introduce fixed order board for practice purposes

### LINKED ISSUE #832: Add two more buttons on Dashboard for Instruction phase for Congo Matching Pairs
Now the stimuli position on the board are randomized each time a new board is loaded. But during the instruction phase, it would be hard for the experimenter to demonstrate the game and at the same time, do the actual task trying to find the matching pairs. In addition, if we couldn't find a pair in the first few trials, we might lose the attention of the testees. 

It would be ideal if we could have two extra buttons on the dashboard for the instructional version of matching pairs: one of them is original matching with the original (O). The other one is original matching with the degraded (D). 
In the instructional version, ensure that the position of the stimuli remains fixed for each board. This will allow the experimenter to demonstrate the game without having to search for pairs themselves.

BeritJanssen:
Close #832: add a MatchingPairs ruleset in which the board is not randomized, but fixed, to be used for creating two experiments for practice.

### DIFF
--- a/backend/experiment/rules/matching_pairs.py
+++ b/backend/experiment/rules/matching_pairs.py
@@ -19,6 +19,7 @@ class MatchingPairsGame(Base):
     show_animation = True
     score_feedback_display = 'large-top'
     contact_email = 'aml.tunetwins@gmail.com'
+    randomize = True
 
     def __init__(self):
         self.questions = [
@@ -79,8 +80,8 @@ class MatchingPairsGame(Base):
                 title='Score',
                 final_text='Can you score higher than your friends and family? Share and let them try!',
                 button={
-                    'text': 'Play again', 'link': '{}/{}'.format(
-                        settings.CORS_ORIGIN_WHITELIST[0], session.experiment.slug)
+                    'text': 'Play again', 'link': '/{}'.format(
+                        session.experiment.slug)
                 },
                 rank=self.rank(session, exclude_unfinished=False),
                 social=social_info,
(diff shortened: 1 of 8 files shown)


backend/experiment/rules/matching_pairs.py
```diff
@@ -79,8 +80,8 @@ def next_round(self, session):
                 title='Score',
                 final_text='Can you score higher than your friends and family? Share and let them try!',
                 button={
-                    'text': 'Play again', 'link': '{}/{}'.format(
-                        settings.CORS_ORIGIN_WHITELIST[0], session.experiment.slug)
+                    'text': 'Play again', 'link': '/{}'.format(
+                        session.experiment.slug)
```
drikusroor, line 84:
```suggestion
                    'text': 'Play again',
                    'link': f'/{session.experiment.slug}'
```

Perhaps this would make it a bit more readable

backend/experiment/rules/tests/test_matching_pairs_fixed.py
```diff
@@ -0,0 +1,44 @@
...
+            playlist=self.playlist
+        )
+        first_run_sections = session.experiment_rules().select_sections(session)
+        another_session = Session.objects.create(
+            experiment=self.experiment,
+            participant=self.participant,
+            playlist=self.playlist
+        )
+        second_run_sections = another_session.experiment_rules().select_sections(another_session)
+        assert first_run_sections == second_run_sections
```
drikusroor, line 44: Should we also test the randomisation part for the `_lite` experiment?

drikusroor (approved):
LGTM 🚀  Thinking out loud: It would be cool to be able to set this in the admin dashboard instead of having to create a separate class for it

BeritJanssen:
> It would be cool to be able to set this in the admin dashboard instead of having to create a separate class for it
Totally agree, but I don't think we'll get to that detail of control over experiments via the admin interface.

PR #869, opened 2024-03-19
Added: Add useDisableRightClickOnTouchDevices hook

### LINKED ISSUE #857: Disable rightclick on Matching Pairs
On the toughbooks touchscreen `press and hold` will open a context menu. (Equals a right mouseclick)
Should we turn this off with javascript? and if so:  Only for matching pairs or on all experiments?

drikusroor:
This pull request adds the `@testing-library/react-hooks` dependency and includes a new hook called `useDisableRightClickOnTouchDevices`. The hook prevents the context menu from appearing on touch devices.

Resolves #857


Evert-R:
Nice!

PR #870, opened 2024-03-19
Save `continuation_correctness` to session json_data

### LINKED ISSUE #851: continuation_offset is missing from SONG_SYNC export
In the older `song_sync` JSON export, we had access to a `continuation_offset` field, which stored the skip time in the audio file between the correct place and the continuation as played to the user during the `correct_place` question. In the newer `recognize` and `correct_place` export, this field seems to be missing. Although we do not use this field for any current analyses, I think it is important to have it available. Let's bring it back!

BeritJanssen:
close #851 .

### DIFF
--- a/backend/experiment/actions/wrappers.py
+++ b/backend/experiment/actions/wrappers.py
@@ -81,7 +81,7 @@ def song_sync(session: Session, section: Section, title: str,
         },
         title=title
     )
-    continuation_correctness = random.randint(0, 1) == 1
+    continuation_correctness = random.choice([True, False])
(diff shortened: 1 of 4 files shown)


backend/experiment/actions/wrappers.py
```diff
@@ -82,6 +82,9 @@ def song_sync(session, section, title,
         title=title
     )
     continuation_correctness = random.randint(0, 1) == 1
```
albertas-jn, line 84: `random.choice([True, False])`

albertas-jn (approved):

PR #872, opened 2024-03-19
Feat: display number of sessions

### LINKED ISSUE #799: Finished-Trial-Count per condition per person for Congo matching pairs
When conducting experiments in Congo the "laboratory" can be quite chaotic. To ensure that we could evenly test different conditions of the game. We would like to have a feature showing the number of boards finished per condition per person.  

Terminology
Condition: A version of an experiment (e.g. Button spectral degradation 1 or Button spectral degradation 2)
A completed board: one board of 8 pairs being successfully matched up. 
 
Flow
Navigate to the individual dashboard through providing ID in the URL
Start the experiment with selecting any button
After each board of 8 pairs being completed, after every board is finished, one will be directed back to the dashboard.
Then start another board from there. 

As I understand it, now we would need to put the ID in the URL and then all the game/data as well as a dashboard would be linked to this person. We have now 9 buttons in total, including 7 different conditions of auditory matching pairs game:  Button Visual Matching Pairs (1), Button Musicality Battery(2), Button Original(3), Button Spectral degradation (4),  Button Spectral degradation 2(5), Button Pitch degradation 1(6), Button Pitch degradation 2(7), Button Temporal degradation 1(8), Button Temporal degradation 2 (9). 

So it would be ideal to see the number updated in the corresponding condition button on the dashboard after every board is finished. So the number will be updated and accumulate after each board is finished. So that we could have an overview of how many times a particular participant has finished a board under each condition of the game. 

The picture shows the hypothetical current status where someone has finished 1 visual matching pairs board, 1 musicality battery test, 2 original condition matching pairs board, 1 TD1 board, 2 TD2 boards, and haven't played other conditions as yet (can show 0 instead of nothing). 

IMG_0368

BeritJanssen:
Close #799 . This branch introduces a counter which indicates how often a given game in an `ExperimentCollection` has been played.

Screenshot 2024-03-18 at 17 09 38

Right now, that means all `ExperimentCollection` dashboards display this: admin infrastructure to customize this pending (#873 ).


frontend/src/components/ExperimentCollection/ExperimentCollection.scss
```diff
@@ -220,7 +220,14 @@
                         padding: 0 15px 15px;
                         opacity: 0.8;
                     }
+                    .counter {
+                        color: white;
+                        position: relative;
```
drikusroor, line 225: Interesting, I would expect the counter class to have `position: absolute;` and the experiment tile to have `position: relative;`. 🤔  But if it works, it works.

frontend/src/components/ExperimentCollection/ExperimentCollection.scss
```diff
@@ -220,7 +220,14 @@
...
+                        top: 30vh;
```
drikusroor, line 226: Isn't `vh` a height relative to the viewport? What happens if you resize the screen?

frontend/src/components/ExperimentCollection/ExperimentCollection.scss
```diff
@@ -220,7 +220,14 @@
...
+                        left: 90%;
```
drikusroor, line 227: You could also use `right` here as well. Additionally, do you want the horizontal positioning to be dependent on the width of its container, or do you want it to be a constant distance to the right (or left?) edge of the container? With things like these (items positioned on the bottom right of a container), I would normally do it something like this:

```css
.container {
  position: relative;
  /* etc. */
}

.count {
  position: absolute;
  bottom: 1rem;
  right: 1rem;
}
```

frontend/src/components/ExperimentCollection/ExperimentCollection.js
```diff
@@ -6,10 +6,9 @@ import Loading  from "../Loading/Loading";
 const ExperimentCollection = ({match}) => {
     const [experimentCollection, loadingExperimentCollection] = useExperimentCollection(match.params.slug);
     const dashboard = experimentCollection?.dashboard;
-    const experiment = dashboard ? undefined : experimentCollection;;
+    const experiment = dashboard ? undefined : experimentCollection;
```
drikusroor, line 9: It's only the removal of the semicolon, but what does this expression actually do? 😅 If dashboard is truthy then experiment should be undefined and otherwise it should be a collection? 🤔

backend/experiment/views.py
```diff
@@ -125,12 +125,9 @@ def serialize_experiment(experiment_object, finished=False):
...
     ''' Check if there is a finished session for this experiment and participant '''
-    try:
-        Session.objects.get(
-            experiment=experiment, participant=participant, finished_at__isnull=False)
-        return True
-    except:
-        return False
+    count = Session.objects.filter(
+        experiment=experiment, participant=participant, finished_at__isnull=False).count()
+    return count
```
drikusroor, line 130: As it returns a number now, I think it would be a good idea to update the name of the method as well. Perhaps something like `def get_finished_sessions_count`?

drikusroor (changes requested):
It's not an explicit request changes but I do think that 

- The css could be improved
- The finished session method could be renamed
- The experiment - undefined - collection expression could use a explanatory comment

frontend/src/components/ExperimentCollection/ExperimentCollection.js
BeritJanssen: I realized that this is a bit of overkill anyway. Added a docstring explaining the possible structures of the backend response, and refactored such that reasoning is over `experimentCollection` (i.e., the object resulting from that response), instead of derived objects.

Issue #877, opened 2024-03-19
Adjust admin interface to add `ExperimentSeriesGroup` and `GroupedExperiment`


BeritJanssen:
Depends on #875. After the data model has been adjusted, we want to still be able to set `ExperimentSeries` up through the admin interface.

PR #884, opened 2024-03-20
Added: Add ExperimentSeriesGroup and GroupedExperiment models

### LINKED ISSUE #875: New data models for ExperimentSeries
After discussion in #865 and #866, a suggestion to tackle both issues crystallized. We could achieve more stable database management and ordering by adding extra models:
```python
class ExperimentSeries(models.Model):
   slug = models.SlugField(max_length=64, default='')

class ExperimentSeriesGroup(models.Model):
   name = models.CharField()
   series = models.ForeignKey(ExperimentSeries)
   order = models.IntgerField()
   dashboard = models.BooleanField()
   randomize = models.BooleanField()

class GroupedExperiment(models.Model):
  experiment = models.OneToOneField(Experiment)
  group = models.ForeignKey(ExperimentSeriesGroup)
  order = models.IntegerField()
```

drikusroor:
This pull request adds the ExperimentSeriesGroup and GroupedExperiment models to the project. It also includes necessary migrations and updates the ExperimentSeriesForm to handle the new models.

It also contains some basic admin interface updates such as the form inputs required to CRUD ExperimentSeriesGroups and GroupedExperiments and related information displayed in the admin overview pages. Perhaps it's enough for now already, see also #877. (see screenshots below).

**⚠️ IMPORTANT: Don't merge this straight after #886 as the migrations will conflict ⚠️** Either update this PR's migration after merged the other PR, or merge this one and update the other PR's migration.

Resolves #875

## Experiment series overview

image

## Experiment series form

image

## Experiment series groups overview

image

## Experiment series group form

image


backend/experiment/models.py
```diff
@@ -44,6 +44,35 @@ def consent_upload_path(instance, filename):
     return 'consent/{0}/{1}'.format(folder_name, filename)
 
 
+class ExperimentSeriesGroup(models.Model):
+    name = models.CharField(max_length=64, blank=True, default='')
+    series = models.ForeignKey(ExperimentSeries, on_delete=models.CASCADE)
+    order = models.IntegerField(default=0)
```
BeritJanssen, line 50: I think we should impose a [unique_together](https://docs.djangoproject.com/en/5.0/ref/models/options/#unique-together)  constraint on order / name / series, such that we cannot have two `ExperimentSeriesGroup`s with the same order integer.

backend/experiment/models.py
```diff
@@ -44,6 +44,35 @@ def consent_upload_path(instance, filename):
...
+
+    class Meta:
+        ordering = ['order']
+        verbose_name_plural = "Experiment Series Groups"
+
+
+class GroupedExperiment(models.Model):
+    experiment = models.ForeignKey('Experiment', on_delete=models.CASCADE)
+    group = models.ForeignKey(ExperimentSeriesGroup, on_delete=models.CASCADE)
+    order = models.IntegerField(default=0)
```
BeritJanssen, line 66: Same as above: perhaps we can enforce that ordering integers are unique through `unique_together`.

backend/experiment/forms.py
```diff
@@ -119,6 +120,45 @@
...
+            help_text=('This field will be deprecated in the nearby future. '
+                       'Please use experiment series groups (see bottom of form).')
+        )
+        self.fields['last_experiments'] = ModelFormFieldAsJSON(
+            queryset=experiments,
+            required=False,
+            help_text=('This field will be deprecated in the nearby future. '
+                       'Please use experiment series groups (see bottom of form).')
+        )
+        self.fields['dashboard'].help_text = ('If you check "dashboard", the experiment collection will have a '
```
BeritJanssen, line 151: This field will also be deprecated.

backend/experiment/models.py
drikusroor: In principle I do agree, but supporting this is hard. You will probably run into problems as the items will be saved individually, so if you change the order of two groups (from GA 0 and GB 1 to GA 1 and GB 0), the db will complain that there is already an group (GB) with order number 1 when you try to save GA with number 1. So then you will have to do magic things in order to facilitate it like first saving GB temporarily with a very high or negative order number like 999 or -1, then save GA with order number 1, and then save GB with order number 0. Perhaps it would be enough to just warn the user in the admin interface? I mean, technically it doesn't break the application when two groups have the same order number, but you might get random orders depending on the request.

BeritJanssen (changes requested):
Awesome! Before merging though, we should perhaps think about enforcing that order integers are distinct (see my comments), as well as telling the user whether lower = first or higher = first, as this is not self-explanatory.

More importantly though: the current views still assume that we have fields called `first_experiments`, `random_experments` and `last_experiments` on `ExperimentSeries`. `backend/experiment/views` therefore is now not doing anything with the new groups that were introduced. Is this by intention, to be left for after data migration?

drikusroor:
> More importantly though: the current views still assume that we have fields called `first_experiments`, `random_experments` and `last_experiments` on `ExperimentSeries`. `backend/experiment/views` therefore is now not doing anything with the new groups that were introduced. Is this by intention, to be left for after data migration?

Yes, my plan was to handle that in a later PR, as to not make this PR too big. Issue #875 also specifically talks about models so I think a new issue could be created to update the logic and phase out the _first__, _random__ and __last_ fields.

backend/session/tests/test_views.py
```diff
@@ -1,8 +1,6 @@
-import json
-
 from django.test import TestCase
 
-from experiment.models import Experiment, ExperimentSeries
+from experiment.models import Experiment, ExperimentSeries, ExperimentSeriesGroup
```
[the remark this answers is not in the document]
drikusroor: Good point. I probably used it in an earlier version of the test and then realized I didn't need it anymore..

backend/experiment/models.py
BeritJanssen: Hm, that is a good point. In that case, consider the "changes requested" to be adding the help_text.

backend/experiment/forms.py
drikusroor: Is that because there will always be a dashboard for a collection?

BeritJanssen: The dashboard is going to be on the group level, not the series level. At least I cannot see a use case in which we would want to have a dashboard for the series *and* a group within the series.

drikusroor: I see. But the description, about page, logos, and so on will still be part of the experiment series, right?

Issue #824, opened 2024-03-05
Configure ExperimentCollection about page through admin


BeritJanssen:
- [x] Add property to experiment collection model for about page
- [x] Add input for about page
- [x] Add about page content to response body of get experiment collection endpoint

PR #885, opened 2024-03-20
Added: Experiment collection about page

drikusroor:
This PR adds a link to an experiment collection's about page if the about_content property on an experiment collection is set and adds and shows that about page as well. It splits off the dashboard page from the experiment collection page. The dashboard and about page are loaded through nested routes in the experiment collection component.

Additionally, this pull request renames all React component files in the repository to have a .jsx extension. This is done to explicitly have Vite regard them as jsx files, making it easier in the future to migrate to tsx (typescript).

Related to #824 (this is the frontend part of the solution)


frontend/src/components/ExperimentCollection/ExperimentCollectionDashboard/ExperimentCollectionDashboard.tsx
```diff
@@ -0,0 +1,60 @@
...
+        <>
+            <div className="hero">
+                <div className="intro">
+                    <p>{experimentCollection?.description}</p>
+                    <nav className="actions">
+                        {nextExperiment && <a className="btn btn-lg btn-primary" href={"/" + nextExperiment.slug}>Volgende experiment</a>}
+                        {aboutContent && <Link className="btn btn-lg btn-outline-primary" to={`/collection/${experimentCollection.slug}/about`}>Over ons</Link>}
+                    </nav>
+                </div>
+                <div className="results">
```
BeritJanssen, line 30: What's this empty div doing?

frontend/src/interfaces/ExperimentCollection.interface.ts
```diff
@@ -0,0 +1,13 @@
+import Experiment from "./Experiment.interface";
+
+export default interface ExperimentCollection {
+    id: number;
+    name: string;
+    description: string;
+    dashboard: Experiment[];
```
BeritJanssen, line 7: this can also be null, right?

frontend/src/components/ExperimentCollection/ExperimentCollectionDashboard/ExperimentCollectionDashboard.tsx
```diff
@@ -0,0 +1,60 @@
...
+    // TODO: get next experiment and about link from experimentCollection
+    const nextExperiment = experimentCollection.next_experiment; // TODO: get next_experiment from experimentCollection
+    const aboutContent = experimentCollection.about_content;
+
+    return (
+        <>
+            <div className="hero">
+                <div className="intro">
+                    <p>{experimentCollection?.description}</p>
+                    <nav className="actions">
```
BeritJanssen, line 25: Where has the redirect gone?

frontend/src/components/ExperimentCollection/ExperimentCollection.jsx
```diff
@@ -1,39 +1,39 @@
...
+ *       finished_session_count: number
+ *     }
+ *   ],
+ *   redirect_to: {
+ *     slug: string,
+ *     name: string,
+ *     finished_session_count: number
+ *   }
+ * }
+ * */
```
BeritJanssen, line 33: perhaps this docstring can be removed, as the interface documents this now.

BeritJanssen (changes requested):
Nice job on factoring out the dashboard!

Though I'm not sure if the functionality is still intact after this change: the redirect functionality seems to be gone (in case the ExperimentCollection receives `redirect_to`).

Also, my preference would be to collect the interfaces in
`models/Experiment.ts`, etc., i.e., without adding interface to the name. Since the .ts ending actually already clarifies what the file is about, and `models` top directory to also be able to collect types or models with utility functions here.

frontend/src/interfaces/ExperimentCollection.interface.ts
drikusroor: I don't think it can or should. If there's no dashboard it should be an empty array.

frontend/src/components/ExperimentCollection/ExperimentCollection.jsx
drikusroor: I could do this and maybe straight away convert the whole file to `.tsx`? Then I can type the props as well.

frontend/src/components/ExperimentCollection/ExperimentCollectionDashboard/ExperimentCollectionDashboard.tsx
drikusroor: It's in `ExperimentCollection.jsx` (this page's parent)

drikusroor:
> Nice job on factoring out the dashboard!
> 
> Though I'm not sure if the functionality is still intact after this change: the redirect functionality seems to be gone (in case the ExperimentCollection receives `redirect_to`).

As mentioned in one of the comments, the redirect component is still there so I would assume it to work still. Have you experienced it not working anymore? It could be a backend issue as well.

> Also, my preference would be to collect the interfaces in `models/Experiment.ts`, etc., i.e., without adding interface to the name. Since the .ts ending actually already clarifies what the file is about, and `models` top directory to also be able to collect types or models with utility functions here.

Although I'm not against that idea per se, it is fairly common to split up interfaces and models (as in classes) as interfaces are used only for typing and models have functionality as well. And then we haven't even considered what to do with TypeScript `type`s yet, which also sometimes get dedicated folders. 😅 Perhaps we could work with a hybrid form? Either use a `models` folder and add `.interface` or `.type` to the name of the file if it's not a class, or use `Experiment.ts` as name but put it into an `interfaces` or `types` folder?

frontend/src/components/ExperimentCollection/ExperimentCollectionDashboard/ExperimentCollectionDashboard.tsx
BeritJanssen: You mean this is a placeholder for the score? In that case, we probably want it to be called "score" rather than "results". But then it can be left for later.

PR #888, opened 2024-03-22
Chore(deps): Update dependencies and configuration files

drikusroor:
This pull request updates the yarnPath in .yarnrc.yml to version 4.1.1, updates vite and storybook dependencies in package.json, adds a bind mount for tsconfig.node.json in docker-compose.yaml, removes custom esbuild configuration in vite.config.ts and vitest.config.ts, refactors Storybook configuration to use TypeScript for loader and removes obsolete onboarding addon, and renames file extensions to jsx/tsx whenever these files use jsx/tsx syntax.

Main reason for doing this was because the Storybook build in the CI/CD kept failing. In the end the best solution was to update Storybook, get rid of custom build configurations for Storybook and Vite for `.js` files. - Vite normally interprets `.js` files as JavaScript only files, but our components, stories and tests all contain `jsx` syntax due to React and I had to add some special rules telling Vite to interpret `.js` files as `jsx` too. It's probably better to use the extension that signifies `jsx` usage (`.jsx`) for those files and use the default build configuration, so that's what I did here.


BeritJanssen (approved):
chore indeed...

PR #896, opened 2024-03-26
fix: use random seed before shuffle

BeritJanssen:
Address problem for Congo-MatchingPairs that the board used for the practice sessions should be fixed, but not too predictable.


Evert-R (approved):

drikusroor:
Would be nice if there had been a test that runs that randomisation method a couple of times to see if it stays in the same random order

PR #903, opened 2024-03-28
Added: Save trial group as question key in results

### LINKED ISSUE #901: [Same-diff] make sure the practice trial data and main experiment data are recorded accordingly
It would be ideal if there will be an extra column in the recorded data indicating which part of the data is from trial section or main section. 

**Additional context**
We would like to have 4 separate experiments: Rhythm/Melody/Timbre/Beat. Each of these four experiments will include the practice trial section (2 stimuli) and the main experiment section (6 stimuli).

drikusroor:
This pull request adds validity checks for experiment sections in the CongoSameDiff class. It ensures that all sections have a group value and at least one section has the tag "practice". Additionally, it includes the following commits:

- feat: Include group in trial key to be able to distinguish them
- feat: Add validity checks for experiment sections in CongoSameDiff class

Resolves #901

### DIFF
--- a/backend/experiment/rules/congosamediff.py
+++ b/backend/experiment/rules/congosamediff.py
@@ -25,6 +25,18 @@ class CongoSameDiff(Base):
         Consent and Playlist are often desired, but optional
         '''
 
+        # Do a validity check on the experiment
+        # All sections need to have a group value
+        sections = experiment.playlists.first().section_set.all()
+        for section in sections:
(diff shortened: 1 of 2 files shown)


BeritJanssen (approved):
Nice! Good idea to make the assumptions about the playlist that goes with an experiment explicit. I wonder if we can generalize this idea to improve the stability of other experiments, too.

PR #906, opened 2024-04-02
Feature/footer

### LINKED ISSUE #826: Configure footer through admin: ManyToMany Image field, short description

BeritJanssen:
Close #826 . This branch adds a `FooterConfig`, which has a `ForeignKey` to the `ThemeConfig`. That way, every `Experiment(Collection)` can have a footer with a disclaimer, logos and/or a privacy note.


backend/theme/admin.py
```diff
@@ -88,13 +96,13 @@ def body_font_preview(self, obj):
         return "No font selected"
 
     def logo_preview(self, obj):
-        if obj.logo_url:
-            return mark_safe(f'<img src="{obj.logo_url}" style="max-height: 50px;"/>')
+        if obj.logo_image:
+            return mark_safe(f'<img src="/upload/{obj.logo_image.file}" style="max-height: 50px;"/>')
```
drikusroor, line 100: I'm not sure if this is going to work with a variable server subpath like `/server/`

backend/theme/admin.py
```diff
@@ -88,13 +96,13 @@ def body_font_preview(self, obj):
...
         return ""
 
     def background_preview(self, obj):
-        if obj.background_url:
-            return mark_safe(f'<div style="background-image: url({obj.background_url}); height: 50px; width: 100px; background-size: cover;"></div>')
+        if obj.background_image:
+            return mark_safe(f'<div style="background-image: url(/upload/{obj.background_image.file}); height: 50px; width: 100px; background-size: cover;"></div>')
```
drikusroor, line 105: Same here, you probably have to insert something like the `SUBPATH` from `base_settings.py`

backend/theme/static/img_preview.js
drikusroor, line 1: Wouldn't it be nice to still have this? Maybe only for the logo and not for the background and footer images?

backend/theme/tests/test_models.py
```diff
@@ -1,31 +1,82 @@
...
         theme_config = ThemeConfig.objects.get(name='Default')
         self.assertEqual(str(theme_config), 'Default')
 
+    def test_footer_to_json(self):
+        expected_json = {
+            'disclaimer': 'Some [more information][https://example.com/our-team]',
+            'logos': [],
+            'privacy': ''
+        }
+        self.assertEqual(self.footer.to_json(), expected_json)
```
drikusroor, line 39: I think it would be nice to test a footer to json _with_ images as well

drikusroor (approved):
LGTM!

backend/theme/static/img_preview.js
BeritJanssen: The image preview is now handled via the `image_preview` function from `image.admin`, so I thought it wasn't necessary to keep this. I may be overlooking something though.

backend/theme/admin.py
BeritJanssen: For the admin interface, this works like a charm.

drikusroor: But also on ACC & TST, where the path might be `/server/upload`?

backend/theme/admin.py
BeritJanssen: You are right though, we can override the `MEDIA_ROOT`, so this should be here instead.

backend/theme/admin.py
BeritJanssen: Will check!

PR #907, opened 2024-04-02
Bugfix er/edit sections

### LINKED ISSUE #905: Edit sections doesn't save artist/name (song) if only one of the two is filled out

Evert-R:
Changes and fixes in this PR:
- Introduction of a central function for retrieval or creation of the song model.
- A song will now be created if either a name or artist has been filled out. 
- `process_csv` will always be set to `False` on the `Playlist` object after adding sections through the admin interface (bug)

resolves #905


backend/section/admin.py
drikusroor, line 1: Would it be nice for this to have a unit test to prevent regression whilst also documenting expected behaviour?

BeritJanssen (approved):
A test is indeed a good idea, otherwise LGTM.

Issue #830, opened 2024-03-05
Select an experiment to be run when clicking "next experiment" button


BeritJanssen:
## Requirements

 - [x] Feature that selects an experiment from the experiment collection that hasn't been done yet by the participant. Reuse logic from ToontjeHoger.
	- TBD: Pick random or from a pre-configured experiment order in the collection
- [x] Click Button makes request to API. Reponse is an id and/or slug that will be used by frontend to redirect to the experiment's slug.

PR #908, opened 2024-04-02
Feature/serialize experiment collection

### LINKED ISSUE #904: Backend changes to select & serialize `ExperimentCollectionGroup`

BeritJanssen:
close #904 : adjust `experiment.views.get_experiment_collection()` in such a way that it returns all the overarching information of the `ExperimentSeries` (e.g., the `about_content`), and the experiment(s) of the `current_group`. The `current_group` is the ExperimentSeriesGroup attached to the queried `ExperimentSeries` with the lowest order and `finished=False`. `finished` will be set to `True` if a non-dashboard group has gone through all experiments.

If `dashboard=True`, the view will return an array of all experiments of the `current_group`, otherwise, the backend's json response will contain ~~`dashboard=None`~~ `dashboard=[]`. Now there is always a `next_experiment` object returned, no matter if `dashboard` is false or true, which will help with the implementation of #830 .


drikusroor (changes requested):
I think some work can be done with regards to the type safety of the Dashboard property - I think a list should always be a list and cannot result in a `None` value when it's empty.

Additionally, the `get_experiment_collection` might return an empty json object or a mix of two models (if I understand correctly). I think it would be wise to have a clear interface of what the frontend can expect and can work with. That would mean the returned json object's data structure should always be the same. Perhaps a separate DTO class could be introduced for this?

Furthermore, I think serialization functions should only serialize a model into a json object and not change too much (or anything for that matter). I would try to do the logic before serialization.

backend/experiment/views.py
```diff
@@ -76,89 +77,59 @@ def default_questions(request, rules):
...
+    active_groups = ExperimentSeriesGroup.objects.filter(
+        series=collection.id, finished=False).order_by('order')
+    if not active_groups:
+        return JsonResponse({})
+    current_group = active_groups.first()
+    serialized_group = serialize_experiment_series_group(
+        current_group, participant)
+    if not serialized_group:
+        # if the current group is not a dashboard and has no unfinished experiments, it will return None
+        return get_experiment_collection(request, slug)
```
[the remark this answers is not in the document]
BeritJanssen: No, we'll end up with an empty JSON response (see above) if we have non-dashboard groups which are all finished.

backend/experiment/views.py
```diff
@@ -76,89 +77,59 @@ def default_questions(request, rules):
...
+    return JsonResponse({
+        **serialize_experiment_series(collection),
+        **serialized_group
```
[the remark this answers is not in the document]
BeritJanssen: Yes. The group is serialized along with more generic information about the `ExperimentCollection`. This follows the original pattern of how information for different stages of an `ExperimentSeries` was retrieved. The resulting response will have the format now defined in the `IExperimentCollection` type.

backend/experiment/views.py
```diff
@@ -76,89 +77,59 @@ def default_questions(request, rules):
...
-            )
-    return JsonResponse()
-
-
-def serialize_experiment_series_group(group: ExperimentSeriesGroup):
-    grouped_experiments = GroupedExperiment.objects.filter(group_id=group.id).order_by('order')
+    active_groups = ExperimentSeriesGroup.objects.filter(
+        series=collection.id, finished=False).order_by('order')
+    if not active_groups:
+        return JsonResponse({})
```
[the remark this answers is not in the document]
BeritJanssen: The `finished` relates to the group, not to a single experiment. This would only ever be set to `True` without a dashboard, so this behaves as required.

BeritJanssen:
NB: I also just bit the bullet and renamed `ExperimentSeries` to `ExperimentCollection` throughout the backend, so that backend and frontend naming is consistent.

backend/experiment/views.py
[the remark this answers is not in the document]
drikusroor: But - considering the name of this method is `get_experiment_collection` - the experiment collection will still be there right? Why doesn't it return the information of the collection and what will the frontend show in this case?

frontend/src/types/ExperimentCollection.ts
```diff
@@ -1,14 +1,10 @@
 import Experiment from "./Experiment";
 
 export default interface ExperimentCollection {
-    id: number;
     slug: string;
     name: string;
     description: string;
     dashboard: Experiment[];
```
drikusroor, line 7: Looking at the code this might be a `null` value or even omitted completely from the object - which I think is a bad practice, as mentioned before. But if you're going to use it, this should probably be:

Assuming an explicit `null` value.
```suggestion
    dashboard: Experiment[] | null;
```

or, when it is omitted from the json body:
```suggestion
    dashboard?: Experiment[];
```

frontend/src/components/ExperimentCollection/ExperimentCollection.tsx
```diff
@@ -20,7 +20,8 @@ interface ExperimentCollectionProps extends RouteComponentProps<RouteParams> {
 
 const ExperimentCollection = ({ match }: ExperimentCollectionProps) => {
     const [experimentCollection, loadingExperimentCollection] = useExperimentCollection(match.params.slug) as [IExperimentCollection, boolean];
-    const experimentToRedirectTo = experimentCollection?.redirect_to;
+    const nextExperiment = experimentCollection?.next_experiment;
+    const displayDashboard = experimentCollection?.dashboard.length;
```
drikusroor, line 24: I think the PR description now says this is always an array. Does that mean this always resolves in `true`?

drikusroor (approved):
I get that this way is how the existing experiment series work and used to work, but I do think this needs refactoring sometime in the future. That could be now, or perhaps after the upcoming research projects later in spring and early summer. I still have an issue with the way that data structures are being mixed and returned.

I think there are two solutions to this problem:

1. We accept the (in my opinion) awkward way in which data is being returned, but we rename the `get_experiment_collection` to something else since we're not really returning an experiment collection but something else and rename the `ExperimentCollection` interface in the frontend. Maybe something like `ExperimentCollectionNextAction`, which is more descriptive of what it does.
2. We always return data in the same format, and the frontend can decide what to do with that data. It could be something like (pseudo-code):

```js
{
  id: 1 // experiment collection id
  // ... other experiment collection values
  dashboardItems: // list of experiment tiles to show in dashboard

  // option a: 
  // either defined or None
  currentGroup: {
    id: 11,
    currentExperiment: {
      id: 111,
      slug: "lorem"
    }
  }

  // option b, doesn't have to be exactly like this. It's more about the idea of accepting different data types while also making them explicit
  nextAction: {
    type: "experiment" | "group" | "debriefing" | "showDashboard"
    payload: {
      // contains the data for that specific next step, but because we have the type, we know what to do with it
    }
  }
}
```

PS My criticism might also have something to do with me not completely understanding how it should work. But then also, in my opinion, the code should explain what is happening, hence my suggestions for new ways to return the data. It could perhaps also help if we have an outline of all the possible scenarios that can happen in a collection? Is there something like that?

frontend/src/components/ExperimentCollection/ExperimentCollection.tsx
BeritJanssen: It's either 0 (=false), or a positive integer (=true).

frontend/src/types/ExperimentCollection.ts
BeritJanssen: As per your suggestion, the value returned if `dashboard=False` is a `[]`, not `None`, so no `null` value needed here.

BeritJanssen:
The `ExperimentCollection` interface is now predictable: it always has a `dashboard` and `next_experiment` member (coming from the `current_group`), as well as about/slug etc. I don't think we need to work with union types here. But I agree that refactoring is desireable (see #928 ), and that we can then handle the distinction between `ExperimentCollection->Experiment` and `ExperimentCollectionGroup->Phase` more elegantly.

BeritJanssen:
> But - considering the name of this method is get_experiment_collection - the experiment collection will still be there right? Why doesn't it return the information of the collection and what will the frontend show in this case?

True, the empty JSON response is a TODO. Should have this return the collection without a group, and then redirect to the debrief. Will leave this for after #922.

frontend/src/components/ExperimentCollection/ExperimentCollection.tsx
drikusroor: Ahhhh yes, one of the infamous javascript features 😊

Issue #841, opened 2024-03-12
Fetching frontend image fails


BeritJanssen:
For Visual Matching Pairs, the Preload component has a `fetch` call, which, in contrast to the `axios` calls, gives me CORS problems:
<img width="570" alt="Screenshot 2024-03-12 at 09 31 34" src="https://github.com/Amsterdam-Music-Lab/MUSCLE/assets/11174072/7dc6df6a-4245-4334-b1b6-9259ca6162c7">

For now, will introduce a `{mode: 'no-cors'}` workaround, but perhaps there's a less awkward solution.

BeritJanssen:
Could this be browser related? I see that Chrome (which I'm using) sets an Authorization header, which other browsers remove. (See table below Fetch API documentation)[https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API]

drikusroor:
> Could this be browser related? I see that Chrome (which I'm using) sets an Authorization header, which other browsers remove. [See table below Fetch API documentation](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#browser_compatibility)

Ah good catch, let me test it on Chrome too then 👍

drikusroor:
@BeritJanssen I'm experiencing no problems on Chrome either. The only thing that I could think of is this line in `base_settings.py`:

```python
# CORS origin white list from .env
CORS_ORIGIN_WHITELIST = os.getenv(
    "AML_CORS_ORIGIN_WHITELIST",
    "http://localhost:3000,http://127.0.0.1:3000,{}".format(HOMEPAGE)
).split(",")
```

`AML_CORS_ORIGIN_WHITELIST` is not set in my `.env` file. Is it set in yours with a different value perhaps? 🤔

drikusroor:
Closed as I could not reproduce it. Let's reopen the issue or create a new one if it pops up again, and then let's try to pin down the reproduction conditions.

PR #909, opened 2024-04-02
Fixed: Fix "unknown view" error and refactor fetch requests

drikusroor:
> This pull request fixes the "unknown view" error by adding the VISUALMATCHINGPAIRS experiment type to the Playback component.

The visual matching pairs rules set still returned the view as `VISUALMATCHINGPAIRS`, causing an "unknown view" error. Either we do it like this (by keeping the rule set and adding the view to `Playback.jsx`) or we remove this rules set and go full monty on the matching pairs rules sets and use the `determine_play_method`, which is more automagically but might also give the user less control (what if the user uses an audio format that's not in the list?).

> It also refactors the fetch requests by removing the 'no-cors' mode.

As for the CORS errors, I'm not getting them 😅 Have you also experienced them on ACC or APP? Maybe we can compare our `.env` files. In the screenshot in the issue it seems like your `Access-Allow-Control-Origin` is set to `http://localhost:3000`, is that correct? My allowed hosts environment variable is set to only localhost (`AML_ALLOWED_HOSTS=localhost`) could you @BeritJanssen try and see what happens if you do that as well?

Related to #841

### DIFF
--- a/frontend/src/components/Playback/Playback.jsx
+++ b/frontend/src/components/Playback/Playback.jsx
@@ -16,6 +16,7 @@ export const BUTTON = "BUTTON";
 export const MULTIPLAYER = "MULTIPLAYER";
 export const IMAGE = "IMAGE";
 export const MATCHINGPAIRS = "MATCHINGPAIRS";
+export const VISUALMATCHINGPAIRS = "VISUALMATCHINGPAIRS";
 export const PRELOAD = "PRELOAD";
 
 const Playback = ({
@@ -214,6 +215,7 @@ const Playback = ({
                     />
                 );
             case MATCHINGPAIRS:
+            case VISUALMATCHINGPAIRS:
                 return (
                     <MatchingPairs
                         {...attrs}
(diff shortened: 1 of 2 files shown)


BeritJanssen:
Ah, my intention was to remove the `visual_matching_pairs` rules. In my opinion, more control for the user also needs more settings, i.e., more coding required, so I'd rather go for taking the audio playback selection out of the user's hands. If this gets problematic with some audio formats, we can adjust the backend accordingly.

My env file also has `AML_ALLOWED_HOSTS="localhost"`, so that's not the reason. I'll try this branch on acc, if it works there, I'd be okay with removing the cors related code in the frontend.

BeritJanssen (changes requested):
Please roll back the changes for Playback.jsx. In my opinion, we should remove the `visual_matching_pairs` rules instead (migrate existing experiments to `matching_pairs_lite`.

BeritJanssen (approved):

Issue #913, opened 2024-04-03
When an experiment collection is opened with the participant_id set to a value, it should show result numbers only for that participant


drikusroor:
**Is your feature request related to a problem? Please describe.**
Related to #872 #873 and #912, we want to see the amount of times an experiment has been started and completed for specific participants.

![image](https://github.com/Amsterdam-Music-Lab/MUSCLE/assets/8208970/39138955-ffeb-475c-b660-641b075c4da1)

**Describe the solution you'd like**
If one navigates to https://app.amsterdammusiclab.nl/collection/samediff?participant_id=123, the dashboard should show the *started* and *finished* count numbers only for the participant with participant id 123.

BeritJanssen:
I thought that was the case, the sessions are filtered for the participant, or aren't they? If not, it should definitely be fixed, as this was always the intention.

drikusroor:
> I thought that was the case, the sessions are filtered for the participant, or aren't they? If not, it should definitely be fixed, as this was always the intention.

Ah okay, good to know. Maybe my test method was wrong. Will try to verify again! 😄

drikusroor:
> > I thought that was the case, the sessions are filtered for the participant, or aren't they? If not, it should definitely be fixed, as this was always the intention.
> 
> Ah okay, good to know. Maybe my test method was wrong. Will try to verify again! 😄

Update, they are indeed filtered for the current participant based on the session, _BUT_ the participant id from the url parameters wasn't being passed to the get experiment collection request, so I've addressed that in #917. Or should it have happened automatically already through the session?

PR #917, opened 2024-04-04
Refactor: Componentify participant condition, loader container, and cleanup several imports

drikusroor:
~This PR fixes not passing the participant id from the url parameters when fetching an experiment collection. It also adds the participant id to the links to the experiments, thereby keeping you in the same participant id context.~

The above has been fixed already in #925. This PR creates some separate components not have the ternary if statement in `App.jsx` anymore.

~Additionally it fixes the experiment not being loaded after clicking on the link to the experiment from the experiment collection. This was due to using a nested browser router where a switch should have been used.~

This too has been fixed in #925 already

Lastly, some imports are improved by using the `@` alias to avoid having to use relative imports.

Related #913 and #939


BeritJanssen (commented):
There's a whole mechanism that sets the participant_id if attaching `?participant_id` (without leading slash) to any route. It does not work as expected now indeed, but I'd prefer to see if we cannot get the intended functionality to work first. The idea is that we set the `participant_id` to the Django session, and then we'd retrieve that same participant for the experiments associated with the ExperimentCollection.

As for the `<Switch>` over `<Route>`, this change is definitely needed. One more thing that is necessary: attach the participant_id query to the redirect when going back to the dashboard view.

BeritJanssen:
Note: I have started working on a branch which incorporates the route / switch fix, and which should fix the participant_id query param issue without introducing new parameters.

frontend/src/components/ExperimentCollection/ExperimentCollectionDashboard/ExperimentCollectionDashboard.test.tsx
```diff
@@ -43,4 +50,28 @@ describe('ExperimentCollectionDashboard', () => {
...
+        render(
+            <MemoryRouter>
+                <ExperimentCollectionDashboard experimentCollection={{ dashboard: [experiment1] }} />
+            </MemoryRouter>
+        );
+        await waitFor(() => {
+            expect(screen.getByRole('menu')).toBeTruthy();
+            expect(screen.getByRole('menu').querySelector('a').getAttribute('href')).toBe('/some_slug');
+        });
+    });
```
BeritJanssen, line 76: Also add a unit test for adding the `participant_id`?

PR #918, opened 2024-04-04
Added: Show started session count in experiment collection tiles

### LINKED ISSUE #912: Experiment collection should show the amount an experiment has been *started*
Related to #872 and #873, which are about showing how often an experiment has been completed, we also want to show how often an experiment has been started (but not necessarily finished). We are expecting that not all participants will necessarily complete each experiment due to concentration span.

On the bottom left of the experiment tile in an experiment collection, the amount of times the experiment has been started in total should be shown as a number.

image

drikusroor:
Fix routing issue by using Switch instead of BrowserRouter in ExperimentCollection.tsx

Add started_session_count property to Experiment interface in Experiment.ts

Update Experiment interface in Experiment.ts to remove optional session count properties

Update ExperimentCollectionDashboard to display started and finished session counts in Experiment links

Update tests for newly added started session counter badge

Resolves #912

### DIFF
--- a/backend/experiment/views.py
+++ b/backend/experiment/views.py
@@ -105,7 +105,6 @@ def get_experiment_collection(request, slug):
         **serialize_experiment_collection(collection),
         **serialized_group
     })
-)
 
 
 def get_associated_experiments(pk_list):
--- a/backend/experiment/views.py
+++ b/backend/experiment/views.py
@@ -1,7 +1,6 @@
 import json
 import logging
 
-
 from django.http import Http404, JsonResponse
 from django.conf import settings
 from django.utils.translation import activate, gettext_lazy as _
--- a/backend/experiment/views.py
+++ b/backend/experiment/views.py
@@ -105,6 +105,7 @@ def get_experiment_collection(request, slug):
         **serialize_experiment_collection(collection),
         **serialized_group
     })
+)
 
 
 def get_associated_experiments(pk_list):
--- a/frontend/src/components/ExperimentCollection/ExperimentCollectionDashboard/ExperimentCollectionDashboard.tsx
+++ b/frontend/src/components/ExperimentCollection/ExperimentCollectionDashboard/ExperimentCollectionDashboard.tsx
@@ -39,7 +39,10 @@ export const ExperimentCollectionDashboard: React.FC<ExperimentCollectionDashboa
                             <Link to={"/" + exp.slug}>
                                 <ImageOrPlaceholder imagePath={exp.image} alt={exp.description} />
                                 <h3>{exp.name}</h3>
-                                <div role="status" className="counter">{exp.finished_session_count}</div>
+                                <div className="status-bar">
+                                    <span title={`Het "${exp.name}" experiment is ${exp.started_session_count} keer gestart`} role="status" className="counter">{exp.started_session_count}</span>
+                                    <span title={`Het "${exp.name}" experiment is ${exp.finished_session_count} keer voltooid`} role="status" className="counter">{exp.finished_session_count}</span>
+                                </div>
                             </Link>
                         </li>
                     ))}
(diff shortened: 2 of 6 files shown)


frontend/src/components/ExperimentCollection/ExperimentCollectionDashboard/ExperimentCollectionDashboard.tsx
```diff
@@ -39,7 +39,10 @@ export const ExperimentCollectionDashboard: React.FC<ExperimentCollectionDashboa
                             <Link to={"/" + exp.slug}>
                                 <ImageOrPlaceholder imagePath={exp.image} alt={exp.description} />
                                 <h3>{exp.name}</h3>
-                                <div role="status" className="counter">{exp.finished_session_count}</div>
+                                <div className="status-bar">
+                                    <span title={`Het "${exp.name}" experiment is ${exp.started_session_count} keer gestart`} role="status" className="counter">{exp.started_session_count}</span>
+                                    <span title={`Het "${exp.name}" experiment is ${exp.finished_session_count} keer voltooid`} role="status" className="counter">{exp.finished_session_count}</span>
+                                </div>
```
BeritJanssen, line 45: We don't want text to be set directly through the frontend, to be able to customize / localize this via the backend. As we'll revert this change later anyway (see #931), this is okay for now though. Perhaps the text had better be in English though, since we're kind of custom building it for the Congo team now?

BeritJanssen (approved):

frontend/src/components/ExperimentCollection/ExperimentCollectionDashboard/ExperimentCollectionDashboard.tsx
BeritJanssen: As another comment - I believe the name of the experiment is displayed right above the status bar? So perhaps this could be just "Started {} times", "Finished {} times".

backend/experiment/views.py
```diff
@@ -175,20 +180,42 @@ def serialize_experiment_series(
...
+        experiment: Experiment,
+        participant: Participant
+        ):
+    ''' Get the number of started sessions for
+        this experiment and participant '''
     count = Session.objects.filter(
-        experiment=experiment, participant=participant, finished_at__isnull=False).count()
+        experiment=experiment,
+        participant=participant,
+        started_at__isnull=False
```
BeritJanssen, line 203: By definition, started_at is not null, since this value will always be filled once a session is created. So this line can be left out.

backend/experiment/views.py
```diff
@@ -175,20 +180,42 @@ def serialize_experiment_series(
     }
 
 
-def serialize_experiment(experiment_object: Experiment, finished=0):
+def serialize_experiment(experiment_object: Experiment, started=0, finished=0):
```
BeritJanssen, line 183: See #933: need to pass in the `participant` here, and call the `get_started_session_count` function to retrieve the number. Otherwise, the dashboard will always display `0`.

frontend/src/components/ExperimentCollection/ExperimentCollectionDashboard/ExperimentCollectionDashboard.tsx
drikusroor, line 45:
```suggestion
                                <div className="status-bar">
                                    <span title={`Started ${exp.started_session_count} times`} role="status" className="counter">{exp.started_session_count}</span>
                                    <span title={`Started ${exp.finished_session_count} times`} role="status" className="counter">{exp.finished_session_count}</span>
                                </div>
```

backend/experiment/views.py
drikusroor: Oh, I liked this approach of keeping the database out of the serializer, but sure 😄 Maybe it's a good idea to look at that in the future and try to create some separation between aggregating the data needed for serialization and the actual serialization part.

PR #923, opened 2024-04-08
Conifg: Refactor SCSS build process in frontend

### LINKED ISSUE #686: Incorporate SCSS into React server/build
It's not really a big "problem", but right now we have to run `yarn scss` before starting the development server or building the application.

Webpack allows us to use a "loader" to also load `scss` files directly into our application. This will result into Webpack compiling the scss live during development and while building the application, saving us one command/step.

An alternative might be to start using [Vite](https://vitejs.dev/) in the future, and also configuring it to handle the scss straight away. Vite is many times faster than Webpack. But maybe that's for later.

**Additional context**
n/a

drikusroor:
This pull request refactors the SCSS build process in the frontend by removing the unnecessary "yarn scss" command from the build process. This improves the build efficiency and reduces the build time.

Resolves #686


BeritJanssen (commented):
LGTM. A question: while this also watch for scss changes during development?

drikusroor:
> LGTM. A question: while this also watch for scss changes during development?

It does! 😄 So no more manual `yarn scss:watch` while developing

PR #925, opened 2024-04-09
Fix/participant

### LINKED ISSUE #913: When an experiment collection is opened with the participant_id set to a value, it should show result numbers only for that participant

BeritJanssen:
Branching off #917 : wait for the participant to be fetched from `App.jsx`. This should fix #913. Not tested yet, though.

### DIFF
--- a/backend/participant/utils.py
+++ b/backend/participant/utils.py
@@ -83,10 +83,8 @@ def get_participant(request):
 
 def get_or_create_participant(request):
     """Get a participant from URL, the session, or create/add a new one"""
-
     # check if query string contains  participant
-    participant_id_url = request.GET.get("participant_id") # can be None
-
+    participant_id_url = request.GET.get("participant_id")  # can be None
     try:
         if participant_id_url:
             # get participant from query string
@@ -110,8 +108,7 @@ def get_or_create_participant(request):
         participant = Participant(country_code=country_code, access_info=access_info, participant_id_url=participant_id_url)
         participant.save()
         set_participant(request, participant)
-
-    return participant
+        return participant
 
 
 def set_participant(request, participant):
--- a/frontend/src/components/App/App.jsx
+++ b/frontend/src/components/App/App.jsx
@@ -12,6 +12,7 @@ import { URLS as API_URLS } from "../../API.js";
 import useBoundStore from "../../util/stores.js";
 import Experiment from "../Experiment/Experiment";
 import ExperimentCollection from "../ExperimentCollection/ExperimentCollection";
+import Loading from "../Loading/Loading";
 import Profile from "../Profile/Profile";
 import Reload from "../Reload/Reload";
 import StoreProfile from "../StoreProfile/StoreProfile";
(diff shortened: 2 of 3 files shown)


backend/participant/utils.py
```diff
@@ -110,8 +108,7 @@ def get_or_create_participant(request):
         participant = Participant(country_code=country_code, access_info=access_info, participant_id_url=participant_id_url)
         participant.save()
         set_participant(request, participant)
-
-    return participant
+        return participant
```
drikusroor, line 111: With the added tab, I think this will result in only returning the participant in the except clause, right?

```suggestion
    return participant
```

frontend/src/components/App/App.jsx
```diff
@@ -50,6 +52,11 @@ const App = () => {
 
     return (
         <Router className="aha__app">
+            { !participant? (
+            <div className="loader-container">
+                <Loading />
+            </div>
+            ) : (
```
drikusroor, line 59: Is the participant required for all routes or only specific route(s)?

drikusroor (approved):
Generally, looks good to me but I'm not too sure about the extra indent in `backend/participant/utils.py`.

I hadn't read your messages about this fix in time so I actually worked on fixing the issue myself too and come up with a similar solution: https://github.com/Amsterdam-Music-Lab/MUSCLE/pull/917/files#diff-ff8cf4543d7ac3b64876ed5499bacca9132f69c8ead3202d2e38f2d86eb73012 (see the `WithParticipantExperimentCollection` component). I did test it and it seemed to work, so I assume your version works as well. Depending on whether participant is required for only this route or for all routes, we could either use my fix or your fix. What I did like, though, is the idea of using a dedicated component to require the participant as being defined. If it should be used for all routes, one might call more generically call it a `ParticipantGate` or something.

frontend/src/components/App/App.jsx
BeritJanssen: In my view, all. There are some Routes here that I don't fully understand though, such as the `URLS.session` one (which is defined no-where).

backend/participant/utils.py
BeritJanssen: Yes, but otherwise, participant is not defined anyway, so this would throw an exception in any case.

drikusroor: Ah, I see. It does return participant in the try part already. nvm

drikusroor: There is one uncovered possibility though, for which I shall give a suggestion in the next comment

drikusroor: Update: I've tested it and it seems to be defined in all cases. So you can either choose the approach of returning `participant` in the end, or by returning `participant` in every clause.

image

PR #933, opened 2024-04-10
Fix/session display

BeritJanssen:
This PR fixes problems around displaying the finished sessions for an experiment in a collection.


backend/participant/utils.py
```diff
@@ -95,10 +95,6 @@ def get_or_create_participant(request) -> Participant:
             # Get participant from session
             participant = get_participant(request)
 
-            # if no participant_id URL parameter in request, but previous participant was created from URL, do not use it and create a new participant
-            if participant.participant_id_url:
-                raise Participant.DoesNotExist
-
```
BeritJanssen, line 101: Mind: we should discuss if we want to enforce having the `?participant_id` query parameter for all experiments and collections, or if we relax this requirement (what is done here). The risk of relaxing the requirement is that we'll unintentionally register results for a participant who played previously.

drikusroor:
> This PR fixes problems around displaying the finished sessions for an experiment in a collection.

What was the problem exactly?

Issue #966, opened 2024-04-24
MatchingPairs: log card position in result


BeritJanssen:
Right now, only section ids are logged, and there's no way of tracking back what the positioning of the sections was. Log the position and refer to the grid as follows:
1 2 3 4
5 6 7 8
9 10 11 12
13 14 15 16

(NB if another ordering is more useful, that's also fine, as long as it's consistent.)

Possible solution: in MatchingPairs component, send the `index` of the clicked card along with the `firstCard` and `secondCard` sent to the backend in the `scoreIntermediateResult` call.

PR #970, opened 2024-04-29
Feature er/matching pairs dry run issues

### LINKED ISSUE #965: MatchingPairs: log time interval between first card and second card clicked
With the reimplementation of MatchingPairs (`calculateIntermediateScore`) we don't log the time interval between the first and second card being clicked anymore. This should be reintroduced.

Possible solution: on the MatchingPairs component, (re)start a timer once the first card is clicked, send the time passed with `scoreIntermediateResult`.

### LINKED ISSUE #968: MatchingPairs: Change naming of lastCard & currentCard
I noticed that the naming of the cards being logged with the results is very confusing. Basically, `lastCard` is the *first card* of the turn, `currentCard` the *second card* of the turn. Change this so it reads "first card" and "second card" instead.

Evert-R:
This PR:

- Logs the board position in the results based on `sections[index+1]`
1   2   3   4
5   6   7   8
9  10 11 12
13 14 15 16

- Renames `first_card` and `second_card` in the results.
- Logs the response time between the click on the first card and the click on the second card as `response_interval_ms` in the results.

resolves: #966 , resolves #965 , resolves #968


frontend/src/components/MatchingPairs/MatchingPairs.jsx
```diff
@@ -90,14 +90,18 @@ const MatchingPairs = ({
...
                 // This is the second card to be turned
                 currentCard.turned = true;                
-                setSecondCard(currentCard);
+                setSecondCard(currentCard);                                
                 // set no mouse events for all but current
                 sections.forEach(section => section.noevents = true);
                 currentCard.noevents = true;
+                currentCard.boardposition = parseInt(index) + 1;
+                currentCard.timestamp = performance.now();
+                firstCard ? currentCard.response_interval_ms = Math.round(currentCard.timestamp - firstCard.timestamp) : {};
```
BeritJanssen, line 99: I've never seen this syntax before: a question mark before assignment? What does this do, is it a conditional assignment? Perhaps add a comment to clarify what the intention is here.

BeritJanssen (approved):

frontend/src/components/MatchingPairs/MatchingPairs.jsx
drikusroor: Yeah, I think this could be rewritten into something simpler. I'm not entirely sure what the intention is either.

Evert-R: I wasn't too happy about this as well, but the test `updates score after a match` kept failing because the value of firstCard was `null`.
I found another solution, I've now set the initial state of firstCard and secondCard to `{}`

PR #982, opened 2024-04-30
reset response interval

Evert-R:
reset response interval for first card, in case the card already has a value from a previous turn


frontend/src/components/MatchingPairs/MatchingPairs.jsx
```diff
@@ -116,7 +116,9 @@ const MatchingPairs = ({
                 currentCard.noevents = true;
                 currentCard.seen = true;
                 currentCard.boardposition = parseInt(index) + 1;
-                currentCard.timestamp = performance.now();                
+                currentCard.timestamp = performance.now();
+                // reset response interval in case this card has a value from a previous turn
+                currentCard.response_interval_ms = '';
```
drikusroor, line 121: Wouldn't a `null` or `0` value be better though since `performance.now()` returns a `number`?

Issue #928, opened 2024-04-09
Refactor ExperimentCollection(Group)


BeritJanssen:
After discussion today, it seems that renaming some of the classes, and refactoring endpoints would be helpful.

```python
class Experiment(models.Model) # aka ExperimentCollection
- name
- slug
- theme # FK
- about # TextField (markdown)
- consent # TextField (markdown)

class Phase(models.Model) # aka ExperimentCollectionGroup
- experiment # FK
- order
- dashboard # boolean
- randomize # boolean

class Rule(models.Model) # aka Experiment
- order
- name
- slug
- questions # may also be FK from dedicated Question model
- rule # ChoiceField
```

This means researchers/maintainers would always have to encapsulate a `Rule` within a `Phase` and `Experiment`, but it also feels more logical in that we're now moving towards adding `Consent` and `Theme` to both `Experiment` and `ExperimentColection` (in our current nomenclature).

drikusroor:
@BeritJanssen  I believe we can close this as everything has been renamed, right?

PR #993, opened 2024-05-02
Add react-helmet-async package and configure head/og tags

### LINKED ISSUE #834: Dynamically change page title, description and OG attributes based on current experiment
Right now, the page title, description and OG attributes are all based on environment variables and are therefore always static. Regardless of which experiment is loaded, sharing the page / link over social media will always show "Amsterdam Music Lab" with the AML logo. It would be nice if we could customize these metadata.

Use the [react-helmet-async](https://github.com/staylor/react-helmet-async) library to dynamically update the aforementioned metadata based on the Experiment data. We probably need #831 to be able to do that.

**Additional context**
- Based on [this request in Microsoft Teams](https://teams.microsoft.com/l/message/19:1fa546778985420497626097b2aaf0c4@thread.tacv2/1709736030360?tenantId=a0f1cacd-618c-4403-b945-76fb3d6874e5&groupId=5c5b7168-27f0-450f-80a6-1a7e2635e09a&parentMessageId=1709736030360&teamName=Amsterdam%20Music%20Lab&channelName=HookedOnTVTunes&createdTime=1709736030360) by @Lia-JX-Li

drikusroor:
This PR adds title, description, meta and structured data information dynamically through `react-helmet` for experiments.

Resolves #834


BeritJanssen (commented):
Let's discuss when it would be good to merge this. It's a nice feature to have, but I wonder how we'll deal with the fact that with #928, we'll reorganize information. For running something like Toontjehoger, it would be good that every Experiment (will be "Rule") has their own image, but if we move towards embedding every rule in an ExperimentCollection (will be "Experiment"), we'll need to refactor where the og information comes from. Or we hold this PR until after #928, but that's quite a long hold-up.

drikusroor:
> Let's discuss when it would be good to merge this. It's a nice feature to have, but I wonder how we'll deal with the fact that with #928, we'll reorganize information. For running something like Toontjehoger, it would be good that every Experiment (will be "Rule") has their own image, but if we move towards embedding every rule in an ExperimentCollection (will be "Experiment"), we'll need to refactor where the og information comes from. Or we hold this PR until after #928, but that's quite a long hold-up.

Will any of the relevant fields on an Experiment(Rule) be removed in the architectural changes? If not, I think we can keep it for now. Then we at least have the foundational code in the frontend already, which makes it easier to add OG information from other levels later on.

We could also think about fallback information in case information at the current level is not available. For example:

- OG Information at Experiment(Rule) level - Added in this PR & relevant fields exist on model. In case of no information, it falls back to:
- OG Information at Group(Phase) level - Non-existent, but something we can add later. In case of no information, it falls back to:
- OG Information at Collection(Experiment) level - Non-existent, but something we can add later. In case of no information, it falls back to:
- OG Information at site level

Or maybe I understood incorrectly... Do you mean that ideally we'll put the OG information on the Collection(Experiment) level later on?

BeritJanssen:
I think that og information should come from the `ExperimentCollection` level eventually. I think communication (such as sharing the experiment on social networks) will be about the `ExperimentCollection`, which in most cases just contains just one rule. `ExperimentCollection` already contains a `description`. I think in most cases the `og_image` (on `ExperimentCollection` level) would be the logo, which may be configured through the theme, and if this is not set, fallback to the site settings (through env variables) is indeed sensible. So perhaps this branch should be adjusted so it points to the `ExperimentCollection`'s description, and logo of the theme (with fallbacks). This will be immediately useful for Toontjehoger, which needs custom og information.

PR #1003, opened 2024-05-06
fix: do not call onNext() from within map

### LINKED ISSUE #1002: [Bug]: "Originals" board MatchingPairsLite does not always play sound
When playing the "Originals" board of MatchingPairsLite, occasionally, there wil be no sound when clicking on a tile.

BeritJanssen:
Close #1002 . The problem was caused by the `checkSectionLoaded` mechanism, which was introduced to prevent reloading files for consecutive views, such as in SongSync. The `onNext` call in the `.map` loop would cause the next view to be triggered before all files were loaded, as the `Originals` board contains duplicates of sections.

Moving the `onNext()` call to after the `.map` loop should fix this.

### DIFF
--- a/frontend/src/components/Preload/Preload.jsx
+++ b/frontend/src/components/Preload/Preload.jsx
@@ -34,13 +34,11 @@ const Preload = ({ sections, playMethod, duration, preloadMessage, pageTitle, on
             }
 
             if (playMethod === 'BUFFER') {
-
                 // Use Web-audio and preload sections in buffers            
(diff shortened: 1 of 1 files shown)


Evert-R (approved):
Good catch!

BeritJanssen:
I did some refactoring after all, also making sure that resources aren't loaded double anymore. I tested this for MatchingPairs, ThatsMySong and DDI / BAT, but more testing (Categorization) is very welcome.

Issue #746, opened 2024-01-30
Proposal: Run E2E tests after deployment to the Acceptance (and/or Test) environment


drikusroor:
## Requirements

- [x] E2E tests are run in a Chrome/Chromium-based browser
- [x] E2E tests are run in headless mode in its CI/CD workflow
- [ ] ~On Production, only smoke tests are run after a deployment to verify the site is still running and experiments can be loaded~ See issue #1060 for the smoke tests on Production after a release
- [x] On Acceptance, the full E2E suite is run after a deployment to verify the most critical features of MUSCLE are still working

## Original proposal

**Is your feature request related to a problem? Please describe.**
Sometimes new features, refactors, and fixes have unintended consequences that might not be captured by frontend and backend unit tests. This might cause is to discover newly introduced bugs later than is necessary.

**Describe the solution you'd like**
It would be nice to at least run some smoke/e2e tests periodically that verify if the MUSCLE platform as a whole is still working like it is supposed to. We already have some selenium tests configured which we can use for now and which we can expand upon over time. We can run these tests after deployment to the Acceptance environment. This means we don't have to configure new environments in our CI/CD pipeline and we can begin testing straight away.

**Describe alternatives you've considered**
I've thought about maybe using Playwright or Cypress, but since we already have Selenium, it might make more sense to just continue using that.

**Additional context**
Running the E2E tests automatically after deployment requires an automated CI/CD pipeline as well. Currently this does not yet exist for Acceptance, but it almost exists for the new Test environment.

BeritJanssen:
As discussed: instead of nightly, perhaps after every build would be sufficient.

BeritJanssen:
Hum, I thought "nightly" would be more frequent than "after push to develop", but you may be right that nightly is actually going to be less frequent. We concluded that it would be most efficient to run the e2e tests against the server itself, so no, the platform wouldn't be run in github actions, but deployed on the server + e2e test triggered to check whether requests to the server give expected responses. The only problem we have for that is a reproducible database - but perhaps we can work around that with Django fixtures (i.e., set up experiments and playlists with json exports from the production server, as is now already done for some Django unit tests). That way we can be sure a calling `experiment/myFancyExperiment` will give a response. Perhaps we would also always want to run tests with a `participant_id=testing` or similar, so as cleanup after e2e tests, we can delete all sessions + results associated with that participant.

drikusroor:
[...]

PR #1010, opened 2024-05-13
CI: Configure Selenium E2E tests to run in Docker and in Podman workflow

drikusroor:
This pull request adds a Dockerfile and test scripts for running Selenium tests in a Docker container. The Dockerfile installs the necessary dependencies, such as the Selenium Python package, Chromium browser, and Chrome WebDriver. The test scripts allow for running the tests either in a container or in an interactive bash terminal.

Related to #746

### DIFF
--- a/e2e/tests-selenium.py
+++ b/e2e/tests-selenium.py
@@ -161,11 +161,16 @@ class TestsSelenium(unittest.TestCase):
         self.check_for_error(experiment_name, experiment_slug)
 
         # Explainer
-        self.driver.find_element(By.XPATH, "//div[text()=\"Let's go!\"]").click()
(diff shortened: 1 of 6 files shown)


e2e/tests-selenium.py
```diff
@@ -0,0 +1,458 @@
...
+            raise Exception("Unknown browser")
+
+        self.driver.set_window_size(1920, 1080)
+
+    def tearDown(self):
+        self.driver.quit()
+
+    # This is a simple test to check if the e2e setup and the internet connection is working
+    def test_google(self):
+        self.driver.get("http://www.google.com")
```
BeritJanssen, line 114: This request could also go to `example.com` instead: less corporate.

BeritJanssen (approved):

PR #1024, opened 2024-05-21
Fixed: Fix reversed section URL prefix

### LINKED ISSUE #1020: 🐛 [BUG] - `Preload` and `audio.js` still include `MEDIA_ROOT`
### Description

On the new production server, requests for non-buffer sections still fail, as the `Preload` component and `audio.js` util still attempt to fetch from `MEDIA_ROOT + src`. Compare #1012.

Either:
1.  fix this in both these functions
2. adjust the backend function in `section/models.py`, `Section.absolute_url()`.

NB: Choose option 2. Option 1 does not work locally.

### Reproduction URL

https://illc-aml-mu-prd.science.uva.nl/rhythm_intro?participant_id=test

### Reproduction steps

```bash
Go through experiment until loading spinner appears. This will keep showing, as the section cannot be fetched from the backend.
```

### Logs

```bash
Network log shows attempted fetch from https://illc-aml-mu-prd.science.uva.nl/server/server/section
```

### Browsers

_No response_

### OS

_No response_

### Additional context

_No response_

drikusroor:
This PR prefixes the section urls from the `absolute_url` method in the section's `model.py` with an optional `BASE_URL` while removing the earlier used `MEDIA_ROOT` prefix from the fetch section actions in the frontend. This will make sure that the backend returns the correct section url while the frontend won't add a redundant `/server` path prefix.

There are two scenarios, which both require a slightly different configuration:

1. Local develop environment
2. Test / Acceptance / Production environments

## 1. Local develop environment

As the frontend runs on `localhost:3000` and the backend on `localhost:8000`, the returned section url needs to be prefixed with `http://localhost:8000` in the backend, as the `MEDIA_ROOT` (aka the `API_ROOT`) doesn't do that anymore in the frontend. In Django's develop settings, the `BASE_URL` setting is set to `http://localhost:8000` by default. The `absolute_url` method therefore returns `"http://localhost:8000" + "/section/123/456" = http://localhost:8000/section/123/456`.

## 2. Test / Acceptance / Production environments

As the frontend and the backend run on the same (sub) domain, `BASE_URL` needn't be configured. It therefore defaults to an empty string. The `absolute_url` method will therefore return `"" + "/section/123/456" = "/section/123/456"`. The frontend does not prefix this url with anything but fetches the path from the domain it is running on: `https://(tst/acc/prd).amsterdammusiclab.nl/section/123/456`.

# How to test

## Local develop environment

You do _not_ need to set `BASE_URL` to `http://localhost:8000` in your `.env` file as it will already default to that value in Django's develop settings. Simply check if the audio and other files in the experiments are still working.

## Other environments

I've deployed this branch to the test environment using the workflow dispatch method (see [here](https://github.com/Amsterdam-Music-Lab/MUSCLE/actions/runs/9181258840)) and verified that the audio playback works for this matching pairs experiment: https://tst.amsterdammusiclab.nl/introductiono

Resolves #1020


drikusroor:
Alternatively, I could also simplify it and hard-code the localhost:8000 in the `absolute_url` method whenever we are in develop mode..

BeritJanssen (approved):
Nice!

PR #1036, opened 2024-05-28
Hotfix/speech2song

BeritJanssen:
This branch hotfixes problems with speech2song: the `first_round` array didn't have the `Explainer` as the last object; additionally, there was a problem with the serialization of style: as the `next_repeated_representation` method generates copies of a Trial, on the second round of serialization, the `Playback` action within the `Trial` has already been transformed to a `dict`, causing an error when calling `style.to_dict()` again.

### DIFF
--- a/backend/experiment/actions/base_action.py
+++ b/backend/experiment/actions/base_action.py
@@ -13,7 +13,8 @@ class BaseAction(object):
         action_dict = self.__dict__
         action_dict['view'] = self.ID
 
-        if self.style is not None:
+        # we may have already converted the style object to a dictionary, e.g., after copying an Action object
+        if self.style is not None and type(self.style) is not dict:
             action_dict['style'] = self.style.to_dict()
 
         return action_dict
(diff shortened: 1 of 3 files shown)


backend/experiment/actions/base_action.py
```diff
@@ -13,7 +13,8 @@ def action(self):
         action_dict = self.__dict__
         action_dict['view'] = self.ID
 
-        if self.style is not None:
+        # we may have already converted the style object to a dictionary, e.g., after copying an Action object
```
drikusroor, line 16: Just interested: Why could this be converted to a dict already? Shouldn't serialization ideally happen at the end?

BeritJanssen: Not if we are using copies of the same object, which speech2song does. It makes on `Trial` object, and from that creates an array of `[Trial]*n_representations`. As the serialization goes through the list of actions, it will have already converted the Playback's style object to a `dict`, so in the next step it will attempt to apply the `to_dict` method again to the `style` attribute. Took me quite a long time to understand what was going on and why.

Issue #827, opened 2024-03-05
Calculate sum of scores experiments associated with ExperimentCollection


BeritJanssen:
## Requirements

- [ ] Create function to calculate the sum of the participant's scores for all experiments done within an experiment collection
- [ ] Scores sum data (number) will be added as a property to the get experiment collection request

Issue #828, opened 2024-03-05
Display score in frontend


BeritJanssen:
## Requirements

- Score will be added as a property to the get experiment collection request (see also #827 )
- If no score:
	- Nog geen punten!
- If score:
	- `x` punten
- Score component can be reused from ToontjeHoger

PR #1038, opened 2024-05-28
Add score and rank to header

Evert-R:
This PR adds the score element to the header of the experimentcollection dashboard. (if `showScore` in the theme is set)

These parameters are being send from the backend, but are hardcoded for now:
Score-label (points), 
score-class (gold), 
no-score-label (no points yet!) 

The data for the social sharing component is hardcoded as well, but in the front-end.

closes: #827, #828

### DIFF
--- a/backend/experiment/serializers.py
+++ b/backend/experiment/serializers.py
@@ -50,12 +50,15 @@ def serialize_experiment_collection_group(group: ExperimentCollectionGroup, part
     next_experiment = get_upcoming_experiment(
         grouped_experiments, participant, group.dashboard)
 
+    total_score = get_total_score(grouped_experiments, participant)
+
     if not next_experiment:
         return None
 
     return {
         'dashboard': [serialize_experiment(experiment.experiment, participant) for experiment in grouped_experiments] if group.dashboard else [],
-        'next_experiment': next_experiment
+        'next_experiment': next_experiment,
+        'total_score': total_score
     }
 
 
@@ -93,3 +96,13 @@ def get_finished_session_count(experiment, participant):
     count = Session.objects.filter(
         experiment=experiment, participant=participant, finished_at__isnull=False).count()
     return count
+
+
+def get_total_score(grouped_experiments, participant):
+    '''Calculate total score of all experiments on the dashboard'''
+    total_score = 0
+    for grouped_experiment in grouped_experiments:
+        sessions = Session.objects.filter(experiment=grouped_experiment.experiment, participant=participant)
+        for session in sessions:
+            total_score += session.final_score
+    return total_score
--- a/backend/theme/serializers.py
+++ b/backend/theme/serializers.py
@@ -20,8 +20,11 @@ def serialize_header(header: HeaderConfig) -> dict:
     return {
         'nextExperimentButtonText': _('Next experiment'),
         'aboutButtonText': _('About us'),
-        'showScore': header.show_score
-    }
+        'showScore': header.show_score,
+        'scoreClass': 'gold',
+        'scoreLabel': _('Points'),
+        'noScoreLabel': _('No points yet!')
+        }
 
 
 def serialize_theme(theme: ThemeConfig) -> dict:
--- a/frontend/src/components/ExperimentCollection/ExperimentCollection.tsx
+++ b/frontend/src/components/ExperimentCollection/ExperimentCollection.tsx
@@ -34,7 +34,8 @@ const ExperimentCollection = ({ match }: ExperimentCollectionProps) => {
     const nextExperiment = experimentCollection?.next_experiment;
     const displayDashboard = experimentCollection?.dashboard.length;
     const showConsent = experimentCollection?.consent;
-
+    const totalScore = experimentCollection?.total_score
+    const scoreClass = experimentCollection?.score_class
     const onNext = () => {
         setHasShownConsent(true);
     }
@@ -72,7 +73,7 @@ const ExperimentCollection = ({ match }: ExperimentCollectionProps) => {
         <div className="aha__collection">
             <Switch>
                 <Route path={URLS.experimentCollectionAbout} component={() => <ExperimentCollectionAbout content={experimentCollection?.aboutContent} slug={experimentCollection.slug} />} />
-                <Route path={URLS.experimentCollection} exact component={() => <ExperimentCollectionDashboard experimentCollection={experimentCollection} participantIdUrl={participantIdUrl} />} />
+                <Route path={URLS.experimentCollection} exact component={() => <ExperimentCollectionDashboard experimentCollection={experimentCollection} participantIdUrl={participantIdUrl} totalScore={totalScore} scoreClass={scoreClass} />} />
             </Switch>
         </div>
     )
--- a/frontend/src/components/ExperimentCollection/ExperimentCollectionDashboard/ExperimentCollectionDashboard.tsx
+++ b/frontend/src/components/ExperimentCollection/ExperimentCollectionDashboard/ExperimentCollectionDashboard.tsx
@@ -12,16 +12,19 @@ interface ExperimentCollectionDashboardProps {
     participantIdUrl: string | null;
 }
 
-export const ExperimentCollectionDashboard: React.FC<ExperimentCollectionDashboardProps> = ({ experimentCollection, participantIdUrl }) => {
-
+export const ExperimentCollectionDashboard: React.FC<ExperimentCollectionDashboardProps> = ({ experimentCollection, participantIdUrl, totalScore }) => {
+    
     const dashboard = experimentCollection.dashboard;
-    const nextExperimentSlug = experimentCollection.nextExperiment?.slug;
+    const nextExperimentSlug = experimentCollection.nextExperiment?.slug;   
+    
     const headerProps = experimentCollection.theme?.header? {
-        nextExperimentSlug,
+        nextExperimentSlug,        
         collectionSlug: experimentCollection.slug,
-        ... experimentCollection.theme.header
+        ...experimentCollection.theme.header,
+        totalScore: totalScore
+        
     } : undefined;
-
+    
     const getExperimentHref = (slug: string) => `/${slug}${participantIdUrl ? `?participant_id=${participantIdUrl}` : ""}`;
 
     return (
--- a/frontend/src/components/Header/Header.tsx
+++ b/frontend/src/components/Header/Header.tsx
@@ -1,5 +1,8 @@
-import React from "react";
+import React, { useEffect, useState, useRef } from "react";
 import { Link } from "react-router-dom";
+
+import Rank from "../Rank/Rank";
+import Social from "../Social/Social"
 
 interface HeaderProps {
     nextExperimentSlug: string | undefined;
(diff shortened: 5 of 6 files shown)


frontend/src/components/Header/Header.tsx
```diff
@@ -1,15 +1,84 @@
...
+            };
+            id = setTimeout(nextStep, 50);
+
+            return () => {
+                window.clearTimeout(id);
+            };
+        }, [targetScore]);
+
+        return score;
+    };
```
drikusroor, line 66: I think this hook (collection of hooks) is overly complicated. You are setting a timeout in several places, which feels kind of iffy. It's probably also better to use `requestAnimationFrame` than `setTimeout` in this context as it is more efficient and works better with your display refresh rate. Furthermore, I don't think you need the `scoreValue = useRef(0)` hook and you can do without. An example of how it could also be done:

```suggestion
    const useAnimatedScore = (targetScore: number) => {
        const [score, setScore] = useState(0);
    
        useEffect(() => {
            if (targetScore === 0) {
                setScore(0);
                return;
            }
    
            let animationFrameId: number;
    
            const nextStep = () => {
                setScore((prevScore) => {
                    const difference = targetScore - prevScore;
                    const scoreStep = Math.max(1, Math.min(10, Math.ceil(Math.abs(difference) / 10)));
    
                    if (difference === 0) {
                        cancelAnimationFrame(animationFrameId);
                        return prevScore;
                    }
    
                    const newScore = prevScore + Math.sign(difference) * scoreStep;
                    animationFrameId = requestAnimationFrame(nextStep);
                    return newScore;
                });
            };
    
            // Start the animation
            animationFrameId = requestAnimationFrame(nextStep);
    
            // Cleanup function to cancel the animation frame
            return () => {
                cancelAnimationFrame(animationFrameId);
            };
        }, [targetScore]);
    
        return score;
    };
```

frontend/src/components/Header/Header.tsx
```diff
@@ -1,15 +1,84 @@
...
+export const Header: React.FC<HeaderProps> = ({ nextExperimentSlug, nextExperimentButtonText, collectionSlug, aboutButtonText, showScore, totalScore, scoreClass, scoreLabel, noScoreLabel }) => {
+    
+    const social = {
+        'apps': ['facebook', 'twitter'],
+        'message': `I scored ${totalScore} points`,
+        'url': 'wwww.amsterdammusiclab.nl',
+        'hashtags': ["amsterdammusiclab", "citizenscience"]
+    }
+
+    const useAnimatedScore = (targetScore) => {
```
drikusroor, line 28: Also, aren't we already using a very similar or perhaps even identical hook in the existing `Score` component (`src/components/Score.jsx`)? We might even think about extracting that hook (or a refined version) to a separate file and use the hook in both that and this `Score` component.

frontend/src/components/Header/Header.tsx
```diff
@@ -18,6 +87,21 @@ export const Header: React.FC<HeaderProps> = ({ nextExperimentSlug, nextExperime
                     {aboutButtonText && <Link className="btn btn-lg btn-outline-primary" to={`/collection/${collectionSlug}/about`}>{aboutButtonText}</Link>}
                 </nav>
             </div>
+            {showScore, totalScore !== 0 && (
+                <div className="results">
+                    <Score
+                    score={totalScore}
+                    scoreClass={scoreClass}
+                    label={scoreLabel}
+                    />
```
drikusroor, line 96:
```suggestion
                    <Score
                        score={totalScore}
                        scoreClass={scoreClass}
                        label={scoreLabel}
                    />
```

backend/experiment/serializers.py
```diff
@@ -93,3 +96,13 @@ def get_finished_session_count(experiment, participant):
...
+
+
+def get_total_score(grouped_experiments, participant):
+    '''Calculate total score of all experiments on the dashboard'''
+    total_score = 0
+    for grouped_experiment in grouped_experiments:
+        sessions = Session.objects.filter(experiment=grouped_experiment.experiment, participant=participant)
+        for session in sessions:
+            total_score += session.final_score
+    return total_score
```
drikusroor, line 108: This seems like a good method to write a test for

backend/theme/serializers.py
```diff
@@ -20,8 +20,11 @@ def serialize_header(header: HeaderConfig) -> dict:
     return {
         'nextExperimentButtonText': _('Next experiment'),
         'aboutButtonText': _('About us'),
-        'showScore': header.show_score
-    }
+        'showScore': header.show_score,
+        'scoreClass': 'gold',
+        'scoreLabel': _('Points'),
+        'noScoreLabel': _('No points yet!')
```
drikusroor, line 26: Thinking out loud... Maybe it'd be nice to put this in a `score` dictionary? Something like:

```suggestion
        'score': {
            'scoreClass': 'gold',
            'scoreLabel': _('Points'),
            'noScoreLabel': _('No points yet!')
            }
```

Or thinking even louder: don't pass the `showScore` boolean. Simply define the score object if `showScore` is true and otherwise set it to `None`. The frontend can interpret the `None` (`null` in js) as not having to show the score.

```suggestion
        'score': {
            'scoreClass': 'gold',
            'scoreLabel': _('Points'),
            'noScoreLabel': _('No points yet!')
            } if header.show_score else None,
```

frontend/src/components/ExperimentCollection/ExperimentCollection.tsx
```diff
@@ -34,7 +34,8 @@ const ExperimentCollection = ({ match }: ExperimentCollectionProps) => {
     const nextExperiment = experimentCollection?.next_experiment;
     const displayDashboard = experimentCollection?.dashboard.length;
     const showConsent = experimentCollection?.consent;
-
+    const totalScore = experimentCollection?.total_score
+    const scoreClass = experimentCollection?.score_class
```
drikusroor, line 38: Isn't the score class defined on the experiment collection's theme's header configuration?

frontend/src/components/ExperimentCollection/ExperimentCollectionDashboard/ExperimentCollectionDashboard.tsx
```diff
@@ -12,16 +12,19 @@ interface ExperimentCollectionDashboardProps {
...
     const dashboard = experimentCollection.dashboard;
-    const nextExperimentSlug = experimentCollection.nextExperiment?.slug;
+    const nextExperimentSlug = experimentCollection.nextExperiment?.slug;   
+    
     const headerProps = experimentCollection.theme?.header? {
-        nextExperimentSlug,
+        nextExperimentSlug,        
         collectionSlug: experimentCollection.slug,
-        ... experimentCollection.theme.header
+        ...experimentCollection.theme.header,
```
drikusroor, line 23: I would pass this as a separate object as more related to the header theme than the specific data & information that needs to be shown. E.g. something like:

```suggestion
        headerTheme: experimentCollection.theme.header,
```

frontend/src/types/Theme.ts
```diff
@@ -2,6 +2,7 @@ export interface Header {
     nextExperimentButtonText: string;
     aboutButtonText: string;
     showScore: boolean;
+    totalScore: BigInteger;
```
drikusroor, line 5: Again, I don't think the total score should be part of a `Theme` interface.

drikusroor (changes requested):
I think this PR begs some refinement in several areas. 

First, I think some work should be done to simplify the `useAnimatedScore` while also abstracting and re-using it / combining it with the already existing `useAnimatedScore` hook from the existing `Score` component. I also think the hook and Score component should not be part of the `Header` component. 

Second, I think it's worth checking out which properties need to be defined where, both in the backend and in the frontend. In my opinion, data-related properties like a total score shouldn't be defined in a theme and should be defined higher up in the experiment collection.

frontend/src/components/Header/Header.tsx
```diff
@@ -18,6 +87,21 @@ export const Header: React.FC<HeaderProps> = ({ nextExperimentSlug, nextExperime
                     {aboutButtonText && <Link className="btn btn-lg btn-outline-primary" to={`/collection/${collectionSlug}/about`}>{aboutButtonText}</Link>}
                 </nav>
             </div>
+            {showScore, totalScore !== 0 && (
```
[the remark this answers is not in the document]
Evert-R: fixed

frontend/src/types/Theme.ts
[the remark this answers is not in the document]
Evert-R: Number is indeed fine

frontend/src/components/Header/Header.tsx
Evert-R: This was the old toontje hoger code. I commited your changes to it

frontend/src/components/Header/Header.tsx
Evert-R: That's a complete view. The idea here was to just implement the code already in toontje hoger.

backend/experiment/serializers.py
Evert-R: Indeed,  done

frontend/src/types/Theme.ts
Evert-R: changed to header

BeritJanssen:
In the interest of time, perhaps it's good to leave the refactor of `Score` for later. It would be indeed neat if we could split `Score` into its parts and avoid duplication, but I guess this will take more time to do & test well. Let's make a separate issue for this.

It is worthwhile making a decision about where exactly to pass the score object. Would it make sense to make this part of the theme.header? Since it would be slightly strange to have a `score` object sent along with the `experimentCollection`, but then not having it shown due to the `experimentCollection.theme.header` not being defined?

backend/theme/serializers.py
BeritJanssen: I agree that having the information in its own dictionary is more readable, testable, and portable.

backend/theme/tests/test_serializers.py
BeritJanssen, line 1: The `expected_json` in the test needs to be adjusted, I think.

frontend/src/components/Header/Header.tsx
```diff
@@ -1,15 +1,87 @@
...
 
+import Rank from "../Rank/Rank";
+import Social from "../Social/Social"
+
 interface HeaderProps {
     nextExperimentSlug: string | undefined;
     nextExperimentButtonText: string;
     collectionSlug: string;
     aboutButtonText: string;
     showScore: boolean;
```
BeritJanssen, line 12: This boolean is not sent by the backend anymore.

frontend/src/components/Header/Header.tsx
```diff
@@ -1,15 +1,87 @@
...
+    totalScore: Number;
+    score: Object;
```
BeritJanssen, line 14:
```suggestion
    score: Score;
```

frontend/src/types/Theme.ts
```diff
@@ -1,7 +1,7 @@
 export interface Header {
     nextExperimentButtonText: string;
     aboutButtonText: string;
-    showScore: boolean;
+    showScore: boolean;
```
BeritJanssen, line 4: This is not consistent with what's implemented in the backend now (i.e., `score` dictionary).

PR #1042, opened 2024-05-29
Added: Validate file existence validation in Experiment & Playlist forms

drikusroor:
This pull request fixes the file existence validation in the Playlist model and admin form. 

The `clean_csv` method in the Playlist model now performs file existence validation for section files. It checks if the referenced files exist using the `file_exists_validator` function. If any file is not found, a `ValidationError` is raised and added to the `csv` field errors.

Additionally, it fixes an existing problem in the experiment rules validation method.

### DIFF
--- a/backend/experiment/forms.py
+++ b/backend/experiment/forms.py
@@ -192,6 +192,8 @@ class ExperimentForm(ModelForm):
         for playlist in playlists:
             errors = rules.validate_playlist(playlist)
 
+            playlist.clean()
+
             for error in errors:
                 playlist_errors.append(f"Playlist [{playlist.name}]: {error}")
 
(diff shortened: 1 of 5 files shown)


backend/experiment/forms.py
```diff
@@ -182,6 +182,10 @@ def clean_playlists(self):
         rules = cl()
 
         playlists = self.cleaned_data['playlists']
+
+        if not playlists:
+            return self.cleaned_data['playlists']
+
```
BeritJanssen, line 188: Repetition of #1039.

backend/experiment/forms.py
```diff
@@ -194,6 +198,8 @@ def clean_playlists(self):
         if playlist_errors:
             self.add_error('playlists', playlist_errors)
 
+        return playlists
```
BeritJanssen, line 201: Will cause merge conflict with #1039.

BeritJanssen (approved):
Nice!  The issues I pointed out in `experiment.forms` will get fixed with merging, just pointing this out in case there's a specific commit that can be dropped here to clean the history.

PR #1043, opened 2024-05-30
Fixed: Do not call onNext in onResult when we should break the round

drikusroor:
This PR fixes a race condition that appeared in the Hooked Eurovision experiment ( `(app|acc).amsterdammusiclab.nl/eurovision_2021` ) whenever you answered "No" to the question on whether you know the music that had just been played. The `onNext` function would called twice, once from the `onResult` hook function, and once from the `Trial` component directly.

This PR then solves this by making `onResult` _not_ call `onNext` when the round should break. There might be better solutions out there that we can discuss at a later time, but this seems to fix it. Some possible alternative solutions: 

### 1. Split `onNext` from `onResult`

I think it might be worth thinking about making sure `onNext` never gets called form `onResult` and making `onResult` only responsible for sending results. The hook function `onResult` is only used in two places: `FeedbackFrom` and `Trial` so it's not a lot of refactoring I suppose.

Pseudo code:

```ts
// onResult won't call onNext anymore
await onResult(...);

// onNext **always** gets called, with the optional doBreak parameter
const doBreak = ...;
onNext(doBreak);
```

### 2. Use `doBreak` in `onResult`
 
3. Alternatively, we could think about passing `shouldBreak` to the `onResult` function so that the `onNext` call from within `onResult` can be called with the appropriate `doBreak` parameter.

```ts
// doBreak param gets passed with onResult
const resultData = ...;
const forceSubmit = ...;
const doBreak = ...;
await onResult(resultData, forceSubmit, doBreak);

// onResult will then call onNext with doBreak
function onResult(result, forceSubmit, doBreak) {

    // ...other code
    if (hasNextRound && !forceSubmit && goToNextAction) {
        onNext(doBreak);
        return;
    }
    // ...other code

    // Advance to next round
    if (goToNextAction) {
        onNext(doBreak);
    }
}
```


BeritJanssen:
Nice! I was thinking to move towards solution 1, i.e., uncouple the logic for sending requests to log/score results from the querying of rounds. One complicating factor though with `await`ing the Result call is that it makes the play experience of `Hooked` experiments less smooth, as we want to present the next round right away, not only after the result call is complete. So indeed, let's figure out how to do this later, and merge this fix for now.

Issue #142, opened 2022-06-17
Back up local database before `migrate` in development


albertas-jn:
After `migrate`, the local database might become incompatible with other branches. We should automatically back up the existing local database so that we could switch between databases when switching branches.

## Related to / blocked by

- #893

BeritJanssen:
I tried the following approaches:
- making use of `pg_dump`: since we don't have information on the migrations, we just dump every time `docker-compose up` is run. Problematic because a `command` in the Dockerfile doesn't wait for the start up of Postgres to be complete. Potential solution: use some `wait-for-it` entrypoint (i.e,. healthcheck & dump rolled into one bash script)
- making use of the `dumpdata` management command. Advantage: can use something along those lines: `python manage.py migrate --check | if [$PIPESTATUS[0] > 0] ; then python manage.py dumpdata -o wherever/we/want/the/backup.dmp; fi `, so only run a backup if there are unapplied migrations. Disadvantage: dumpdata will exit as well if the migrations haven't been applied, as it relies on Django's model state.

PR #1045, opened 2024-05-31
Added (CI): Add schedule for database backup

### LINKED ISSUE #893: Automate production database dumps
The current setup on the new servers does not have an automated dump of the Django database. This should be added as a Github action. My instinct is to do so daily, with some kind of automatic rotation.

drikusroor:
This pull request adds a new schedule for database backup in the `.github/workflows/schedule-db-backup.yml` file, which uses the workflow template as defined in `db-backup-template.yml`.

## Requirements

- Creates a backup every night for ~all~ the production environment~s~ (~tst, acc,~ prd)
- Uses the following backup filename format: `{YYYYMMDD}.dump`
- Removes backups older than one week

## When runs it?

- The backup will run every night at midnight UTC and includes jobs to backup ~test, acceptance, and~ the production environment~s~.
- Manual workflow dispatch trigger (run a backup whenever we want)

## Example of backup workflow template

https://github.com/Amsterdam-Music-Lab/MUSCLE/blob/fa5290e6a7d6b03ad64ddb5c0d77820b776a1a69/.github/workflows/db-backup-template.yml#L15-L25

## Example of successful workflow

https://github.com/Amsterdam-Music-Lab/MUSCLE/actions/runs/9315480458

Resolves #893 

## Next steps

- Backup the db before a migration / deploy? #142


drikusroor:
@jaburgoyne If you happen to read this, could you let me know if you agree with the requirements as stated in this PR description? I can tweak the frequency of the backups, which environments should be backupped, and the max. age of a db backup before it gets deleted.

BeritJanssen:
Nice! One thing I'd like to still coordinate with Auke / dev ops is whether there is a path on the host's file system we can also copy out the files to. I believe that the file system of the servers gets backed up to other drives, which would be what we want in order to avoid ever losing data beyond 24 hours. Also, perhaps we should delete dumps older than a month (?) to make sure the volumes don't grow beyond the allocated limits, to avoid the nefarious "no space left on device" error.

BeritJanssen:
NB: I think the daily dump is really essential for production only. Test and acceptance can make do with weekly dumps, or even only dumps before migration IMO.

drikusroor:
> Also, perhaps we should delete dumps older than a month (?) to make sure the volumes don't grow beyond the allocated limits, to avoid the nefarious "no space left on device" error.

This already happens in the last step - though it deletes backups older than a week instead of older than a month.

jaburgoyne:
I think these are reasonable guidelines, but I would maintain production backups for a much longer cycle. I think we need logic along the lines of:

- daily backups for one or two weeks
- weekly backups for a year (i.e., after one/two weeks, delete all daily backups except the last one of each week)
- monthly backups for multiple years/indefinitely (i.e., after a year, hold onto the last backup of each month and delete the rest)

PR #1048, opened 2024-06-03
Feature/display footer

### LINKED ISSUE #980: Display footer in frontend
As of #906 , we have an (optional) `FooterConfig` on the `ThemeConfig` model, which adds logos. The frontend implementation to display the footer is still missing.

## Blocked

Blocked by #989 as they most likely share the fetching of the theme in the experiment collection.

BeritJanssen:
Close #980 . Display a disclaimer, logos and privacy statement, cf. [Toontjehoger](https://toontjehoger.amsterdammusiclab.nl/toontjehoger). The styling may still need to be adjusted, but leaving this for later.

### DIFF
--- a/frontend/src/components/AppBar/AppBar.jsx
+++ b/frontend/src/components/AppBar/AppBar.jsx
@@ -8,7 +8,7 @@ const AppBar = ({ title, logoClickConfirm = null }) => {
 
     const theme = useBoundStore((state) => state.theme);
 
-    const logoUrl = theme?.logoUrl? (API_BASE_URL + theme.logoUrl) : LOGO_URL;
+    const logoUrl = theme?.logoUrl ? (API_BASE_URL + theme.logoUrl) : LOGO_URL;
 
     // Handle click on logo, to optionally confirm navigating
     const onLogoClick = (e) => {
--- a/frontend/src/components/AppBar/AppBar.jsx
+++ b/frontend/src/components/AppBar/AppBar.jsx
@@ -8,7 +8,7 @@ const AppBar = ({ title, logoClickConfirm = null }) => {
 
     const theme = useBoundStore((state) => state.theme);
 
-    const logoUrl = theme? (API_BASE_URL + theme.logo_url) : LOGO_URL;
+    const logoUrl = theme?.logoUrl? (API_BASE_URL + theme.logoUrl) : LOGO_URL;
 
     // Handle click on logo, to optionally confirm navigating
     const onLogoClick = (e) => {
--- a/frontend/src/components/Footer/Footer.scss
+++ b/frontend/src/components/Footer/Footer.scss
@@ -8,7 +8,7 @@
     flex-direction: column;
     text-align: center;
 
-    .disclaimer > * {
+    .disclaimer, .privacy > * {
         max-width: 700px;
         margin: 0 auto;
     }
@@ -17,7 +17,7 @@
         padding-top: 30px;
         padding-bottom: 30px;
         display: flex;
-       gap: 45px;
+        gap: 45px;
         flex-wrap: wrap;
         align-items: center;
         justify-content: center;
--- a/frontend/src/components/Footer/Footer.scss
+++ b/frontend/src/components/Footer/Footer.scss
@@ -1,0 +1,56 @@
+.aha__footer {
+    margin-top: 30px;
+    padding-top: 20px;
+    padding-bottom: 30px;
+    display: flex;
+    justify-content: center;
+    align-items: center;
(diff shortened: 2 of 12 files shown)


frontend/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,34 @@
...
+                {logos.map((logo: ILogo, index: number) => (
+                    <a
+                        href={logo.href}
+                        target="_blank"
+                        rel="noopener noreferrer"
+                        key={index}
+                    >
+                        <img src={API_BASE_URL + logo.file} alt={logo.alt} />
+                    </a>
+                ))};
```
drikusroor, line 24: This semicolon results into a weird layout:

```suggestion
                ))}
```

image

frontend/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+import { API_BASE_URL } from "@/config";
+
+import { Footer as IFooter, Logo as ILogo } from "@types/Footer";
+
+export const Footer: React.FC<IFooter> = ({disclaimer, logos, privacy}) => {
+    return (
+        <div className="aha__footer">
+            <div className="disclaimer" dangerouslySetInnerHTML={{
+                __html: disclaimer,
```
drikusroor, line 11: Perhaps this should be fixed in a separate issue, but I tested the markdown in the disclaimer and the privacy statement and it all looks a bit weird. I think some styles & classes are conflicting with each other, resulting in texts being centered, max-width's being applied, but also the default markdown/prose styling.

I think it would already help a lot if we just apply a `max-width: 700px` and `margin: 0 auto;` on the containers so that everything gets centered in a consistent way.

image

frontend/src/components/Footer/Footer.scss
```diff
@@ -0,0 +1,56 @@
...
+    p {
+        max-width: 700px;
+    }
+
+    .logos {
+        padding-top: 30px;
+        padding-bottom: 30px;
+        display: flex;
+        column-gap: 45px;
+        row-gap: 45px;
```
drikusroor, line 20:
```suggestion
       gap: 45px;
```

frontend/src/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,34 @@
...
+                        href={logo.href}
+                        target="_blank"
+                        rel="noopener noreferrer"
+                        key={index}
+                    >
+                        <img src={API_BASE_URL + logo.file} alt={logo.alt} />
+                    </a>
+                ))};
+            </div>
+            <div className="disclaimer" dangerouslySetInnerHTML={{
```
drikusroor, line 26: I see toontje hoger originally used to have a designated `.privacy` css class. Did you not use that class consciously?

frontend/src/components/Footer/Footer.scss
```diff
@@ -0,0 +1,56 @@
...
+    padding-bottom: 30px;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    flex-direction: column;
+    text-align: center;
+
+    p {
+        max-width: 700px;
+    }
```
drikusroor, line 13: I would apply this to every child element of the disclaimer and the privacy containers, or to the disclaimer and privacy containers themselves. And probably with centering them using margins as well:

```suggestion
    .disclaimer > * {
        max-width: 700px;
        margin: 0 auto;
    }
```

frontend/src/components/AppBar/AppBar.jsx
```diff
@@ -8,7 +8,7 @@ const AppBar = ({ title, logoClickConfirm = null }) => {
 
     const theme = useBoundStore((state) => state.theme);
 
-    const logoUrl = theme? (API_BASE_URL + theme.logo_url) : LOGO_URL;
+    const logoUrl = theme?.logoUrl? (API_BASE_URL + theme.logoUrl) : LOGO_URL;
```
drikusroor, line 11:
```suggestion
    const logoUrl = theme?.logoUrl ? (API_BASE_URL + theme.logoUrl) : LOGO_URL;
```

drikusroor (approved):
LGTM! I added some code change suggestions that you could apply which probably eliminates a large part of the styling issues already.

frontend/src/components/Footer/Footer.tsx
BeritJanssen: No, this is indeed a copy & paste error. Good catch!

BeritJanssen:
Thanks for the style improvements!

PR #1051, opened 2024-06-03
Fixed: Fix play again URL rendering

### LINKED ISSUE #1046: That’s My Song ‘Play Again’ button fails to load experiment
### Description

The title says it all: at the end of a That’s My Song round, the Play Again button brings up a screen with the error message that the experiment cannot be loaded. 

### Reproduction URL

_No response_

### Reproduction steps

```bash
1. Play a complete That’s My Song round 
2. Click on 'Play Again' on the trophy screen
```

### Logs

_No response_

### Browsers

Safari

### OS

Mac

### Additional context

Could we add iOS and Android to the list of platforms?

## Drikus' notes

Link from backend seems to be valid:

```json
"button": {
  "text": "Play again",
  "link": "https://acc.amsterdammusiclab.nl/thats_my_song?participant_id=test"
}
```

However, the frontend uses React Router's `Link' component, which expects a relative path instead of an absolute url. It subsequently prefixes the url with a slash `/`.

image

### Proposed solution

- For Play Again, only send relative urls (that's my song & hooked) from the backend.
- Depending on whether the Final actions' button link is a relative path or an absolute url, render a `Link` component or render an achor tag (`<a href...>`) for the absolute urls.

drikusroor:
This pull request fixes the rendering of the "Play again" button URL in the Final component. Previously, the backend would send an absolute url (inc. domain) to the frontend, which is incompatible with React Router's Link component, which expects a relative path.

First, the backend now only sends relative paths for the Play again button link in the That's my song & Hooked experiment rules.

Second, the frontend checks whether the received button link from the Final action is relative or absolute (whether it is prefixed with "/") and renders a Link or anchor tag (<a>) component/element accordingly.

Resolves #1046

### DIFF
--- a/backend/experiment/rules/base.py
+++ b/backend/experiment/rules/base.py
@@ -5,12 +5,12 @@ from django.utils.translation import gettext_lazy as _
 from django.conf import settings
 
 from experiment.actions import Final, Form, Trial
-from experiment.models import Experiment
 from section.models import Playlist
 from experiment.questions.demographics import DEMOGRAPHICS
 from experiment.questions.goldsmiths import MSI_OTHER
 from experiment.questions.utils import question_by_key, unanswered_questions
 from result.score import SCORING_RULES
+from session.models import Session
 
 from experiment.questions import get_questions_from_keys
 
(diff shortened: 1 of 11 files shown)


BeritJanssen (approved):
Nice!

PR #1052, opened 2024-06-04
Implement ToontjeHoger Kids experiments

BeritJanssen:
These are the ToontjeHogerKids rules files, with updated texts from the Leiden team. Some of them also have a slightly different setup (Plink, Preverbal), but most revolve around minor changes.

Also changed the problem of all ToontjeHoger rules files in that they included an (unnecessary) Playlist action in the `first_round` array.


backend/experiment/rules/toontjehogerkids_4_absolute.py
```diff
@@ -0,0 +1,71 @@
+import logging
+
+from os.path import join
+from django.template.loader import render_to_string
+from .toontjehoger_1_mozart import toontjehoger_ranks
+from experiment.actions import Trial, Explainer, Step, Final, Info
+from .toontjehoger_4_absolute import ToontjeHoger4Absolute
+
+logger = logging.getLogger(__name__)
```
drikusroor, line 9: This logger isn't being used I think

```suggestion
```

drikusroor (approved):
LGTM. I didn't look at all the code in detail, but I think it's more a matter of testing the whole series when it's on TST / ACC.

PR #1054, opened 2024-06-04
Bugfix/final on next

BeritJanssen:
The ToontjeHoger experiments continue after the "Final" round, but the `onNext` method would get a `SyntheticBaseEvent` instead of undefined / false for the `doBreak` condition. This could be solved through wrapping the onNext call from the `onClick` in a function, which would not pass through an event anymore.

Also changed the `Link` to an `a` element. This to avoid an error being thrown when `button` was defined, but `button.link`, and therefore `<Link to='...`>` was not defined.

### DIFF
--- a/frontend/src/components/Experiment/Experiment.tsx
+++ b/frontend/src/components/Experiment/Experiment.tsx
@@ -93,7 +93,7 @@ const Experiment = ({ match }) => {
     };
 
     // trigger next action from next_round array, or call session/next_round
-    const onNext = async (doBreak) => {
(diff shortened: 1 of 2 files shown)


drikusroor:
> Also changed the Link to an a element. This to avoid an error being thrown when button was defined, but button.link, and therefore <Link to='...>` was not defined.

I've addressed this in #1051 too. Let's coordinate on how to approach ;)

BeritJanssen:
This change will be incorporated in in #1051.

PR #1063, opened 2024-06-04
Fixed: Fix internal Final button link (v2)

### LINKED ISSUE #1046: That’s My Song ‘Play Again’ button fails to load experiment

drikusroor:
Fix internal Final button link, which wouldn't cause an experiment to reload as the user was already on that page, by introducing an InternalRedirect.

The InternalRedirect has its own route and then redirects the user to the correct page with the correct search params. In short, it will lead you from `/thats_my_song?participant_id=123` to `/redirect/hats_my_song?participant_id=123`, which in turn will lead you back to `/thats_my_song?participant_id=123`, which will then trigger an experiment reload.

Alternatively, we could also just revert back to the earlier used anchor tag (<a>), which will do a full page reload but causes a little screen flash / glitch to appear as it reloads all the HTML / JS & CSS and shows an empty page for some milliseconds.

## Additional notes

There was an earlier attempt #1051  to fix problems related to the Play again button in the Final component, but it generated some unanticipated problems.

Resolves #1046

### DIFF
--- a/frontend/src/components/App/App.tsx
+++ b/frontend/src/components/App/App.tsx
@@ -18,6 +18,7 @@ import Profile from "../Profile/Profile";
 import Reload from "../Reload/Reload";
 import StoreProfile from "../StoreProfile/StoreProfile";
 import useDisableRightClickOnTouchDevices from "../../hooks/useDisableRightClickOnTouchDevices";
+import { InternalRedirect } from "../InternalRedirect/InternalRedirect";
 
 
 // App is the root component of our application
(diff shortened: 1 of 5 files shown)


BeritJanssen (approved):
LGTM. We had the anchor solution in place before, which did appear glitchy to some users, so let's go this way.

PR #1064, opened 2024-06-04
Feature/thk debriefs

BeritJanssen:
Add the "Info" templates for the end of the ToontjeHoger experiments. Since the templates were virtually the same (text & two videos), I decided to work with *one* template and a context dictionary for the render. This will hopefully make it easier to update the video links later (the current links are preliminary).


backend/experiment/rules/toontjehogerkids_1_mozart.py
```diff
@@ -60,13 +60,15 @@ def get_final_round(self, session):
...
+            horen van Mozart? Dit noemen we het 'Mozart effect'! Wil je meer weten over het Mozart effect? Bekijk dan de filmpjes!"
         body = render_to_string(
-            join('info', 'toontjehoger', 'experiment1.html'))
+            join('info', 'toontjehogerkids', 'debrief.html'), {'debrief': debrief_message, 'vid1': "https://www.youtube.com/embed/iPnZOrPPcBM?si=hFwmD4bVKGW2eGs1", 'vid2': "https://www.youtube.com/embed/pqAt-dvAqoI?si=z06Bd5UBVaA_go3w"})
         info = Info(
             body=body,
             heading="Het Mozart effect",
             button_label="Terug naar ToontjeHoger",
-            button_link="/toontjehoger"
+            button_link="/collection/thkids"
```
drikusroor, line 71: Although good for now and keeping in mind that we have a deadline to make, perhaps we can get the value of this variable by checking if the current experiment is part of a collection, or the collection that is related to the current session (if that's possible).

drikusroor (approved):
LGTM 🚀

backend/experiment/rules/toontjehogerkids_1_mozart.py
BeritJanssen: Yes, I thought the same. Hard-coding here for now, but eventually, would like to have a utility function to pull out the adequate information from the session data.

PR #1068, opened 2024-06-05
Added: Style play buttons according to neutral-inverted style

### LINKED ISSUE #1057: Toontje Hoger styling adjustments: Absolute / Tempo
## Requirements

- Playbuttons' border color extends as background color to the underlying labels
- The second play button should be orange
- Choice buttons [A] & [B] should be smaller

drikusroor:
This pull request adds the "neutral-inverted" style to the play buttons in the codebase. It also updates the necessary components and files to use this new style.

Resolves #1057

### DIFF
--- a/frontend/src/components/Playback/Multiplayer.scss
+++ b/frontend/src/components/Playback/Multiplayer.scss
@@ -25,8 +25,37 @@
 
     .instruction {
         min-height: 55px;
+
         h3 {
             margin: 0;
         }
+    }
+
+    &.neutral-inverted {
+
+        .player-wrapper:nth-child(1) {
+
+            .aha__play-button {
+                background-color: $blue;
+                box-shadow: 0 0 0 0.2em rgba($blue, 0.5);
+            }
+
+            .banner {
+                background-color: rgba($blue, 0.75);
+            }
+        }
+
+        .player-wrapper:nth-child(2) {
+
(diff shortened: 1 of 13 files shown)


frontend/src/components/Playback/Multiplayer.scss
```diff
@@ -29,8 +25,37 @@
...
+                box-shadow: 0 0 0 0.2em rgba($yellow, 0.5);
+            }
+
+            .banner {
+                background-color: rgba($yellow, 0.75);
+            }
+        }
+
+    }
+}
```
BeritJanssen, line 61: Could this bit be added to the `color-schemes.scss` file instead? My idea was that we'd gather all the styling related to the `style` attributes sent from the backend in that file.

BeritJanssen (approved):

frontend/src/components/Playback/Multiplayer.scss
drikusroor: When we're talking about the styling of the basic html building blocks like headings, paragraphs, lists, links, inputs, labels and buttons, I would agree. But this Multiplayer component's styling is so specific, with that hanging banner underneath the play button I think it would be better to keep it here. I'm not saying that it's ideal, but having to switch files every time I want to change the styling of a specific component also annoys me in advance 😅

PR #1081, opened 2024-06-07
TH5 Playlist Validation: Add playlist tag validation to ToontjeHoger(Kids)5Tempo rules

### LINKED ISSUE #1076: Sections need to have a tag value that matches the following regex: `^[CJR][1-5]_P[12]_(OR|CH)$`. Valid are: `C3_P2_OR, C2_P1_OR, C4_P2_OR, C4_P2_OR, C5_P2_OR, C5_P2_CH, C4_P1_OR, C2_P1_CH, C3_P1_OR, C2_P2_OR,`. Invalid are: `C6_P1_OR, C1_P3_OR, C1_P1_OZ` @BeritJanssen can you confirm this?

drikusroor:
This pull request adds tag validation to the ToontjeHoger(Kids)5Tempo rules. The `validate_tags` method has been implemented to check if the tags in the playlist are valid. Additionally, a test case has been added to ensure that the `validate_playlist` method does not raise any errors for invalid tags.

Resolves #1076

### DIFF
--- a/backend/experiment/rules/tests/test_toontjehoger_5_tempo.py
+++ b/backend/experiment/rules/tests/test_toontjehoger_5_tempo.py
@@ -7,7 +7,7 @@ from section.models import Playlist
 from experiment.rules.toontjehoger_5_tempo import ToontjeHoger5Tempo
 
 
-class ToontjeHoger5TempoTest(TestCase):
+class TestToontjeHoger5Tempo(TestCase):
 
     def setUp(self):
         # Mock the file_exists_validator function from section.models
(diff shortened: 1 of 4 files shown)


BeritJanssen (approved):

BeritJanssen:
One thing I wonder about: if we attach a playlist with 100s of incorrectly tagged sections, won't this mean the admin interface may actually get too crowded with warnings for tags? Perhaps it's better to just raise *one* error if any of the tags are invalid, which just states (in natural language) the naming convention for the tags?

PR #1085, opened 2024-06-10
Fix: Teletunes questions import and circular import

BeritJanssen:
After merging, as discussed on another channel, I ran into problems due to a missing import in the Teletunes questions, which in turn, revealed a problem with a circular import, which I propose to fix by removing the defaults for the `questions=` argument in `question_by_key` and `total_unanswered_questions` of `question.utils`.


drikusroor (approved):
LGTM and the tests are successful so I'm confident it works 👍

PR #1089, opened 2024-06-10
Refactor deployment process to properly stop containers before deployment

drikusroor:
This PR properly stops the running containers before deployment using `podman-compose down` before deploying the containers using `podman-compose up` in both ACC & PROD. Additionally, I've moved the `podman-compose down` command for the TST environment to run *after* the build step to minimze downtime.

### DIFF
--- a/.github/workflows/podman.yml
+++ b/.github/workflows/podman.yml
@@ -264,7 +264,7 @@ jobs:
       - name: Shut down running containers
         run: podman compose -f docker-compose-deploy.yml down
       - name: Deploy Podman images
-        run: podman-compose -f docker-compose-deploy.yml up -d --force-recreate
+        run: podman-compose -f docker-compose-deploy.yml up -d
       - name: Notify Sentry of new release
         run: |
           curl -X POST "https://sentry.io/api/0/organizations/uva-aml/releases/" \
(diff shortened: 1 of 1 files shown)


.github/workflows/podman.yml
```diff
@@ -172,6 +172,8 @@ jobs:
           cp .env frontend/.env
       - name: Build Podman images
         run: podman-compose -f docker-compose-deploy.yml build
+      - name: Shut down running containers
+        run: podman compose -f docker-compose-deploy.yml down
       - name: Deploy Podman images
         run: podman-compose -f docker-compose-deploy.yml up -d --force-recreate
```
BeritJanssen, line 178:
```suggestion
        run: podman-compose -f docker-compose-deploy.yml up -d
```

.github/workflows/podman.yml
```diff
@@ -259,6 +261,8 @@ jobs:
...
```
BeritJanssen, line 267:
```suggestion
        run: podman-compose -f docker-compose-deploy.yml up -d
```

BeritJanssen (approved):
LGTM, but I think the `--force-recreate` flags should still be left out?

PR #1090, opened 2024-06-10
TH4 Playlist validation (groups & tags)

### LINKED ISSUE #1075: TH4 - Sections need to have a group value of a number from 1 to `PLAYLIST_ITEMS`
While the original Toontjehoger Absolute game has 13 playlist items, this validator should be flexible: it should be entirely possible to implement ToontjeHoger Absolute with less playlist items, without having to override the validator. I don't think we can formulate this as a REGEX necessarily, perhaps we can check that there are distinct groups in a given range.

drikusroor:
This pull request adds Toontje Hoger 4 Absolute playlist group validation.

It checks if the group values in the sections are sequential and have the same length as the amount of sections. So let's say we have 6 sections, the (unique) group values should then be 1, 2, 3, 4, 5, 6. - Question to the reviewer: is that how the group values are supposed to be?

It also checks if the tags don't have values other than "a", "b" or "c".

Additionally, it moves the matching pairs test to the tests folder.

Resolves #1075 and #1091


backend/experiment/rules/tests/test_toontjehoger_4_absolute.py
```diff
@@ -0,0 +1,114 @@
...
+        # instead of section.validators as it is imported in the Playlist class
+        # which is in the section.models module
+        patcher = patch('section.models.file_exists_validator')
+        self.mock_file_exists_validator = patcher.start()
+        self.mock_file_exists_validator.return_value = None
+        self.addCleanup(patcher.stop)
+
+    @classmethod
+    def setUpTestData(self):
+        self.section_csv = (
```
BeritJanssen, line 27: Is this test data actually used? It seems to be Matching Pairs test data?

backend/experiment/rules/tests/test_toontjehoger_4_absolute.py
```diff
@@ -0,0 +1,114 @@
...
+    def test_validate_valid_groups(self):
+        csv_data = (
+            "Albania 2018 - Eugent Bushpepa,Mall,7.046,45.0,ToontjeHoger4Absolute/audio-1.mp3,a,1\n"
+            "Albania 2018 - Eugent Bushpepa,Mall,7.046,45.0,ToontjeHoger4Absolute/audio-2.mp3,c,2\n"
+            "Albania 2018 - Eugent Bushpepa,Mall,7.046,45.0,ToontjeHoger4Absolute/audio-3.mp3,a,3\n"
+            "Albania 2018 - Eugent Bushpepa,Mall,7.046,45.0,ToontjeHoger4Absolute/audio-4.mp3,b,4\n"
+            "Albania 2018 - Eugent Bushpepa,Mall,7.046,45.0,ToontjeHoger4Absolute/audio-5.mp3,c,5\n"
+            "Albania 2018 - Eugent Bushpepa,Mall,7.046,45.0,ToontjeHoger4Absolute/audio-6.mp3,b,6\n"
+        )
+        playlist = PlaylistModel.objects.create(name='TestToontjeHoger5Tempo')
```
BeritJanssen, line 67: Is this a typo? Or why is the Tempo playlist assigned to the Absolute rules? For clarity, it would be better if the playlist and the experiment had the same name.

backend/experiment/rules/toontjehoger_4_absolute.py
```diff
@@ -184,3 +185,20 @@ def get_final_round(self, session):
         )
 
         return [*score, final, info]
+
+    def validate_playlist(self, playlist: Playlist):
+        errors = super().validate_playlist(playlist)
+
+        # Get group values from sections, ordered by group
+        groups = list(playlist.section_set.values_list('group', flat=True).order_by('group'))
```
BeritJanssen, line 193: This will give you a query set as follows:
```python
<QuerySet ['1', '1', '1', '10', '10', '10', '11', '11', '11', ...]>
```

Not sure if this is your intention here? It seems as if the unit test would fail if you went for values above 9.

backend/experiment/rules/toontjehoger_4_absolute.py
```diff
@@ -184,3 +185,20 @@ def get_final_round(self, session):
...
+
+        # Check if the groups are sequential and unique
+        if groups != list(map(str, range(1, len(groups) + 1))):
+            errors.append(f"Groups in playlist sections should be sequential and unique from 1 to {len(groups)}. E.g. {list(range(1, len(groups) + 1))}")
+
+        # Check if the tags are 'a', 'b' or 'c'
+        tags = list(playlist.section_set.values_list('tag', flat=True).distinct())
```
BeritJanssen, line 200: I'm surprised the test doesn't fail, because, testing with the actual playlist, this code gives me the following:
```python
QuerySet ['a', 'b', 'c', 'a', 'b', 'c', 'a', ...]>
```
You need to linclude a `order_by` clause (previous to `distinct` in the chain) in order to truly deduplicate values.

BeritJanssen: Ah, I see, you didn't add a test with valid tags.

BeritJanssen (changes requested):
The test includes irrelevant `setUpTestData` code, and does not test a scenario with valid tags.

The validation code itself seems problematic - for the groups, because they are strings, and will be listed in alphabetic order, for the tags, because `distinct()` will not deduplicate this way.

backend/experiment/rules/toontjehoger_4_absolute.py
drikusroor: The test with valid groups also has valid tags

drikusroor: Also, it seems that distinct() does work without a group by. If I do this:

```py
# Check if the tags are 'a', 'b' or 'c'
tags = list(playlist.section_set.values_list('tag', flat=True).distinct())
print(tags, len(tags))
```

I get:

```sh
['a', 'b', 'c'] 3
['a', 'b', 'c', 'd'] 4
['a', 'b', 'c'] 3
```

BeritJanssen: But did you try adding the actual playlist? If that works, I concede - even though I locally ran this code with the actual playlist and got a different result.

drikusroor: Apparently `distinct` works differently in unit tests than in real life 🤯 I've added the `order_by` now

backend/experiment/rules/toontjehoger_4_absolute.py
drikusroor: Fixed

BeritJanssen (approved):

PR #1093, opened 2024-06-10
fix: include `BASE_URL` in image file serialization

BeritJanssen:
This branch fixes a problem on the servers, in which image requests are incorrectly prefixed with `API_ROOT`. Instead, this branch prefixes the `BASE_URL` for the backend image serialization.

### DIFF
--- a/backend/aml/base_settings.py
+++ b/backend/aml/base_settings.py
@@ -195,4 +195,4 @@ MARKUP_SETTINGS = {
     }
 }
 
-SUBPATH = os.getenv('AML_SUBPATH', None)
+SUBPATH = os.getenv('AML_SUBPATH', '')
--- a/backend/experiment/tests/test_views.py
+++ b/backend/experiment/tests/test_views.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.test import TestCase
 from django.utils import timezone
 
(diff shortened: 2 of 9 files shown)


drikusroor (approved):
Nice!

backend/aml/base_settings.py
```diff
@@ -195,4 +195,4 @@
     }
 }
 
-SUBPATH = os.getenv('AML_SUBPATH', None)
+SUBPATH = os.getenv('AML_SUBPATH', '')
```
drikusroor, line 198: Now that you are using `BASE_URL` and not `AML_SUBPATH` to prefix the image urls, is this change still needed?

backend/experiment/tests/test_views.py
```diff
@@ -190,7 +191,7 @@ def test_serialize_experiment(self):
         )
         self.assertEqual(
             serialized_experiment['image'], {
-                'file': '/upload/test-image.jpg', 'href': '', 'alt': ''}
+                'file': f'{settings.BASE_URL}/upload/test-image.jpg', 'href': '', 'alt': ''}
```
drikusroor, line 194: You might as well test the serialization of the `href` & `alt` properties here :)

drikusroor (approved):
Looks good to me. Did `AML_SUBPATH` not work well in the end?

backend/aml/base_settings.py
BeritJanssen: No, but I like the empty string fallback better, anyway. I twill still evaluate to `False` in the backend.

PR #1096, opened 2024-06-11
fix: TH Preverbal styles and player labels

BeritJanssen:
I noticed that the "Preverbal" experiment, which uses the "ImagePlayer", still had unstyled play and response buttons. I added an extra colour category for `neutral` and `neutral-inverted`, which styles the last element in an array to teal in both cases.

I also fixed a bug: the preverbal rules file still passed a `label_style` argument (which was originally picked up by the frontend), instead of `labels`.

### DIFF
--- a/backend/experiment/actions/playback.py
+++ b/backend/experiment/actions/playback.py
@@ -86,7 +86,6 @@ class Multiplayer(PlayButton):
     '''
     This is a player with multiple play buttons
     - stop_audio_after: after how many seconds to stop audio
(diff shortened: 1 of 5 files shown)


backend/experiment/actions/playback.py
```diff
@@ -109,7 +108,7 @@ def __init__(
             self.labels = labels
 
 
-class ImagePlayer(PlayButton):
+class ImagePlayer(Multiplayer):
```
drikusroor, line 111: This is a conscious change, right? 😄

drikusroor (approved):
👍

PR #1097, opened 2024-06-11
Fix/logos

BeritJanssen:
This branch ensures that logos will be displayed correctly, i.e., centered and big on the dashboard, and smaller top-left for the experiments (these would require defining separate themes for now).


frontend/src/components/ExperimentCollection/ExperimentCollectionDashboard/ExperimentCollectionDashboard.scss
```diff
@@ -0,0 +1,7 @@
+.aha__dashboard {
+
+    .aha__logo {
+        margin-left: 35%;
```
drikusroor, line 4: What's the reason for using a percentage as margin? Should it scale with device width?

backend/theme/serializers.py
```diff
@@ -39,8 +39,8 @@ def serialize_theme(theme: ThemeConfig) -> dict:
         'description': theme.description,
         'headingFontUrl': theme.heading_font_url,
         'bodyFontUrl': theme.body_font_url,
-        'logoUrl': join(settings.MEDIA_URL, str(theme.logo_image.file)) if theme.logo_image else None,
-        'backgroundUrl': join(settings.MEDIA_URL, str(theme.background_image.file)) if theme.background_image else None,
+        'logoUrl': f'{settings.BASE_URL.strip("/")}/{settings.MEDIA_URL.strip("/")}/{str(theme.logo_image.file)}' if theme.logo_image else None,
+        'backgroundUrl': f'{settings.BASE_URL.strip("/")}/{settings.MEDIA_URL.strip("/")}/{str(theme.background_image.file)}' if theme.background_image else None,
```
drikusroor, line 43: Can't you use the `serialize_image` method like in the other PRs of today?

https://github.com/Amsterdam-Music-Lab/MUSCLE/blob/38669ad060e69c85fcc43f1676ac8beebe5c5641/backend/image/serializers.py#L8-L13

drikusroor (approved):
LGTM. I had some questions and remarks but I'll let you decide what to do with them :-)

frontend/src/components/Logo/Logo.scss
```diff
@@ -0,0 +1,19 @@
...
+    height: $logo-height;
+    background-position: left center;
+    background-repeat: no-repeat;
+    background-size: contain;
+    margin-left: 15px;
+    font-size: 0;
+    color: transparent;
+    width: 30%;
+
+    @media (max-width: 500px) {
```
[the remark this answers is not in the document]
BeritJanssen: I totally agree that we should clean this up. I copied and pasted this breakpoint from former code. I'll see about reusing the breakpoint from variables here instead.

backend/theme/serializers.py
BeritJanssen: Eventually, that's 100% the better way, yes. For now, the frontend changes involved to adapt to the objects here are kind of daunting (especially to use the href defined in the images).

PR #1099, opened 2024-06-11
Fix/thk playlists

BeritJanssen:
This branch fixes remaining issues with playlist validation of Toontjehoger rules files.
It also implements a much simplified strategy for picking sections and giving feedback in `ToontjeHogerKids5Tempo`.


backend/experiment/rules/toontjehoger_4_absolute.py
```diff
@@ -199,15 +199,16 @@ def validate_playlist_groups(self, groups):
         # Check if the groups are sequential and unique
         integer_groups.sort()
         if integer_groups != list(range(1, len(groups) + 1)):
-            return ['Groups in playlist sections should be sequential numbers starting from 1 to the number of sections in the playlist ({}). E.g. "1, 2, 3, ... {}"'.format(len(groups), len(groups))]
+            return ['Groups in playlist sections should be sequential numbers starting from 1 to the number of items in the playlist ({}). E.g. "1, 2, 3, ... {}"'.format(self.PLAYLIST_ITEMS, self.PLAYLIST_ITEMS)]
```
drikusroor, line 202: > While the original Toontjehoger Absolute game has 13 playlist items, this validator should be flexible: it should be entirely possible to implement ToontjeHoger Absolute with less playlist items, without having to override the validator.

I thought this validator should be flexible, or have I misunderstood what you wrote? 😄

drikusroor (approved):

backend/experiment/rules/toontjehoger_4_absolute.py
BeritJanssen: This was a misunderstanding indeed: with "playlist items" I meant, "items, of which there are three variants". So `n_sections` != `self.PLAYLIST_ITEMS`, but it shouldn't be hardcoded to 13 either. Hurrying code reviews is not a good idea, I should've caught this before.

backend/experiment/rules/toontjehoger_4_absolute.py
```diff
@@ -199,15 +199,16 @@ def validate_playlist_groups(self, groups):
...
 
         return []
 
     def validate_playlist(self, playlist: Playlist):
         errors = super().validate_playlist(playlist)
 
         # Get group values from sections, ordered by group
-        groups = list(playlist.section_set.values_list('group', flat=True))
+        groups = list(playlist.section_set.values_list(
+            'group', flat=True).distinct())
```
BeritJanssen, line 211: This is the actual needed change though - before, there were as many groups as there were sections. I still think the reference to `PLAYLIST_ITEMS` n the `validate_playlist_groups` function is more readable.

Issue #1106, opened 2024-06-12
🐛 [BUG] - TH(Kids) issues on TST


drikusroor:
### Description

* [x] Even though TH1 only has 2 rounds, the header shows 1/10 and 2/10 rounds. UPDATE: Fixed by setting only 2 rounds in the experiment instead of 10.
* [x] The Final component seems to link back to `/toontjehoger` instead of the collection (which on TST has the route `/collection/thkids`). NB: I see that the link _should_ be configured correctly already for Kids, and it seems the adult TH1 is being used in the collection on TST. UPDATE: Fixed by configuring the THK1 - Mozart rules.
* [x] When the audio is playing, it counts down to 0 but it already goes to the next question when there are still 15 seconds left (on the countdown). -> this is caused by incorrect duration value in the playlist, fixed by editing from admin
* [x] "Maatgevoel" seems to not reliably play a "or" vs "ch" section, check selection mechanism
* [x] "Absoluut gehoor" doesn't stop showing the loading spinner **UPDATE: ** This seems to work on [ACC](https://acc.amsterdammusiclab.nl/the_absolute?participant_id=test)

### Screenshots

![image](https://github.com/Amsterdam-Music-Lab/MUSCLE/assets/8208970/e77a7a6a-733e-4979-9525-6db426e33e15)

### Reproduction URL

https://tst.amsterdammusiclab.nl/toontje-hoger-1-mozart?participant_id=test)

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

drikusroor:
The issue of point 4. seems to be here, in the `section_from_any_song` method:

https://github.com/Amsterdam-Music-Lab/MUSCLE/blob/53dd7e2a08c266aba22e244ca8841c0c67edffc5/backend/session/models.py#L142-L153

On line 152, it queries sections using the song ids. There is only one song, as the OR and CH sections are seen as the same song - they have the same artist and song name. However, it then selects a section based on the song id _randomly_ (`random.choice`). As both sections (with group `or` & `ch`) are connected to the same song, it randomly selects one of those sections. 

Now, I tried to also add the `filter_by` parameter to that same query:

```py
    def section_from_any_song(self, filter_by={}):
        """Get a random section with a Dutch IP check.

        To ensure appropriate IP restrictions, most rules should use this
        method instead of operating on the playlist directly.
        """

[...]

PR #1113, opened 2024-06-12
Fixed: Also apply filter_by filter on section filter in `section_from_any_song` session model method

drikusroor:
See also this comment: https://github.com/Amsterdam-Music-Lab/MUSCLE/issues/1106#issuecomment-2163369748

This seems to fix the problems in TH5 Kids - Tempo experiment as discussed in #1106, but I'm not completely sure what effect it might have on many other experiments. Maybe other experiments *do* want to pick a random section from the same song? It almost seems like it's the whole point of this method...

Maybe we need something like a `get_specific_section_from_song` method; in which the filter is leading and not a random pick?

Related to #1106

### DIFF
--- a/backend/experiment/rules/toontjehogerkids_5_tempo.py
+++ b/backend/experiment/rules/toontjehogerkids_5_tempo.py
@@ -48,7 +48,7 @@ class ToontjeHogerKids5Tempo(ToontjeHoger5Tempo):
 
         if not section_original:
             raise Exception(
-                "Error: could not find original section: {}".format(tag_original))
+                "Error: could not find original section: {}".format(section_original))
 
         section_changed = self.get_section_changed(
             session=session, song=section_original.song)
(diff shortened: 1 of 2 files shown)


BeritJanssen:
Well caught! The whole `section_from_song` stuff is pretty chaotic at the moment. Not everything is adapted to the fact that we have a separate song model now, and there are no unit tests. Will make an issue for this. The `filter_by` is now applied twice. As far as I can see, that shouldn't be a problem. Say, we filter all sections which have tag "banana", then get the associated song ids, then again, for a given song id, filter out sections with tag "banana" - this does not make much sense in this case, but will get us there, right?

BeritJanssen (approved):

BeritJanssen:
I still pushed a small fix to the output of an exception in thk5 that I discovered while going over the code yesterday.

PR #1122, opened 2024-06-14
Added: Add rel=0 url param to THK embedded youtube video src

drikusroor:
This pull request adds the `rel=0` URL parameter to the embedded YouTube video sources in the code. This parameter should prevent auto-suggested videos from appearing at the end of playback.


BeritJanssen:
This didn't solve the problem, as far as I gathered? Perhaps this is a reason to recommend Vimeo or another paid service in the future.

drikusroor:
> This didn't solve the problem, as far as I gathered? Perhaps this is a reason to recommend Vimeo or another paid service in the future.

Yeah, it seems like it didn't :-( Maybe the UvA needs its own CDN, that would be cool

drikusroor:
Didn't solve el problemo of unwanted auto-suggested videos

PR #1127, opened 2024-06-14
Refactor: Convert API & useGet to TypeScript

drikusroor:


frontend/src/API.ts
```diff
@@ -98,20 +111,18 @@ export const createSession = async ( {experiment, participant, playlist} ) => {
...
-            qs.stringify({
-                playlist_id: playlistId,
-                csrfmiddlewaretoken: participant.csrf_token
-            })
-        )
-        return response.data;
-    } catch(err) {
-        console.error(err);
-        return null;
-    }
```
drikusroor, line 114: NB: This code seems to be unused and I have therefore removed it

frontend/src/components/Experiment/Experiment.test.jsx
```diff
@@ -23,7 +23,7 @@ vi.mock('../../util/stores', () => ({
     __esModule: true,
     default: (fn) => {
         const state = {
-            session: 1,
+            session: { id: 1 },
             participant: 'participant-id',
```
BeritJanssen, line 27: It would be nice to be consistent: either just pass the id as a string / integer, or pass an object for both session and participant.

BeritJanssen (approved):
Not necessary to address in this branch, but the update to the Experiment unit test shows how the stores are somewhat inconsistent at the moment in how they handle sessions (=object) and participants (=string).

frontend/src/components/Experiment/Experiment.test.jsx
drikusroor: They stores already carry objects with ids so it only makes sense that the mock stores do the same.

PR #1142, opened 2024-06-20
Added: Add error message for missing experiment collection

drikusroor:
This will probably fix several Sentry errors related to missing properties on the experiment collection (which is null) when it is not found.

And of course it's a better message to show to users. :-)

### DIFF
--- a/frontend/src/components/ExperimentCollection/ExperimentCollection.tsx
+++ b/frontend/src/components/ExperimentCollection/ExperimentCollection.tsx
@@ -56,6 +56,10 @@ const ExperimentCollection = ({ match }: ExperimentCollectionProps) => {
         );
     }
 
+    if (!loadingExperimentCollection && !experimentCollection) {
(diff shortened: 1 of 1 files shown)


frontend/src/components/ExperimentCollection/ExperimentCollection.tsx
```diff
@@ -56,6 +56,10 @@ const ExperimentCollection = ({ match }: ExperimentCollectionProps) => {
         );
     }
 
+    if (!loadingExperimentCollection && !experimentCollection) {
+        return <p className="aha__error">Experiment collection not found</p>;
```
BeritJanssen, line 60: Might as well drop the "collection" here.

drikusroor, line 60:
```suggestion
        return <p className="aha__error">Experiment not found</p>;
```

PR #1143, opened 2024-06-21
CI: Refactor e2e tests and add smoke tests for production environment

### LINKED ISSUE #1060: Create and run smoke tests that run after a release/deploy to production
Straight after a deploy, it would be nice to have an automated check that will test if the _core_ features of the platform still work.

- Run a designated smoke test suite using the Selenium setup we have already in a Github Actions workflow.
- This smoke test suite tests the following:
  - Does the admin dashboard page return a 200 OK?
  - Does the frontend give a 200 OK?
  - Can it load experiment A (A being the most used / relevant experiment)
  - Can it play audio files and go to the next round?

**Additional context**
- Check [this](https://github.com/Amsterdam-Music-Lab/MUSCLE/tree/7e71ce62b88c905e607f426ef68c8a2e84cbf182/e2e) folder for the current Selenium E2E tests configuration. It's probably best to copy and/or re-use this configuration.
- Check here for the CI/CD part of running the E2E tests https://github.com/Amsterdam-Music-Lab/MUSCLE/blob/0a835cf9b234f439035db5cc656707ea241c591e/.github/workflows/podman.yml#L281-L294

drikusroor:
This PR adds simple short-lived smoke tests that we can use to do a sanity check after a deploy to production. Furthermore, it refactors the e2e suite as a whole by splitting up the *actual* e2e tests from the smoke tests so that we can execute them separately if needed. Both the e2e tests and the smoke tests rely on the same base and execution command, but one can filter which one(s) to run using the parameters `--smoke` and `--e2e`. Omitting the parameters executes both smoke and e2e tests.

The added smoke tests test the following things for now:
- If you go to `/server/admin`, will it show you the admin login page? This tells us that the backend service is online.
- If you go to `/collection/thkids`, does it show you the dashboard with experiments? This tells us that collections can be properly loaded and that the connection to the backend and its database works as well.
  - If so, can you click on an experiment and will it show you the first experiment page? This tells us that experiments can be loaded.

Resolves #1060

This workflow shows that it worked as expected when I temporarily turned off the requirements for the e2e/smoke tests (branch / tag / post deploy / etc.) => https://github.com/Amsterdam-Music-Lab/MUSCLE/actions/runs/9610351545


albertas-jn (approved):

e2e/tests/smoke/test_google.py
BeritJanssen, line 1: Is there a way to test internet connection without relying on elgoog?

BeritJanssen (changes requested):
A teensy and somewhat nitpicky request: most of the committed files don't have an empty last line. While my editor isn't bothered by that, some editors are. That's why GitHub also flags this. Please configure your editor to add an empty last line, so future collaborators/maintainers don't curse over our graves. ;-)

drikusroor:
> A teensy and somewhat nitpicky request: most of the committed files don't have an empty last line. While my editor isn't bothered by that, some editors are. That's why GitHub also flags this. Please configure your editor to add an empty last line, so future collaborators/maintainers don't curse over our graves. ;-)

This is kind of new to me. If we want empty last lines in every file, shouldn't we have an (opinionated) settings file for VS Code at least?

e2e/tests/smoke/test_google.py
drikusroor: Maybe, but why would we not use google? It's probably the service in the world with the most up-time.

BeritJanssen: I thought there might be a non-corporate alternative. But yes, for the foreseeable future, Google probably has the most reliable up-time.

BeritJanssen:
We have an `.editorconfig` which should take care of that. That has `insert_final_newline = true`, as a matter of fact.

e2e/tests/smoke/test_google.py
drikusroor: I've added Cloudflare to the mix just in case

drikusroor:
> We have an `.editorconfig` which should take care of that. That has `insert_final_newline = true`, as a matter of fact.

Ah I see.. VS Code does need a plugin to be able to work with that. Maybe it would be an idea to add it to the "Recommended plugins" in the `.vscode` folder or would that be overkill?

PR #1156, opened 2024-06-26
Delete playlist crashes the admin when section has no song

Evert-R:
This PR returns an empty string on the Sectiojn model if no song is attached


drikusroor (commented):
How is your fix related to deletion of a playlist when you are modifying the stringify method of the `Section` model?

Evert-R:
> How is your fix related to deletion of a playlist when you are modifying the stringify method of the `Section` model?

The admin will list all the sections to be deleted to confirm the action, then it crashes cause there is no song

PR #1164, opened 2024-07-02
Chore: Add missing Django Experiment Phase migrations

drikusroor:
I don't exactly know where these came from, but the migrations for these hadn't been generated and applied yet, so let's do that using this PR.


backend/experiment/migrations/0040_alter_phase_order_alter_phase_randomize.py
```diff
@@ -0,0 +1,23 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('experiment', '0039_auto_20240625_1055'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='phase',
+            name='order',
```
BeritJanssen, line 15: While we're at it, perhaps also call this field "index" instead? May have missed this detail when reviewing the refactor PR.

PR #1167, opened 2024-07-02
Feature/back to collection

### LINKED ISSUE #1153: Remove hard-coded link from ToontjeHoger(Kids) rules
Instead of a hard-coded link to a fixed slug, we want to save the slug of the current collection being played to the Django session, and retrieve from there.

NB: this issue blocks moving ToontjeHoger to the new production server right now: after completing an experiment, going "back to ToontjeHoger" will fail.

BeritJanssen:
Resolves #1153: use the Django session data to set the link to return to the collection. Additional advantage is that also a (purely hypothetical) `participant_id` will be conserved. The branch also does some cleanup by removing the `toontjehoger_home` rules file in the backend and the `ToontjeHoger` frontend component, which as of now aren't used anymore. ✌️


backend/experiment/actions/utils.py
```diff
@@ -6,23 +6,30 @@
...
+    if not collection_slug:
+        return None
 
     if session.participant.participant_id_url:
         participant_id_url = session.participant.participant_id_url
-        redirect_url = f'/collection/{collection_slug}?participant_id_url={participant_id_url}'
+        return f'/collection/{collection_slug}?participant_id_url={participant_id_url}'
     else:
-        redirect_url = f'/collection/{collection_slug}'
+        return f'/collection/{collection_slug}'
```
drikusroor, line 23: This seems like a good function for a unit test or three (no collection slug, session with participant id url, and else)

drikusroor (approved):
LGTM, nice that we don't have to worry about using the "correct" thkids slug anymore

PR #1168, opened 2024-07-02
Added: Add `bonus_points` to block view and use it in Matching Pairs

### LINKED ISSUE #1165: MatchingPairs: in frontend component, set the `total` score from the experiment's `bonus_points` field
Right now, the frontend MatchingPairs component sets the "bonus score" to 100:
https://github.com/Amsterdam-Music-Lab/MUSCLE/blob/79282db665329813aa1ae1fb509b9b72519f034b/frontend/src/components/MatchingPairs/MatchingPairs.jsx#L34

This is not easily configurable. We already have a `bonus_points` field on the `Experiment` model. This should be used instead to initialize the score for MatchingPairs.

Split off from #1163

drikusroor:
This PR adds the `bonus_points` property to block view and uses it in Matching Pairs as the starting value for the total score through the newly added block slice in the zustand store.

Additionally, the `MatchingPairs` component has been converted to TypeScript along with adding interfaces and types and slightly improving the component use of the `index`, which was ambiguously seen and used as a `string` and `number`.

Resolves #1165

### DIFF
--- a/frontend/src/types/Section.ts
+++ b/frontend/src/types/Section.ts
@@ -1,5 +1,4 @@
 export default interface Section {
     id: number;
-    group: string;
     url: string;
 }
--- a/frontend/src/types/Section.ts
+++ b/frontend/src/types/Section.ts
@@ -1,0 +1,5 @@
+export default interface Section {
+    id: number;
+    group: string;
+    url: string;
+}
--- a/frontend/src/util/stores.ts
+++ b/frontend/src/util/stores.ts
@@ -8,16 +8,12 @@ import IBlock from "@/types/Block";
 
 interface BlockSlice {
     block?: IBlock;
-    blockLoading: boolean;
     setBlock: (block: IBlock) => void;
-    setBlockLoading: (blockLoading: boolean) => void;
 }
 
 const createBlockSlice: StateCreator<BlockSlice> = (set) => ({
     block: undefined,
-    blockLoading: true,
     setBlock: (block) => set(() => ({ block })),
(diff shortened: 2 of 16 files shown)


frontend/src/types/Section.ts
```diff
@@ -0,0 +1,5 @@
+export default interface Section {
+    id: number;
+    group: string;
```
BeritJanssen, line 3: The group prop can be removed, here and in the backend serialization. It was added when the evaluation of the matching pairs logic was done in the frontend. Now that this is moved to backend, the gorup prop is not needed anymore.

frontend/src/util/stores.ts
```diff
@@ -4,6 +4,21 @@ import { StateCreator, create } from "zustand";
...
+interface BlockSlice {
+    block?: IBlock;
+    blockLoading: boolean;
+    setBlock: (block: IBlock) => void;
+    setBlockLoading: (blockLoading: boolean) => void;
+}
+
+const createBlockSlice: StateCreator<BlockSlice> = (set) => ({
+    block: undefined,
+    blockLoading: true,
```
BeritJanssen, line 18: Why is it necessary to add `blockLoading` to the store? Is this not sufficient to maintain the state of whether or not this is loading in the component which triggers the request to the block data from the backend?

BeritJanssen (commented):
Nice! Only wondering about the necessity of adding the loading state of `block` to the store.

frontend/src/util/stores.ts
drikusroor: True, it's not even used rn

PR #1169, opened 2024-07-03
Refactor: Refactor Experiment to Block in backend (and partly in frontend)

### LINKED ISSUE #1147: Backend: Rename Experiment to Block
- [ ] move code
  - [ ] create migration file
  - [ ] add data migration (including backwards migration)
  - [ ] create `views.py` and `urls.py`

drikusroor:
This pull request includes refactoring the names of Experiment and GroupedExperiment to Block and GroupedBlock in the backend code. It also updates the frontend code to use the new names.

120 files changed... The backend tests & lints seem to succeed. Furthermore, I did some basic testing by going through my local Toontjehoger Experiment(Collection) and its Blocks (yes, Blocks) and found no problems. Perhaps it'd be a good idea if you did the same? Of course it's always a good idea to backup your local db just in case as the migrations are strong in this one.

Resolves #1147


backend/experiment/rules/base.py
```diff
@@ -152,29 +152,29 @@ def get_questionnaire(self, session, randomize=False, cutoff_index=None):
             ))
         return trials
 
-    def social_media_info(self, experiment, score):
-        current_url = f"{settings.RELOAD_PARTICIPANT_TARGET}/{experiment.slug}"
+    def social_media_info(self, block, score):
```
BeritJanssen, line 155: I think eventually the `social_media_info` will be associated with the `Experiment` object. Just saying that this will need another refactor later on.

drikusroor: I could add a comment that deprecates this function. But I could also create a [deprecation decorator](https://stackoverflow.com/a/48632082/4496102) that we can use in the future for deprecating functions which will then also warn us of not using deprecated functions and point is in to the correct way of doing things. What do you think?

backend/experiment/templates/collection-dashboard.html
```diff
@@ -132,20 +132,20 @@ <h1>Experiments in collection: {{ collection.name }}</h1>
...
+                </p>
             </td>
-            
+
             <td class="table-sessions">
                 <input type="button" value="Sessions" id="show-sessions-{{ exp.id }}" />
                 <script>
                     // Show all session for this experiment in popup
-                    showExperimentSessions({{ exp.id }})                                        
+                    showExperimentSessions({{ exp.id }})
```
BeritJanssen, line 142: should this be `showBlockSessions`?

BeritJanssen:
So far, I haven't found anything broken after this major refactor. Nice work! I did click a bit through the admin interface and found the following leftover references to experiment instead of block (in the custom templates for the ExperimentCollection "dashboard" view, as well as the inline template for setting question series for a block):
Screenshot 2024-07-09 at 13 33 43
Screenshot 2024-07-09 at 13 34 21

PR #1174, opened 2024-07-05
Refactor: Refactor and convert Question-related components to TypeScript

drikusroor:
This pull request includes several refactorings and updates to the codebase. The main changes are:

- Converted the `Question` component to TypeScript and updated related tests.

- Converted the `FeedbackForm` component to TypeScript.

- Updated the usage of the word "index" to refer to the index in the list of questions.

- Added tests for the `TextRange` component.

- Converted the `TextRange` component to TypeScript.

- Converted the `_RangeTitle.jsx` component to TypeScript and updated related imports.

- Converted the `_RangeLimits.jsx` component to TypeScript.

- Converted the `label` component to TypeScript.


frontend/src/components/Question/__snapshots__/Question.test.tsx.snap
BeritJanssen, line 1: What is this file? Ought it to be committed?

BeritJanssen (commented):
Nice! Check though whether that snapshot should be in version control.

frontend/src/components/Question/__snapshots__/Question.test.tsx.snap
drikusroor: No, I don't think so... It's one of those snapshots you can use to test if components aren't changing without noticing. But we don't do that form of testing (anymore?)

PR #1175, opened 2024-07-05
Refactor: Refactor Profile component to TypeScript

drikusroor:
The Profile component is currently written in JavaScript and this pull request converts it to TypeScript. This will improve type safety and make the codebase more maintainable.

Fix "points" string variable in Profile component (see screenshot below) as it tried to get the "points" string from data directly instead of from `data.messages`

image

### DIFF
--- a/frontend/src/components/Profile/Profile.tsx
+++ b/frontend/src/components/Profile/Profile.tsx
@@ -44,10 +44,8 @@ interface ProfileViewProps {
         points: string;
     };
     scores: {
-        // This might need to be renamed to block_slug and experiment_slug
-        // after the great refactor of Summer 2024
-        experiment_slug: string;
-        experiment_name: string;
+        block_slug: string;
+        block_name: string;
         score: number;
         date: string;
         rank: {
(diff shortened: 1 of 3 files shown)


frontend/src/components/Profile/Profile.tsx
```diff
@@ -76,7 +97,7 @@ export const ProfileView = (data) => {
                                         {score.block_name}
```
BeritJanssen, line 97: This line seems to indicate that the rename to `block_name` has already been applied?

frontend/src/components/Profile/Profile.tsx
```diff
@@ -37,7 +36,29 @@ const Profile = () => {
...
+        title: string;
+        summary: string;
+        continue: string;
+        points: string;
+    };
+    scores: {
+        // This might need to be renamed to block_slug and experiment_slug
+        // after the great refactor of Summer 2024
+        experiment_slug: string;
+        experiment_name: string;
```
BeritJanssen, line 50: Check if this shouldn't be already be renamed before merging.

frontend/src/components/Profile/Profile.tsx
drikusroor: Yeah, that's because this branch has been rebased already on the latest develop changes

frontend/src/components/Profile/Profile.tsx
drikusroor: Good point. This is exactly why I placed the comment :-)

PR #1189, opened 2024-07-18
Chore: Improve security & fix some linting errors

drikusroor:
I ran [Vulture](https://github.com/jendrikseipp/vulture) and [Bandit](https://github.com/PyCQA/bandit) on the backend project and found some security / linting issues which I fix here.

### DIFF
--- a/backend/experiment/admin.py
+++ b/backend/experiment/admin.py
@@ -1,7 +1,6 @@
 import csv
 from zipfile import ZipFile
 from io import BytesIO
-from django.utils.safestring import mark_safe
 
 from django.conf import settings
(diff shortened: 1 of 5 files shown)


albertas-jn (changes requested):
I don't think it is a good idea to remove `parent_obj=None`, because it is in the signature of `django-inline-actions` action method.

drikusroor:
> I don't think it is a good idea to remove `parent_obj=None`, because it is in the signature of `django-inline-actions` action method.

Is there a way I can tell python/linting that I am declaring the parameter but not using it on purpose? 🤔

Issue #973, opened 2024-04-30
Remove `first_round` from rules files


BeritJanssen:
After #1132, consent will always be shown via the `consent field` tied to the `Experiment` parent.

Calling the endpoint `block/{id}` will do the following:
- [x] check the participant
- [x] start a session with the current / new participant and the current block, and return its id to the frontend See also #1191

After that, the block will be handled through the `session/{id}/next_round` endpoint
Steps:
**Backend**
- [x] remove Consent actions from all `first_round` methods
- [x] move `Playlist` views to `next_round` (only for those rules where it's conceivable we may have multiple playlists tied to them)
- [x] move every `Explainer` to `next_round`
- [x] remove `first_round` methods
- [ ] do not pass the `next_round` array to the `block` request
~~- [ ] migrate existing Consent forms to Experiment level if applicable~~ -> #1211

**Frontend**
- [ ] remove `Consent` as an option from `Block`
- [ ] adjust `Block` so that it doesn't expect a `next_round` array in the response; instead go to `session/next_round` once created

drikusroor:
> After #1132, consent will always be shown via the `consent field` tied to the `Experiment` parent.

I mean... technically the field is there, but do we have functionality that also checks if the participant has accepted the consent form before starting the experiment / block and shows the consent form if that's not the case? If not, perhaps we should make that before doing this refactor. 🤔

PR #1191, opened 2024-07-19
Refactor: Start phasing out `first_round` and create session in `get_block`

drikusroor:
This pull request includes a refactoring of the `get_block` function to phase out the `first_round` methods and create a session directly in the `get_block` function. If the `first_round` does not exist anymore in a block's rules file, it will call `next_round` directly. The frontend has also been updated to take the session_id from the get_block response. `get_block` will still call `first_round` if it does exist on the block's rules file in question. This allows us to phase it out over time and prevent this PR from becoming a monster to review.

Additionally, tests have been fixed and added after the refactoring of the `first_round` methods and session creation in `get_block`.

Finally, the `Block` and `Trial` component have been partially configured to TypeScript and some types have been updated/added to accomodate this.

Related to #973

### DIFF
--- a/backend/session/models.py
+++ b/backend/session/models.py
@@ -7,14 +7,18 @@ from django.utils import timezone
 class Session(models.Model):
     """Experiment session by a participant"""
 
-    block = models.ForeignKey('experiment.Block', on_delete=models.CASCADE, blank=True, null=True)
-    participant = models.ForeignKey('participant.Participant', on_delete=models.CASCADE)
-    playlist = models.ForeignKey('section.Playlist', on_delete=models.SET_NULL,
-                                 blank=True, null=True)
+    block = models.ForeignKey(
+        "experiment.Block", on_delete=models.CASCADE, blank=True, null=True
+    )
+    participant = models.ForeignKey("participant.Participant", on_delete=models.CASCADE)
+    playlist = models.ForeignKey(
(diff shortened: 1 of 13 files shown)


backend/session/models.py
```diff
@@ -75,15 +80,25 @@ def load_json_data(self):
...
 
+    def save_json_data_key(self, key: str, value):
+        """Set json data key"""
+        data = self.load_json_data()
+        data[key] = value
+        self.save_json_data(data)
+
+    def load_json_data_key(self, key: str, default=None):
+        """Get json data key"""
+        return self.load_json_data()[key] if key in self.load_json_data() else default
```
BeritJanssen, line 91:
```suggestion
        return self.load_json_data().get(key, default)
```

backend/session/models.py
```diff
@@ -75,15 +80,25 @@ def load_json_data(self):
         """Get json data as object"""
         return self.json_data if self.json_data else {}
 
+    def save_json_data_key(self, key: str, value):
```
BeritJanssen, line 83: What are these two methods used for?

BeritJanssen (approved):
Nice to do this in steps, indeed. Only puzzed by the `save / load_json_data_key` functions, which I couldn't find being used.

backend/session/models.py
drikusroor: I was using them, but then I think I didn't need them anymore so I'll remove the methods as well here.

PR #1194, opened 2024-07-22
Refactored: Remove GroupedBlock model and relate Block directly to Phase

### LINKED ISSUE #1134: Proposal: remove `GroupedBlock` model
Instead of using a separate `GroupedBlock` model, we might already give the `Block` model itself an `index` field.

*Advantage*: admin users need to navigate less models
*Disadvantage*: `Block` objects cannot be reused, using the same block in different experiment phases will mean users will have to create duplicate data

As the admin form for `Experiment` should eventually support creating inline `Block` objects, in practice this will mean admin users will probably create new objects.

Steps:
- [x] add `index` field to `Block` model
- [x] create migration file
- [x] write a data migration which moves indices from `GroupedBlock` to `Block.index` (including backwards migration), and optionally creates new objects if there are multiple `GroupedBlock`s linked to the same `Block`
  - [x] try to avoid slug conflicts whenever blocks have to be duplicated 
  - [x] make sure index information is copied in migration too
- [x] remove `GroupedBlock` model
- [x] create migration file to delete the model from the database

drikusroor:
Resolves #1134


backend/experiment/tests/test_admin_experiment.py
```diff
@@ -246,24 +251,40 @@ def test_related_series_with_series(self):
...
+            name='Test Group',
+            index=1,
+            randomize=False,
+            dashboard=True,
+            series=series
+            )
         blocks = self.admin.blocks(phase)
         self.assertEqual(blocks, "No blocks")
 
     def test_blocks_with_blocks(self):
```
BeritJanssen, line 263: I guess this test should be called `test_experiment_with_blocks`, and that the first `experiment` was renamed when finding & replacing.

BeritJanssen (approved):

backend/experiment/tests/test_admin_experiment.py
drikusroor: Excellent find!

PR #1197, opened 2024-07-22
Refactored: Refactor Listen and ListenFeedback components to TypeScript

drikusroor:


BeritJanssen (approved):
Nice to have this done! Though I think that in the long run we should also refactor the Circle / Listen / ListenFeedback components. Especially the latter one I think has become obsolete (it used to be part of the Hooked game), it seems to be used in Preload, but could certainly be replaced by other components.

PR #1199, opened 2024-07-23
Chore(deps): Update npm dependencies react-router, react-router-dom, @sentry/react & next-share to latest version for frontend

drikusroor:
Migrating from `react-router(-dom)` v5 to v6 requires some changes:

- `withRouter` is no longer available and should be replaced by `useNavigate` or `useParams` (depending on what you need)
- `Redirect` no longer exists and should be replaced by a custom `Redirect` component that uses the `navigate` function from `useNavigate`.
- `Switch` should be replaced by `Routes`
- Nested `Route`s (not `Routes`) don't use a full absolute route (`/toontjehoger/about`) but a route relative to its parent route (`/about` relative to its parent route `/toontjehoger` aka `/:slug`)
- In order to make nested routes available, you need to add `/*` to the route. In other words: `/:slug/*` will enable you to use the nested child routes for the Experiment page like Dashboard & About. If you only use `/:slug`, the child routes will never get hit and the Experiment page won't show the Dashboard or About child pages.
- A `Route` now requires you to set the component through the element prop as jsx, so not `component={Experiment}` but `element={<Experiment />}`


frontend/src/components/Experiment/Experiment.test.tsx
```diff
@@ -125,9 +141,10 @@ describe('Experiment', () => {
 
     it('shows a footer if a theme with footer is available', async () => {
         mock.onGet().replyOnce(200, { dashboard: [blockWithAllProps], nextBlock: block1, theme });
+        console.log("Knalvis deluxe!")
```
BeritJanssen, line 144: Stray log statement

BeritJanssen (approved):

frontend/src/components/Experiment/Experiment.test.tsx
drikusroor:
```suggestion
```

Issue #847, opened 2024-03-12
Clean up and document `session.models`


BeritJanssen:
The current `Session` model comes with many utility functions to filter sections, check the number of finished rounds, etc. Many of these methods may not be used anymore, or have missing / obsolete docstrings. Any Django-specific query syntax (`filter_by`) in many of the section selection methods should also still be documented, including a link to [Django's query documentation](https://docs.djangoproject.com/en/4.2/topics/db/queries/).

PR #1200, opened 2024-07-23
Feature/cleanup session

BeritJanssen:
Working towards #847: instead of having different section getters both on session and section models, everything is now changed to the following situation:
- if we're interested in sections that have been played or not in a given session, we can do so via `get_(un)used_song_ids`
- to retrieve a section from a playlist belonging to specific songs and/or adhering to specific filter conditions, use `playlist.get_section`

This branch also includes a major refactor of the Hooked rules, meaning that the Eurovision / Christmas experiment mainly differ now in the conditions they impose on the selected sections for the `heard_before` rounds.


Evert-R (changes requested):
Nice to see this cleaned up. Spotted one thing:

The categorization rules file still uses:
`section = session.section_from_song(this_section)` *line 477*

This should now be:
`section = session.playlist.get_section(song_ids=[this_section])`

PR #1206, opened 2024-07-30
Changed experiment to block

Evert-R:
Changed the term experiment to block in csv export screen

### DIFF
--- a/backend/experiment/templates/csv-export.html
+++ b/backend/experiment/templates/csv-export.html
@@ -38,7 +38,7 @@
         margin-right: 2em;
     }
 </style>
-<h1>Export results for experiment: {{experiment.name}}</h1>
+<h1>Export results for block: {{experiment.name}}</h1>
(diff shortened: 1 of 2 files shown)


backend/experiment/templates/csv-export.html
```diff
@@ -38,7 +38,7 @@
         margin-right: 2em;
     }
 </style>
-<h1>Export results for experiment: {{experiment.name}}</h1>
+<h1>Export results for block: {{experiment.name}}</h1>
```
drikusroor, line 41: It still says `experiment.name` though 🤔

drikusroor (approved):
🫡

backend/experiment/templates/csv-export.html
Evert-R: Yes, you are a bit too fast ;-)
Now called `exp_block.name` as `block` is a reserved name in django templates.

Issue #1204, opened 2024-07-29
Translate Experiment using a related "TranslatedContent" model


drikusroor:
**Is your feature request related to a problem? Please describe.**
Right now, we don't have a clear way to provide translated / localised / internationalised experiments to users. The only option we currently have is to create a "new Experiment" for a new language and write the title, description, etc. in that language. However, the app's non-experiment-related or non-configurable content (button's, links, translation keys in the experiment blocks)  might still show up in a different language.

**Describe the solution you'd like**
A better way, as discussed on Tuesday July 23, might be to create a new model (named `ExperimentTranslatedContent` or something) which contains all the textual content of an experiment for a given language. In pseudocode, that model would look something like this:

```
ExperimentTranslatedContent
- experiment (FK)
- index (integer)
- language ('en' | 'nl' | 'zh', etc.)
- title
- description
- content
- about_content
```

**Which content should then be shown to the user?**

Based on the preference from the client, in the form of a URL parameter (?lang=jp), we can serialise the experiment with the right language content and get a fallback in case the chosen language's content is not available.

**Which fallback should then be chosen?**

We could use the index to arrange the fallback priority order. In other words, if an ETC (Exp. Translated Content) for Dutch (nl) is listed on top (index = 0), it can be the fallback for an English ETC with index = 1.

Alternatively, we could think about a more elaborate system of fallbacks in the future if needed. An option could be a fallback_for property with an array of languages (en, nl) for an ETC and make Django check that list if it cannot find the content for a preferred language. E.g. the user prefers Catalan, but the Spanish ETC has `fallback_for=[ca,pt]` and therefore falls back to the Spanish ETC.

## Requirements

- Create a model named `ExperimentTranslatedContent` (see pseudocode above)
- Make sure to use `django-nested-admin` to create a nested inline form.
- Try to make at least _one_ related `ExperimentTranslatedContent` mandatory in order to create/save an experiment. (An Experiment without content should not be possible).
- Serialise the `get_experiment`'s content according to the preferred language and make sure the json structure stays the same (so the frontend needs no revising)
- The language in which the experiment will be returned should be based on the `Accept-Language` header supplied by the browser.
- If an experiment's language content is not available for the language from the `Accept-Language` header, fallback to the _highest_ available experiment language content bases on the index (index=0 would be first fallback).
- The language ultimately selected (preferred or fallback) should be used for _all_ (both experiment- and non-experiment-related) content by setting the language cookie to that language.

[...]

PR #1208, opened 2024-08-02
Added: Add translated_content model and update Experiment model

drikusroor:
This PR adds the `ExperimentTranslatedContent` model to the application, along with its related functionality. It has the following properties:

## Model

- `experiment` - FK to the `Experiment` model
- `language` - The language code for the content's language
- `index` - Sorting property, the content with the lowest index will be considered the fallback language/content (for now)
- `title` - This was originally the Experiment's title
- `description` - This was originally the Experiment's description
- `consent` - This was originally the Experiment's consent
- `about_content` - This was originally the Experiment's about_content

The PR also adds some helper methods to the `Experiment` model, `get_fallback_content` and `get_translated_content`. These are used to get the appropriate content for the given context.

https://github.com/Amsterdam-Music-Lab/MUSCLE/blob/5d90e4b27b777530ad5ad19d6a0fe166fd25550e/backend/experiment/models.py#L68-L87

## Serialisation & active language handling

The languages of the configured `ExperimentTranslatedContent`s for an `Experiment` determine the possible languages in which an experiment can be used. These languages will also determine the languages used at the `SocialMediaConfig`, `ThemeConfig`, and `Block` level. Right now, what happens is that the user has a preferred language that is used to get the appropriate content and fall back to the experiment's fallback language if it is not available. Django has the `get_language` helper function to determine the user's preferred language. This language is used to find the appropriate `ExperimentTranslatedContent`. If it cannot find it, it will take the experiment's translated content with the lowest `index` number. If that doesn't exist either, it will throw an error as an `Experiment` needs _a_ content in order to work. (We might want to require at least one translated content per experiment during creation and updating.)

The language of the `ExperimentTranslatedContent` (preferred or fallback) will be used as the `active` language for the whole experiment, its related entities, and the non-experiment related content using Django `activate(language)` function (which was used earlier at the `Block` level depending on the configured language there.

https://github.com/Amsterdam-Music-Lab/MUSCLE/blob/066b9381fd64d57a502b5516a93212f207cd9916/backend/experiment/views.py#L128-L136

### Consent

The retrieval of the consent form works slightly different. If a consent form cannot be found in the preferred language, but _does_ exist at the fallback level, the consent form from the fallback language will be used. This gives the user flexibility in configuring consent form: Configure consent forms in all languages, configure consent forms in only one or a limited selection of the experiment languages, or configure no consent forms at all.

## Things to do after merging this PR

The migrations copy the Experiment's content to a newly created ExperimentLanguageContent, with the language set to English. This means that we have to go over all experiments after the migration and set the language to Dutch or Chinese if that's the Experiment's original language.

Related to #1204


backend/experiment/migrations/0050_migrate_experiment_content_to_translatedmakemigrations.py
```diff
@@ -0,0 +1,50 @@
...
+        if not primary_translation:
+            continue
+
+        experiment.name = primary_translation.name
+        experiment.description = primary_translation.description
+        experiment.consent = primary_translation.consent
+        experiment.about_content = primary_translation.about_content
+        experiment.save()
+
+    ExperimentTranslatedContent.objects.filter(language="en").delete()
```
BeritJanssen, line 40: Why not delete all models in this reverse migration?

BeritJanssen (approved):

backend/experiment/migrations/0050_migrate_experiment_content_to_translatedmakemigrations.py
drikusroor: This was easier. When I was working on migrating Blocks from a FK to a OneToOne I had a lot of issues with migrations, especially if I needed to change something in the migrations.

drikusroor: I could still squash or re-create them, now that it probably won't need any changes anymore?

BeritJanssen: I still don't quite understand why it would be easier to just delete the English language objects, instead of deleting all `ExperimentTranslatedContent` objects. (Sorry, I erroneously wrote "models" before, of course you shouldn't just go ahead and delete the models without data migration.)

drikusroor: Oh, I see what you mean now. Yeah, deleting all objects would make more sense, thanks!

Issue #1210, opened 2024-08-05
Make an experiment language content validator method. This method checks if all experiment and non-experiment related content is available in the languages for which the content has been added to the Experiment level. If this is not the case, it should warn and guide the user to the right places to add the missing language content for the related entities (blocks, theme, question).


drikusroor:

drikusroor:
Closed per #1214

PR #1214, opened 2024-08-05
Added: Validate experiment in experiment admin and fix non-existent consent upload path slug

drikusroor:
This pull request adds an experiment validation method that results into showing "remarks" in the experiment overview page. Right now it only checks whether there is at least one translated content related to the experiment and whether it has a consent form. If there is no translated content, it shows a warning in the form of a badge. If there's no consent form, it shows an information badge. If there are no problems, it shows a success badge.

Later, this method can be expanded with checking whether all supported experiment languages (based on the experiment's translated contents) are represented in the other layers as well (block, theme config, and so on).

## Screenshot

image

Related to #1210


backend/experiment/admin.py
```diff
@@ -349,6 +356,55 @@ def dashboard(self, request, obj, parent_obj=None):
...
+
+        # Check if there is any translated content
+        content = obj.translated_content.first()
+
+        if not content:
+            remarks_array.append(
+                {
+                    "level": "warning",
+                    "message": "📝 No transl. content",
+                    "title": "Please add at least one translated content to this experiment.",
```
BeritJanssen, line 370:
```suggestion
                    "title": "Add content in at least one language to this experiment.",
```

backend/experiment/admin.py
```diff
@@ -349,6 +356,55 @@ def dashboard(self, request, obj, parent_obj=None):
...
+            )
+
+        has_consent = obj.translated_content.filter(consent__isnull=False).exclude(consent="").first()
+
+        if not has_consent:
+            remarks_array.append(
+                {
+                    "level": "info",
+                    "message": "📋 No consent form",
+                    "title": "Please add at least one consent form to this experiment.",
```
BeritJanssen, line 381: It's not always desired for consent forms to be present (e.g., when users come in from external links with a `participant_id`).
```suggestion
                    "title": "You may want to add a consent form (approved by an ethical board) to this experiment.",
```

BeritJanssen (approved):

drikusroor:
> The title of this PR is "Fix slug and..." - which changes adjust the slug?

When uploading a consent form file, it tries to save it in a folder whose name is based on the "instance"'s slug. A translated content has no slug, so I've updated that function to look for the slug on its parent experiment.

## Old version of function

https://github.com/Amsterdam-Music-Lab/MUSCLE/blob/93a3e1c7531d6bf9b7c06aa01a7857ff21344329/backend/experiment/models.py#L20-L24

## Commit that fixes it

https://github.com/Amsterdam-Music-Lab/MUSCLE/pull/1214/commits/783547dfb56f74996385f7dc613222ea4109425a#

PR #1218, opened 2024-08-06
fix: missing max_length of TextQuestion

BeritJanssen:
I discovered this while testing the Speech2Song experiment: `TextQuestion` types wouldn't display anymore, since the default value for `max_length` is `None` instead of a number. This teensy PR fixes that.

### DIFF
--- a/backend/experiment/actions/form.py
+++ b/backend/experiment/actions/form.py
@@ -59,7 +59,7 @@ class NumberQuestion(Question):
 class TextQuestion(Question):
     def __init__(self, input_type='text', max_length=64, **kwargs):
         super().__init__(**kwargs)
-        self.max_length = max_length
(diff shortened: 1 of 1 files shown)


backend/experiment/actions/form.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, input_type='number', min_value=0, max_value=120, **kwargs):
 
 
 class TextQuestion(Question):
-    def __init__(self, input_type='text', max_length=None, **kwargs):
+    def __init__(self, input_type='text', max_length=64, **kwargs):
         super().__init__(**kwargs)
         self.max_length = max_length
```
drikusroor, line 62:
```suggestion
        self.max_length = max_length # the maximum length of the question's answer in characters
```

Issue #1221, opened 2024-08-09
🐛 [BUG] - Warning: A props object containing a "key" prop is being spread into JSX in `Block.tsx`


drikusroor:
### Description

```
stderr | src/components/Block/Block.test.tsx > Block Component > renders with given props
Warning: A props object containing a "key" prop is being spread into JSX:
  let props = {key: someKey, block: ..., participant: ..., loadingText: ..., onResult: ..., onNext: ..., playlist: ..., view: ..., button_label: ...};
  <Info {...props} />
React keys must be passed directly to JSX without using spread:
  let props = {block: ..., participant: ..., loadingText: ..., onResult: ..., onNext: ..., playlist: ..., view: ..., button_label: ...};
  <Info key={someKey} {...props} />
    at Block (/client/src/components/Block/Block.tsx:53:55)
    at Router (/client/.yarn/__virtual__/react-router-virtual-62af35b638/2/root/.yarn/berry/cache/react-router-npm-6.25.1-63bb35966d-10c0.zip/node_modules/react-router/dist/umd/react-router.development.js:1163:17)
    at MemoryRouter (/client/.yarn/__virtual__/react-router-virtual-62af35b638/2/root/.yarn/berry/cache/react-router-npm-6.25.1-63bb35966d-10c0.zip/node_modules/react-router/dist/umd/react-router.development.js:1058:7)
Warning: findDOMNode is deprecated and will be removed in the next major release. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://reactjs.org/link/strict-mode-find-node
    at Transition (/client/.yarn/__virtual__/react-transition-group-virtual-4430b7cd58/2/root/.yarn/berry/cache/react-transition-group-npm-4.4.5-98ea4ef96e-10c0.zip/node_modules/react-transition-group/cjs/Transition.js:135:30)
    at CSSTransition (/client/.yarn/__virtual__/react-transition-group-virtual-4430b7cd58/2/root/.yarn/berry/cache/react-transition-group-npm-4.4.5-98ea4ef96e-10c0.zip/node_modules/react-transition-group/cjs/CSSTransition.js:120:35)
    at div
[...]

PR #1223, opened 2024-08-09
Fixed: Fix warning about key in spread props and make key explicit

drikusroor:
Resolves 🐛 [BUG] - Warning: A props object containing a "key" prop is being spread into JSX in `Block.tsx` #1221

Propose removing unused data argument from setView function in Playback.tsx

## Question to reviewer

Do we even need to pass a `key` to the children components? Perhaps we can simplify it and remove it.

### DIFF
--- a/frontend/src/components/Playback/Playback.tsx
+++ b/frontend/src/components/Playback/Playback.tsx
@@ -40,10 +40,8 @@ const Playback = ({
     const activeAudioEndedListener = useRef<() => void>();
     const [state, setState] = useState<PlaybackState>({ view: PRELOAD });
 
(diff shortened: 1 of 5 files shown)


BeritJanssen:
TBH, I never quite understood the necessity of the `key` set in the `Block` component. Does it have something to do with that every element in a template needs its unique key? Could we be getting away with renaming `view` to `key` instead?

Indeed, child components don't need the key, or view prop for that matter as far as I can see.

frontend/src/components/Playback/Playback.tsx
BeritJanssen, line 1: Removing data argument is fine with me.

BeritJanssen (approved):

PR #1227, opened 2024-08-21
Added: Add multi-language`BlockTranslatedContent` in relation to `Block`

drikusroor:
This PR does the following:
- It adds a `BlockTranslatedContent` model with an FK to `Block`
  - with the properties `block`, `language`, `name`, `description`
- It adds a data migration that migrates the data currently in the `Block` (like name & description) to a `BlockTranslatedContent`with the same language as the block (or defaults to English)
- It that migration succeeded, it removes `consent`, `hashtag`, `language` and `url` from the `Block model.
- I kept `name` and `description` for now as it might be convenient for the experiment administrator. They aren't used by anything afaik
- I added a small javascript fix that makes sure tabular inline headings are displayed correctly as there currently [is a bug](https://github.com/theatlantic/django-nested-admin/pull/259) in `django-nested-admin` whose fix is not yet released.

# Video of experiment form

https://github.com/user-attachments/assets/a06509d1-8051-4050-bf5b-aa14c989ee94


backend/experiment/migrations/0054_migrate_block_content.py
```diff
@@ -0,0 +1,89 @@
...
+        language = block.language if hasattr(block, "language") and block.language else "en"
+
+        BlockTranslatedContent.objects.create(
+            block=block,
+            language=language,
+            name=block.name,
+            description=block.description,
+        )
+
+        if block.phase:
```
BeritJanssen, line 23: This part of the migration is also in #1240 - that also includes a reverse migration to remove the `Experiment` and `Phase` objects that would be created here. The condition for removing them, i.e., checking for a match to the auto-generated name of the phase, may cause errors though, but when I tested the migration locally, it did prove handy to roll this back. So *either* we merge #1240 first and remove this part here, or include a backwards migration for the created `Experiment`, `ExperimentTranslatedContent`, and `Phase` objects, and close #1240. (The latter is probably easier.)

backend/experiment/rules/base.py
```diff
@@ -114,66 +109,78 @@ def rank(self, session, exclude_unfinished=True):
...
+                Trial(
+                    title=_("Questionnaire %(index)i / %(total)i") % {"index": index + 1, "total": open_questions},
+                    feedback_form=Form([question], is_skippable=question.is_skippable),
+                )
+            )
         return trials
 
     def social_media_info(self, block, score):
-        ''' ⚠️ Deprecated. The social media info will eventually be present on the Experiment level, not the Block level. '''
+        """⚠️ Deprecated. The social media info will eventually be present on the Experiment level, not the Block level."""
```
BeritJanssen, line 158: Can this docstring go now?

backend/experiment/static/block_admin.js
```diff
@@ -39,7 +42,34 @@ document.addEventListener("DOMContentLoaded", (event) => {
...
-            location.reload()
+            location.reload();
         }
     }
 })
+
+/** Function to fix the headings for tabular inline forms
+  * @todo TODO: Remove this `fixHeadings` function once the issue with headings is fixed in `django-nested-admin`.
+  * - https://github.com/theatlantic/django-nested-admin/issues/261
+  * - https://github.com/theatlantic/django-nested-admin/pull/259
```
BeritJanssen, line 53: It's now been merged!

backend/experiment/tests/test_admin_experiment.py
```diff
@@ -228,7 +228,7 @@ def setUp(self):
     def test_related_experiment_with_experiment(self):
```
BeritJanssen, line 228: Should the name of this test be adjusted? `related_experiment_with_experiment` is not clear, I'm guessing we're testing blocks & phases here?

backend/experiment/static/block_admin.js
drikusroor: Ah yes! It wasn't mentioned anywhere on Github though 🙄

backend/experiment/rules/base.py
drikusroor: I still see this method being called from `hooked`, `matching_pairs` and `thats_my_song` though 🤔 Are you sure it's safe?

image

backend/experiment/models.py
```diff
@@ -112,29 +117,31 @@ class Block(models.Model):
     phase = models.ForeignKey(Phase, on_delete=models.CASCADE, related_name="blocks", blank=True, null=True)
     index = models.IntegerField(default=0, help_text="Index of the block in the phase. Lower numbers come first.")
     playlists = models.ManyToManyField("section.Playlist", blank=True)
+
+    # TODO: to be deleted?
```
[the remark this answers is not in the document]
drikusroor: I was wondering though: What will be the name of a Block that has no Translated Content? How should we display it in the form/list?

backend/experiment/rules/base.py
drikusroor: As discussed, in the meeting of september 3rd, we cannot remove this _yet_ as we first have to decide on how to deal with social media translations and then implement them.

backend/experiment/models.py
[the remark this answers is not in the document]
BeritJanssen: Hmm, perhaps just the object name with id (which would be Django's default)?

Issue #1138, opened 2024-06-19
i18n of text in the database


albertas-jn:
Think about i18n of text in the database, e.g. questions

BeritJanssen:
I think we've thought about this, and come up with solutions. We should still document them, though.

PR #1228, opened 2024-08-21
Creating custom questions from admin

### LINKED ISSUE #1082: Allow admin users to create new questions in admin
Add possibility to create new questions in admin, in addition to existing hardcoded default questions.

albertas-jn:
Go to Admin - Questions - Add Question and have fun!

I did not add all questions available in `experiment.actions.form.py` because:

* `DropdownQuestion` seems to be the same as `ChoiceQuestion` with `DROPDOWN` view. Not used anywhere in the code except for one test.
* `RadiosQuestion` seems to be the same as `ChoiceQuestion` with `RADIOS` view.
* `ButtonArrayQuestion` same as  `ChoiceQuestion` with `BUTTON_ARRAY` view.
* `RangeQuestion` seems not to work at all.

`BooleanQuestion` looks very similar to  `ChoiceQuestion` with `BUTTON_ARRAY` view. But because it adds a different style, I included it here.

Closes #1082


albertas-jn:
Another change that was made is to leave the creation of questions in the database to the bootstrap command. Before the change, they were also created by the `0002_add_question_model_data.py` data migration, but it is actually not necessary. I disabled that migration, we can perhaps delete  it as well. Then bootstrap command was added to all tests that process questions. Bootstrap also creates superuser, so `DJANGO_SUPERUSER_USERNAME` and `DJANGO_SUPERUSER_EMAIL` had to be added to `.env-github-actions`.

backend/participant/tests.py
```diff
@@ -6,12 +6,14 @@
...
 from session.models import Session
 from result.models import Result
+from django.core.management import call_command
 
 
 class ParticipantTest(TestCase):
 
     @classmethod
     def setUpTestData(cls):
+        call_command('bootstrap')
```
BeritJanssen, line 16: Why is it necessary to call the bootstrap command? Wouldn't it be sufficient to directly call the `create_default_questions` function here?

albertas-jn: Yes, you're right, `create_default_questions` should be sufficient.

BeritJanssen (changes requested):
While I like the fact that this implementation adds little extra lines of code, I do think it takes some shortcuts that will come back to bite us later on: e.g., implementing choices as a flat text field, and requiring the user to enter the text in a certain way, will make it more difficult to add translations for choices, and to edit / delete choices if required.

I would suggest a setup as follows:
```python
class Question(models.Model):
   key = models.SlugField(...) # requires migration, but will ensure we don't allow spaces etc. here
   question = models.CharField(...) # <- should be translatable later on
   explainer = models.CharField(...) # <- should be translatable later on
   editable = models.BooleanField(...)
   view = models.CharField(choices=VIEWS)

class ChoiceQuestion(models.Model):
   question = models.OneToOneField(Question, on_delete=models.CASCADE)
 
class Choice(models:Model):
   choice_question = models.ForeignKey(ChoiceQuestion)
   choice_text = models.CharField(...) # <- this should be translatable later on

class LikertQuestion(models.Model):
   question = models.OneToOneField(Question, on_delete=models.CASCADE)
   scale_steps = models.IntegerField(default=7) # make it possible to choose other values than 5 or 7 here!
    scoring_rule = models.CharField(...)

class NumberQuestion(models.Model):
   question = models.OneToOneField(Question, ...)
   min_value = models.FloatField(default=0)
   max_value = models.FloatField(default=120)

```
In other words, my preference here would be to create objects dedicated to a specific question type. That is much more wordy, of course, but has the advantage of leaner objects in the database and smaller, less confusing forms for the user (they won't get to see a `choices` text field when editing a Likert question: this happened when I was testing; the logic in `forms.py` to select fields presented in the form doesn't seem to work). The admin form could then still be a dropdown in which the user gets to enter a question key, question text, (optional) explainer, and view. In the following step, they would get to see a tailored form for the view. Indeed, `RADIOS / CHECKBOXES / DROPDOWN / AUTOCOMPLETE ` would all be a `ChoiceQuestion`, but we would still need to register the specific view / widget the user would like to see rendered in the frontend.

Adding `Choice` objects through a `ForeignKey` would mean the user would need to click a '+' icon in the admin interface to add extra choices. I left out the `key` here, because we can autogenerate keys.

Two additional / potential problems I see at the moment:
- for the questions which are defined in the code base: will they still be translated after entering them in the database?
- while questions defined in the code base are not `editable`, they can still be deleted by the user

albertas-jn:
It certainly could be done the way you suggest.
* `choices` field for LikertQuestion is shown on purpose, because you can enter custom choices for this question.
* `choices` can be used  by not only by `ChoiceQuestion`, but also `LikertQuestion`, `BooleanQuestion` and `AutocompleteQuestion`, so we'll need to add one more intermediate class.
* You can edit the text content of `choices`
* Is TextField is more difficult to translate than CharField?
* The parameters for built-in questions are kept in the code, so it can only be translated by native django translation mechanism. To make them consistent with the custom questions, we'd have to move built-in question definitions to the database.
* Built-in questions can be deleted by the user, but they will be restored every time the app starts (bootstrap command)

albertas-jn:
The issue I see with this approach that it might be more difficult for the user to find a question by the key in the admin. Questions will be spread among different types, whereas now they are all under `Questions`. The build-in questions will remain under `Questions`, until we redesign them.

BeritJanssen:
I think discussion IRL would help a great deal. Good point about perhaps making it more difficult to find back questions in the admin interface. At the same time, I would also see a benefit of having the questions grouped into different categories, instead of having them all lumped into one very long list. Searching / sorting by "view" etc. would definitely be needed, also in the implementation you suggest.

albertas-jn:
@BeritJanssen questions are already grouped with `QuestionGroups`, and you can assign newly created questions to new groups.
Shall we just ask the opinion of @drikusroor and @Evert-R here?

BeritJanssen:
For the `Question` model, if there's no strong preference to subclass question types in the team, we can probably stick with Alberta's suggestion to keep it all in one model. The reason I was arguing for splitting out the choices in a model was that I was thinking to go for the [django-modeltranslation](https://github.com/deschler/django-modeltranslation) plugin here, which would already give us a neat admin form. Why I think that it makes sense to use this approach (of which I was hesistant for the Experiment & Block texts) here, is that I expect questions to be reused & localized much more heavily. And that it would make it easier for this particular model to say that we want to add, say, Spanish as an extra language, and to have the fields already created for us, so we don't need to fiddle with manually creating objects for all questions that I want to translate. From Drikus' experiment with django-modeltranslations (https://github.com/Amsterdam-Music-Lab/MUSCLE/pull/1178) I remember that tab view for adding different languages. I had this in mind for the (inline) choices of a question.

albertas-jn:
Added `Choice` model and `django-modeltranslation` for `Question`.

Thanks @drikusroor for exploring `django-modeltranslation`  in #1178

Closes #1138

BeritJanssen (commented):
Nice! Perhaps we can still clean up the migrations, by [squashing them](https://docs.djangoproject.com/en/5.1/topics/migrations/#migration-squashing).

albertas-jn:
Rebased on `develop`. Because of the recent changes in Experiment/Phase/Block admin on `develop`, can't add default `QuestionSeries` to `Block` from `Experiment` admin anymore, so this functionality will have to be integrated separately. At the moment `QuestionSeries` object has to be created manually and linked to a `Block`.

BeritJanssen (approved):

PR #1241, opened 2024-08-27
Refactor block retrieval

### LINKED ISSUE #973: Remove `first_round` from rules files

BeritJanssen:
This PR implements the last steps to close #973.


drikusroor (commented):
I made some minor comments, but I found it hard to review this PR as I don't really understand what its intention is

BeritJanssen:
To clarify: the intention is that the `get` request to `block` doesn't return the `next_round` array anymore, but only information on theme etc., and most importantly, the session id. That id is then used to make a `session/{id}/next_round` request. I've always found the construct of having the request to `block` already contain actions defined in the rules files confusing and awkward to work with. That also does away with the necessity of having a distinction between `Block` / `ExtendedBlock`, as far as I could see.

BeritJanssen:
Ah yes, but now I read back the changes I realize I also mixed in commits to remove the `Consent` option from the potential block actions.

PR #1243, opened 2024-08-27
Refactor: Properly type Block child components

drikusroor:


frontend/src/components/Trial/Trial.tsx
```diff
@@ -17,7 +17,7 @@ export interface IFeedbackForm {
     is_skippable: boolean;
 }
 
-interface TrialProps {
+export interface TrialProps {
     playback: any;
```
BeritJanssen, line 21: We should still type the playback arguments as well (can be left for later).

BeritJanssen (approved):

frontend/src/components/Trial/Trial.tsx
drikusroor: Good spot. It might be possible to use the `PlaybackArgs` type that we already have. I will investigate it.

PR #1250, opened 2024-09-03
Fix `speech2song`, `musical_preferences` and `huang_2022` after refactor

BeritJanssen:
I noticed some issues with the above-mentioned experiments, which this PR fixes:
- `speech2song` still used the `session.current_round` field; replaced with `session.get_rounds_passed()` method and adjusted conditions to progress from one set of trials (i.e., different languages, environmental sounds) to another
- `musical_preferences` had a `Playlist` action, while it shouldn't have one -> removed; also adjusted the counter so it starts at 1 and not at 0.
- `huang_2022` didn't have a `Playlist` action, while it should have one -> added


backend/experiment/rules/musical_preferences.py
```diff
@@ -21,7 +21,7 @@
 class MusicalPreferences(Base):
     ID = 'MUSICAL_PREFERENCES'
     default_consent_file = 'consent/consent_musical_preferences.html'
-    preference_offset = 20
+    preference_offset = 21
```
drikusroor, line 24: What is preference offset?

backend/experiment/rules/tests/test_speech2song.py
```diff
@@ -70,3 +70,30 @@ def test_next_round_serialization(self):
...
+                self.assertIsInstance(actions[0], Explainer)
+                self.assertEqual(self._number_of_sound_trials(actions), 1)
+            elif i == self.block.rounds - 1:
+                self.assertIsInstance(actions[0], Final)
+            elif i == speech2song.n_trials_per_block * 2 * 3 + 1:
+                self.assertIsInstance(actions[0], Explainer)
+            elif i % 2 == 0:
+                self.assertEqual(self._number_of_sound_trials(actions), speech2song.n_presentations)
+            elif i % 2 == 1:
+                self.assertEqual(self._number_of_sound_trials(actions), 1)
```
drikusroor, line 95: Is this a test of the complete flow? I was thinking of trying to make this happen for more experiments. Simulate the frontend and so on, almost like an E2E test but faster

drikusroor (approved):
LGTM. Perhaps you could add a single comment to explain what preference offset does

albertas-jn (approved):

backend/experiment/rules/tests/test_speech2song.py
BeritJanssen: Yes, I thought to go through all the rounds which should be theoretically there. Indeed, I think we can test a lot of the stuff that can go wrong in this way. Perhaps something to discuss: where do we expect problems to happen? Perhaps the e2e tests could also work with pretty dumb mock rules, which only present each action (with potential flavours) once? And testing the logic of the rules files in this way?

PR #1251, opened 2024-09-04
Added: Make submit row fixed on the bottom of the screen until the user scrolls down completely

### LINKED ISSUE #1226: Add save button on top of form(s) or make it float
As the Experiment / Phase / Block form + TranslatedContent form might get fairly big in the nearby future, the user will have to scroll down a lot in order to be able to save the form, even when it only changes something in the top of the form (like the name or slug).

Add a Save & Cancel button on top of the form.

Make the Save & Cancel button (that are on the bottom of the page right now) float along using `position: fixed;` so that they are always visible.

**Additional context**

drikusroor:
We can also look into integrating this little feature into *all* forms automatically, but I think right now only the experiment form can grow big enough to actually need this feature, right?

Resolves #1226

### DIFF
--- a/backend/experiment/static/experiment_form.css
+++ b/backend/experiment/static/experiment_form.css
@@ -6,3 +6,28 @@
 .inline-group .inline-heading:hover {
     text-decoration: underline;
 }
+
+/* Floating class for the submit row */
+.submit-row.floating {
+    position: fixed;
+    bottom: 0;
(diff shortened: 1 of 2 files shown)


BeritJanssen (approved):
Indeed, I don't think we need to add the `floating` class to other submit buttons just yet.

PR #1254, opened 2024-09-06
Refactor: Refactor makeResult function and form handling in Trial component

### LINKED ISSUE #1245: Improve the handling of missing form elements / questions in `Trial.tsx`
In theory, a feedback form could be undefined or an empty array. In case that happens, we now declare it to be an array with an empty object in it. This seems like something that can be improved. TypeScript also doesn't like it. What is the intended behavior in case the feedback form is null/empty? Should we perhaps declare/keep it to be an empty array, or instead declare the empty object to be like a "stub" question with a timeout on it or something?

https://github.com/Amsterdam-Music-Lab/MUSCLE/blob/6734719bda2447748aa25a5ed3c3bcc65a977822/frontend/src/components/Trial/Trial.tsx#L71-L73

drikusroor:
This pull request includes several refactorings in the Trial component. The `checkBreakRound` function has been refactored to use explicit types for parameters. The handling of the `form` variable has been moved inside the `feedback_form` check, as it is only utilized from within that if statement. The assignment of `form` has been simplified, and it is now initialized as an empty array if undefined. Additionally, there is a fix for assertions in tests.

Resolves #1245


frontend/src/components/Trial/Trial.tsx
```diff
@@ -114,18 +82,56 @@ const Trial = (props: TrialProps) => {
...
+                form.map((formElement) => (formElement.value = "TIMEOUT"));
+            }
+
+            if (feedback_form.is_skippable) {
+                form.map((formElement => (formElement.value = formElement.value || '')))
             }
+
+            const breakRoundOn = config.break_round_on;
+            const shouldBreakRound = breakRoundOn && checkBreakRound(form.map((formElement) => formElement.value), breakRoundOn);
+            const shouldCallOnNextInOnResult = !shouldBreakRound
```
BeritJanssen, line 100: Perhaps we can even lose this variable, as the docstring below actually already explains why a boolean is passed into the `onResult` function.

frontend/src/components/Trial/Trial.tsx
```diff
@@ -63,47 +63,15 @@ const Trial = (props: TrialProps) => {
...
-                if (shouldBreakRound) {
-                    onNext(true);
-                }
 
-            } else {
+            submitted.current = true;
 
+            if (!feedback_form) {
 
                 if (result_id) {
```
BeritJanssen, line 76: This bit is in here for "historical reasons" - the `Trial` has a `result_id` at top level, but this is not necessary anymore. (Before a refactor, the "breaking round" mechanism was handled through this top level result id.) I checked, and indeed, no `Trial` in the current rules files contains a `result_id`. We should remove `result_id` from the backend `Trial` constructor and this condition here. (Can be left for later though.)

BeritJanssen (approved):

frontend/src/components/Trial/Trial.tsx
drikusroor: I've created an issue for this #1257 Can you expand upon it if necessary?

frontend/src/components/Trial/Trial.tsx
drikusroor: Can't we change this to:

```tsx
const shouldBreakRound = breakRoundConditions && checkBreakRound(form.map((formElement) => formElement.value), breakRoundConditions);
            const shouldCallOnNextInOnResult = !shouldBreakRound

            await onResult(
                {
                    decision_time: getAndStoreDecisionTime(),
                    form,
                    config
                },
                false,
                false // We already call `onNext` _after_ `onResult` as we cannot pass the `shouldBreak` param `onResult`'s `onNext` call.
            );

            onNext(shouldBreakRound);
```

BeritJanssen (approved):

PR #1256, opened 2024-09-09
Fixed(migration): Refactor block content migration script

drikusroor:
This pull request includes a fix for the migration script that was causing some errors as it tried to get the rules of one or more blocks whose configured rules id was invalid on the test environment.

### DIFF
--- a/backend/experiment/migrations/0054_migrate_block_content.py
+++ b/backend/experiment/migrations/0054_migrate_block_content.py
@@ -35,9 +35,9 @@ def migrate_block_content(apps, schema_editor):
             description=block.description,
         )
 
-        # Attempt to add consent file
-        rules = block.get_rules()
         try:
+            # Attempt to add consent file
+            rules = block.get_rules()
             consent_path = Path("experiment", "templates", rules.default_consent_file)
             with consent_path.open(mode="rb") as f:
                 content.consent = File(f, name=consent_path.name)


BeritJanssen (commented):
See my Teams comment: perhaps better to `continue` in the `except` clause, and log which blocks got caught, so we can identify if specific blocks are the culprit here. (At least I wouldn't be confident that messing around with `block.phase` further on would actually work if anything with the block object itself is off.)

backend/experiment/migrations/0054_migrate_block_content.py
```diff
@@ -5,6 +5,17 @@
 from pathlib import Path
 
 
+def get_rules(block):
```
BeritJanssen, line 8: Not too sure if this is necessary. I'm still thinking that faulty database values are at the root of this.

drikusroor: https://uva-aml.sentry.io/share/issue/3d054044a62b442a89fe5e7d8f9773f7/ The error is literally `'Block' object has no attribute 'get_rules'` though

drikusroor:
> See my Teams comment: perhaps better to `continue` in the `except` clause, and log which blocks got caught, so we can identify if specific blocks are the culprit here. (At least I wouldn't be confident that messing around with `block.phase` further on would actually work if anything with the block object itself is off.)

I could also add the phase earlier in the process and do the rules/consent part afterwards?

PR #1260, opened 2024-09-10
Feature er/mkdocs documentation

### LINKED ISSUE #960: Add API documentation
If somebody would like to get more information about our backend code, there is no easy way to inspect functions and methods, see the arguments they expect and the information they return.

We could make use of [Sphinx](https://www.sphinx-doc.org/en/master/) or [mkdocs](https://www.mkdocs.org/). I personally have only worked with mkdocs, but I think they are comparable in what they do.

No matter whether we decide to create and host API documentation, all our Python code will benefit from
- clear docstrings which indicate which parameters a function expects and returns
- typing

Evert-R:
This PR: 
- sets up MkDOCS to automatically generate documentation for the python code
- contains (and will replace) the wiki documentation
-  resolves #960 

Todo:
- Update the docstrings to conform to google style
- Update the documentation that was copied from the wiki
- Move the screenshots from the wiki into `docs/assets/images` 
  (most likely we will make new screenshots)

The source code from which the docs are rendered is located in:
`/docs`

## Automatic documentation generated with mkdocs

[Mkdocs](https://www.mkdocs.org/) is used in combination with [mkdocstrings](https://mkdocstrings.github.io/) to automatically generate this documentation for the python classes and functions.

The mkdocs configuration file can be found here: `backend/mkdocs.yml`  
The documentation files used to build the site can be found here: `backend/docs/.`

## Build the documentation

To build the documentation in a custom destination folder:
`mkdocs build --site-dir <destination-folder>`
The default folder is `backend/site`

## serve the documentation with mkdocs

`mkdocs serve --dev-addr localhost:8080`

For this to work you either have to install mkdocs and its requirements locally or connect the  workspace in your IDE to the `server` docker/podman container.

## Add a new python file to the documentation:

- Create a new file `<django-app-name>_<python-file-name>.md` in `backend/docs/.`
- Edit this new file and add:
    - `## <django-app-name>.<python-file-name>` Subtitle for this section
    - `::: <django-app-name>.<python-file-name>` The python file that will be read
    - More than one python file can be added to a single mkdocs `.md` file if needed 
- Add this file to to the mkdocs config file `backend/mkdocs.yml`
    - `- <title>: '<django-app-name>_<python-file-name>.md'`

## Guideline for writing docstrings

```
    """Description of the class or function    

    Args:
        Param (type): description

    Attributes:
        attr (type): description

    Returns:
        (type): description

    Raises:
        Error-code: description

    Note:
        Optional note

    Example:
        Examples should be written in doctest format, and should illustrate how
        to use the function.
    """
```

[Click here to read more about formatting docstrings](https://realpython.com/python-project-documentation-with-mkdocs/#step-3-write-and-format-your-docstrings)


backend/experiment/actions/base_action.py
```diff
@@ -2,14 +2,24 @@
...
     ID = 'BASE'
     style = None
 
     def __init__(self, style: FrontendStyle = None):
         self.style = style
         pass
 
-    def action(self):
+    def action(self) -> dict:
+        """The action that can be send to the frontend
```
BeritJanssen, line 17:
```suggestion
        """The action that can be sent to the frontend
```

backend/experiment/actions/consent.py
```diff
@@ -18,16 +18,16 @@ def get_render_format(url: str) -> str:
...
 
 def render_html_or_markdown(dry_text: str, render_format: str) -> str:
-    """
-    render html or markdown
+    """render html or markdown
 
-    Parameters:
-        dry_text: contents of a markdown or html file
-        render_format: type of contents, either 'HTML' or 'MARKDOWN'
+    Args:
```
BeritJanssen, line 23: Should this be called "Arguments" instead of "Args"?

backend/experiment/actions/consent.py
```diff
@@ -18,16 +18,16 @@ def get_render_format(url: str) -> str:
...
+        dry_text (str): contents of a markdown or html file
```
BeritJanssen, line 24: I thought it wasn't necessary to add types here if it's already in the function description.

backend/experiment/actions/consent.py
```diff
@@ -64,7 +66,7 @@ class Consent(BaseAction):  # pylint: disable=too-few-public-methods
                 amet, nec te atqui scribentur. Diam molestie posidonium te sit, \
                 ea sea expetenda suscipiantur contentiones."
 
-    def __init__(self, text: File, title="Informed consent", confirm="I agree", deny="Stop", url=""):
+    def __init__(self, text: File, title: str="Informed consent", confirm: str="I agree", deny: str="Stop", url: str=""):
```
BeritJanssen, line 69: Might also explicitly state the return type:
```suggestion
    def __init__(self, text: File, title: str="Informed consent", confirm: str="I agree", deny: str="Stop", url: str="") -> dict:
```

backend/experiment/models.py
```diff
@@ -9,16 +9,27 @@
...
 language_choices[0] = ("", "Unset")
 
 
-def consent_upload_path(instance, filename):
-    """Generate path to save consent file based on experiment.slug and language"""
+def consent_upload_path(instance: models.Model, filename: str) -> str:
+    """Generate path to save consent file based on experiment.slug and language
+    
+    Args:
+        instance (Experiment): Experiment instance to determine folder name
```
BeritJanssen, line 24: Is the type `Experiment` or  `models.Model`? As stated above, remove the type definitions here and use the most specific type in the function definition.

backend/experiment/models.py
```diff
@@ -27,7 +38,17 @@ def consent_upload_path(instance, filename):
...
-    """A model to allow nesting multiple phases with blocks into a 'parent' experiment"""
+    """A model to allow nesting multiple phases with blocks into a 'parent' experiment
+    
+    Attributes:
+        slug (str): Slug
+        translated_content (list[ExperimentTranslatedContent]): Translated content
+        theme_config (theme.ThemeConfig): ThemeConfig instance
+        dashboard (bool): Show dashboard?
+        active (bool): Set experiment active
+        social_media_config (SocialMediaConfig): SocialMediaConfig instance
```
BeritJanssen, line 49: Here the types *may* be necessary, but perhaps also worth checking if they aren't derived from Django's field class.

backend/experiment/models.py
```diff
@@ -45,30 +66,60 @@ def __str__(self):
...
+        Returns:
+            (list[Block]): Associated blocks
+        """
+
         phases = self.phases.all()
         return [block for phase in phases for block in list(phase.blocks.all())]
 
-    def export_sessions(self):
-        """export sessions for this experiment"""
+    def export_sessions(self) -> list[models.Model]:
```
BeritJanssen, line 79: shouldn't this be a `list[Session]`?

backend/experiment/models.py
```diff
@@ -45,30 +66,60 @@ def __str__(self):
...
+        """
+
         all_sessions = Session.objects.none()
         for block in self.associated_blocks():
             all_sessions |= Session.objects.filter(block=block).order_by("-started_at")
         return all_sessions
 
-    def current_participants(self):
-        """Get distinct list of participants"""
+    def current_participants(self) -> list[models.Model]:
```
BeritJanssen, line 91: And this `list[Participant]`?

backend/experiment/models.py
```diff
@@ -45,30 +66,60 @@ def __str__(self):
...
-        """Get content for a specific language"""
+    def get_translated_content(self, language: str, fallback: bool = True) -> list[models.Model]:
+        """Get content for a specific language
+        
+        Args:
+            language (str): Language code
+            fallback (bool): Return fallback language if language isn't available
+
+        Returns:
+            (list[ExperimentTranslatedContent]): Translated content
```
BeritJanssen, line 120: I believe here we only return *one* `TranslatedContent` object. Or the function names are misleading.

backend/experiment/serializers.py
```diff
@@ -86,31 +124,66 @@ def serialize_block(block_object: Block, participant: Participant):
 
 def get_upcoming_block(block_list, participant, repeat_allowed=True):
```
BeritJanssen, line 125: add typing in the function definition rather than the docstring.

backend/experiment/utils.py
```diff
@@ -38,23 +46,58 @@ def slugify(text):
     return text.lower()
 
 
-def non_breaking_spaces(s):
-    # Convert regular spaces to non breaking spacing
-    # on the given string
+def non_breaking_spaces(s) -> str:
```
BeritJanssen, line 49: perhaps we can give this argument a more descriptive name, e.g., `input_string`.

BeritJanssen (approved):
Noice! I did notice that the docstrings add types to the `Args`. As far as I know, typing in the `def` is sufficient, so it's not necessary to add it to the docstrings. Can you check that, and if that is indeed so, remove the typing from the docstrings, and in case there's a discrepancy between the docstring and type hints, check which one should be used? Other than that, awesome work!

Evert-R:
> Noice! I did notice that the docstrings add types to the `Args`. As far as I know, typing in the `def` is sufficient, so it's not necessary to add it to the docstrings. Can you check that, and if that is indeed so, remove the typing from the docstrings, and in case there's a discrepancy between the docstring and type hints, check which one should be used? Other than that, awesome work!

The types are taken from the docstring only. I haven't found a way to get this from the function definition, unfortunately.

backend/experiment/models.py
Evert-R: Good catch. `Experiment` wasn't recognized before, so I used `models.Model`. It does work with an import.

backend/experiment/models.py
Evert-R: Without this in the docstring the docs only contain the name of the class (and the complete source code behind the dropdown)

Evert-R:
> Noice! I did notice that the docstrings add types to the `Args`. As far as I know, typing in the `def` is sufficient, so it's not necessary to add it to the docstrings. Can you check that, and if that is indeed so, remove the typing from the docstrings, and in case there's a discrepancy between the docstring and type hints, check which one should be used? Other than that, awesome work!

Earlier in the proces I couldn't get mkdocs to take the types from the function def, only from the docstring. but I just tried this again and it does work now.  Don't know what I missed. I'll go over them again, cause this makes a lot more sense.

backend/experiment/models.py
BeritJanssen: That's not what the type says.

Issue #1257, opened 2024-09-09
Clean up (usage of) `onResult`, `makeResult`, and `onNext`


drikusroor:
## 1. Remove `onNext` calls from the `onResult` callback function

We can probably just wait for `onResult` to finish and call `onNext` manually (in `makeResult`)

## 2. Remove `result_id` from the backend `Trial` constructor and the `if (result_id)` condition in `Trial.tsx`.

https://github.com/Amsterdam-Music-Lab/MUSCLE/blob/00c3e63c61245db7b807859feb874d77196df58f/frontend/src/components/Trial/Trial.tsx#L76-L83

> This bit is in here for "historical reasons" - the `Trial` has a `result_id` at top level, but this is not necessary anymore. (Before a refactor, the "breaking round" mechanism was handled through this top level result id.) I checked, and indeed, no `Trial` in the current rules files contains a `result_id`. We should remove `result_id` from the backend `Trial` constructor and this condition here. (Can be left for later though.)

_Originally posted by @BeritJanssen in https://github.com/Amsterdam-Music-Lab/MUSCLE/pull/1254#discussion_r1749781336_

## UPDATE

Some work in progress can be found in #1258

PR #1262, opened 2024-09-10
Refactored: Clean up `onResult`, `makeResult`, and `onNext`

drikusroor:
Related to #1257


frontend/src/stories/ButtonArray.stories.jsx
```diff
@@ -58,7 +58,6 @@ export const CategorizationWithHiddenText = {
             view: "BUTTON_ARRAY",
             explainer: "",
             question: "",
-            result_id: 16549,
```
BeritJanssen, line 61: Was the `result_id` removed because it's not necessary? I thought it could still be passed in a `Question` type?

frontend/src/hooks/useResultHandler.ts
```diff
@@ -74,24 +48,10 @@ const useResultHandler = ({ session, participant, onNext, state }: UseResultHand
                 result: mergedResults,
```
BeritJanssen, line 48: Why not simply `result`? We don't need to assign it to the `mergedResults` variable anymore, or do we?

BeritJanssen (approved):
Nice cleanup!

frontend/src/hooks/useResultHandler.ts
drikusroor: Yes, good catch

frontend/src/stories/ButtonArray.stories.jsx
drikusroor: I also removed it from the Trial props, was that unnecessary?

BeritJanssen: It's not necessary in the `Trial` props, but will be necessary in the `Question` props, and this is a `Question`, right? For storybook, it won't matter, but as far as I'm aware it's still a valid prop here.

drikusroor: I've re-added it, also as an optional `Question` interface property

PR #1263, opened 2024-09-10
Remove unnecessary fields from admin

### LINKED ISSUE #1133: Delete unused Block fields
After #973
- [ ] block.title no longer required but optional (is not necessary per se, except for when you use dashboard)
- [ ] remove block.consent field (is on experiment level now)
- [ ] remove block.active field (we can use the `active` field on an experiment level to deactivate the whole experiment)
~- remove block.slug field~

### LINKED ISSUE #1135: Redesign admin interface after refactor
After #1133.

The admin interface, after the refactor, should provide users with the following utilities:
~~- [ ] when creating an experiment, auto-create phase & block within it~~
- [x] #1192
  - If possible, we can use nested inline forms. See also #1188 for how you can use `django-nested-admin` to do that.

For the latter, much of the logic of the current `ExperimentCollection` admin can be reused

Update: after the phase / block are nested inlines, we can probably achieve a cleanish admin interface if we do the following:
In principle, with the inline forms for phase & block, everything's already in pretty good working shape.
Some things to change:
- [ ] Remove "Name" from the Phase. I think Phases should be pretty easy to see at a glance with the blocks which are inside it.
- [ ] Turn the inline block's `playlists' field into a checkbox selection
- [x] Turn the inline block's `rules` field into a dropdown
- [ ] Remove `Phase` and `Block` forms altogether? I think in the long run, we want to manage this all inline in the `Experiment` form, so to avoid clutter and confusion, I would argue we can unregister `Phase` and `Block` models from the admin interface.

BeritJanssen:
Close #1135 , close #1133.

This PR removes several fields from the experiment app's models. It also unregisters the `Block(TranslatedContent)`, and `Phase` admin forms, so they're only editable inline, as well as removes some columns from the experiment overview which either aren't available anymore (i.e., the shortcut to edit a phase), or are often empty (i.e., the `description_excerpt` field).

Screenshot of Experiment list view after change:
Screenshot 2024-09-10 at 17 29 14

Experiment form:
Screenshot 2024-09-10 at 17 32 09


backend/experiment/actions/tests/test_wrappers.py
```diff
@@ -15,7 +15,7 @@ def setUp(self):
         self.participant = Participant.objects.create()
         self.section = Section.objects.create(
             filename='some/audio/file.mp3', playlist=self.playlist)
-        self.block = Block.objects.create(name='TestBlock')
+        self.block = Block.objects.create(slug="TestBlock")
```
drikusroor, line 18:
```suggestion
        self.block = Block.objects.create(slug="test-block")
```

drikusroor (approved):

PR #1280, opened 2024-09-24
Fix button array spacing

BeritJanssen:
This branch sets different column gaps between buttons in an array, based on the number of buttons, and the viewport width.

Screenshot 2024-09-24 at 14 18 29
Screenshot 2024-09-24 at 14 18 50
Screenshot 2024-09-24 at 14 19 22

### DIFF
--- a/frontend/src/components/Question/_ButtonArray.scss
+++ b/frontend/src/components/Question/_ButtonArray.scss
@@ -12,7 +12,18 @@
 
     .btn-group-toggle-custom {
         display: flex;
-        column-gap: 2.5vw;
+        &:has(> :last-child:nth-child(2)) {
+            column-gap: 8vw;
+        }
+        &:has(> :last-child:nth-child(3)) {
+            column-gap: 6vw;
+        }
+        &:has(> :last-child:nth-child(4)) {
+            column-gap: 3vw;
+        }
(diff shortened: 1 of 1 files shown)


frontend/src/components/Question/_ButtonArray.scss
```diff
@@ -12,7 +12,18 @@
...
+        }
+        &:has(> :last-child:nth-child(3)) {
+            column-gap: 6vw;
+        }
+        &:has(> :last-child:nth-child(4)) {
+            column-gap: 3vw;
+        }
+        &:has(> :last-child:nth-child(5)) {
+            column-gap: 2vw;
+        }
```
drikusroor, line 26: I don't know if you had considered it, but it's also possible to just do `justify-content: space-between` for automatic spacing between buttons. https://www.w3schools.com/cssref/playdemo.php?filename=playcss_justify-content

drikusroor (approved):
LGTM, but see the comment I made about `space-between` :-)

frontend/src/components/Question/_ButtonArray.scss
BeritJanssen: I think I experimented with that before, but then ended up with spacing that was sup-optimal, hence switched to column-gap. But I'll give it another try, thanks!

PR #1284, opened 2024-09-24
Added: Disable pinch zoom on iOS

### LINKED ISSUE #1274: Control behaviour when zooming in
When zooming in accidentally, the view may be so distorted to such an extent that it's difficult to navigate through the application. Perhaps we should enforce sizes with `@media / min-width / max-width` css.

Or should we hard-disable zooming, as suggested [here](https://stackoverflow.com/questions/4389932/how-do-you-disable-viewport-zooming-on-mobile-safari)? That means the app is going to be less accessible.

drikusroor:
This pull request adds a new hook called useDisableIOSPinchZoomOnTouchDevices to the frontend codebase. The hook is responsible for preventing pinch zoom gestures on iOS devices. It is used in the App component to disable pinch zoom on touch devices. Additionally, the pull request includes necessary file changes to implement the hook and add the touch-action property to disable pinch zoom on iOS.

Additionally, the css-property `touch-action: manipulation;` is added to the body root class in order to prevent touch actions.

Resolves #1274

### DIFF
--- a/frontend/src/components/App/App.tsx
+++ b/frontend/src/components/App/App.tsx
@@ -18,6 +18,7 @@ import Profile from "../Profile/Profile";
 import Reload from "../Reload/Reload";
 import StoreProfile from "../StoreProfile/StoreProfile";
 import useDisableRightClickOnTouchDevices from "../../hooks/useDisableRightClickOnTouchDevices";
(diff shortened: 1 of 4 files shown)


BeritJanssen (approved):
LGTM. Unfortunately, cannot test it.

drikusroor:
> LGTM. Unfortunately, cannot test it.

Same, I will have to test this on my phone on staging 😅

PR #1288, opened 2024-09-26
Added: Refactor social media message handling and improve social media info method

### LINKED ISSUE #1270: Make the social media message translateable by moving the content property to `ExperimentTranslatedContent`
See also: https://github.com/Amsterdam-Music-Lab/MUSCLE/issues/1204#issuecomment-2270777135

I have already started this ticket, but I have some things that need to be discussed:
- Should a user be allowed to use _points_, _experiment_name_ and potential other placeholders in the message?
- Should the social media message always be passed in the `get_experiment` endpoint and/or should we ideally be able to use it in some `get_next_round` endpoint with a `Final` action too?
  - How do we deal with the placeholder values then and their potential non-existence (like no points available)?

## Proposed requirements

- Modify individual social media configurations in 4 rules files to get its message from the experiment level
- In the rules base class, get the experiment level's social media message 
- Create a simple but elegant way for handling placeholders, including handling singular and plural words like point & points

drikusroor:
Resolves #1270


backend/experiment/rules/musical_preferences.py
```diff
@@ -230,38 +236,9 @@ def calculate_score(self, result, data):
...
-            },
-            "url": url,
-            "hashtags": [*tags, "amsterdammusiclab", "citizenscience"],
-        }
-
     def get_final_view(self, session, top_participant, known_songs, n_songs, top_all):
         # finalize block
-        social_info = self.social_media_info(session.block, top_participant, known_songs, n_songs, top_all)
-        social_info["apps"] = ["weibo", "share"]
+        social_info = self.social_media_info(session)
```
BeritJanssen, line 241: The channels of these rules files are different from base (i.e., "weibo", "share"), those should still be adjusted.

BeritJanssen: Ah, nevermind. Realising now it's set through the database.

BeritJanssen (approved):
Nice, but this change really makes it especially weird that the `social_media_info` comes from the rules file. But fixing that can be left for later. Other than the different default/help text in the migration, this looks good to me.

As a general remark: reviewing is now a bit of a needle-haystack situation with finding the relevant changes. Perhaps configure your code formatter to only operate on the lines that were actually modified by you?

backend/experiment/migrations/0058_remove_socialmediaconfig_content_and_more.py
```diff
@@ -0,0 +1,59 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="experimenttranslatedcontent",
+            name="social_media_message",
+            field=models.TextField(
+                blank=True,
+                default="I scored {points} points in {block_name}!",
+                help_text="Content for social media sharing. Use {points} and {block_name} as placeholders.",
```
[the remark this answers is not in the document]
drikusroor: I had fixed the text in models, but forgot to update the migration. Good spot :)

drikusroor:
> Nice, but this change really makes it especially weird that the `social_media_info` comes from the rules file. But fixing that can be left for later. Other than the different default/help text in the migration, this looks good to me.
> 
> As a general remark: reviewing is now a bit of a needle-haystack situation with finding the relevant changes. Perhaps configure your code formatter to only operate on the lines that were actually modified by you?

So yeah... that rules base class method is just there to give the four experiments the correct social media information from the experiment level whenever they want to use it in a `Final` component. I thought that would be easier than calling the model's social media method in each individual rules file.

PR #1295, opened 2024-10-01
ToontjeHoger Kids hotfix

BeritJanssen:
This branch fixes the following issues:
- on Safari, some images would be squashed on narrow viewports. This was caused by the height setting, I removed it. I'm not sure what the intention was behind it, but after the fix I don't see any problems on Safari or Chrome with different viewports.
- the name & composer of sections used in the Toontjehoger Mozart Effect are very long (they include not only the name of the composer, but also of the ensemble and conductor), so I removed the formatting with non-breaking-spaces to address this problem:
screenshot 3
Again, I couldn't see obvious drawbacks of allowing the interface to break between words in the section's name / artist description.


drikusroor (approved):
LGTM 👍

PR #1297, opened 2024-10-01
Added: Make the `Histogram` component display the actual audio instead of showing random bar heights

drikusroor:
As described in the title: This PR makes the Histogram component display real-time audio frequency data instead of random bar heights. There's probably more to improve, like playing around with the amount of bars. For example, the [fftSize](https://developer.mozilla.org/en-US/docs/Web/API/AnalyserNode/fftSize) on the analyzer is set to 32. This is mandatory as the value should be "(...) a power of two in the range 32 to 32768".

I don't know if setting that size to 32 means that it has data for showing 32 bars. I set the default amount of bars to 10 now as that (subjectively) seemed to represent the audio pretty well, but it may mean we are missing a huge chunk of the frequency spectrum. Maybe it's also worth investigating taking the average of every 4 bars and then show 8 bars in the end, or taking every 4th bar in the 32 bars, I don't know, I'm open to ideas.

https://github.com/user-attachments/assets/f67e8d0f-13a0-4334-9c09-d8334bcce8ac


BeritJanssen:
Very nice, Drikus! I do wonder about @jaburgoyne's opinion though. What do you think? Perhaps also good to add a screen capture with music with a beat?

jaburgoyne:
Looks great! I also cannot hear the music, though, and I wonder a little bit about the bars. As for 8 vs. 10, that's really an aesthetic choice; either is fine with me. I worry more about the frequency bands chosen: is it _always_ a descending slope with more low frequencies and fewer high frequencies, or are there musical examples in our experiments with other shapes? If it's always or almost always a descending slope, then we may want to cut out the bars corresponding to the highest frequencies

drikusroor:
> Looks great! I also cannot hear the music, though, and I wonder a little bit about the bars. As for 8 vs. 10, that's really an aesthetic choice; either is fine with me. I worry more about the frequency bands chosen: is it _always_ a descending slope with more low frequencies and fewer high frequencies, or are there musical examples in our experiments with other shapes? If it's always or almost always a descending slope, then we may want to cut out the bars corresponding to the highest frequencies

I've altered the histogram strategy by now selecting 11 frequency bands and removing the first three. That seems to give better results already. What do y'all think? Do you want a new example video or should I merge this and tweak it whenever we aren't satisfied with it?

BeritJanssen:
If you think it looks good, let's merge and see indeed! Thanks.

PR #1300, opened 2024-10-01
Document `section.models`

### LINKED ISSUE #846: Clean up and document `section.models`
The current `Section` and `Playlist` models include many methods. Some might not be used anywhere, or might have a more logical place in `section.admin`. All methods should be documented with a proper docstring. The relationship between `Section`, `Playlist`, and eventually, `Session`, also should be documented.

BeritJanssen:
Close #846. This PR includes updates of wiki pages 1-3, and also reorganizes and documents `section.models`. It also adds unit tests for some untested functions in the `section` app. All functions starting with underscore are now excluded from the mkdocs generation, and all functions which are *not* called by the rules files are now preceded by underscores.


backend/aml/urls.py
```diff
@@ -49,8 +49,8 @@
     urlpatterns = [path('server/', include(urlpatterns))]
 
 # Debug toolbar
-if settings.DEBUG:
+if settings.DEBUG and not settings.TESTING:
```
BeritJanssen, line 52: NB: These changes were necessary as the debug toolbar was interfering with tests. So now, there's a distinction between debugging and testing. Side note: perhaps we might also just let the debug toolbar go, as it's not being really used by anyone?

backend/section/models.py
```diff
@@ -232,87 +245,153 @@ def update_admin_csv(self):
...
 
 class Section(models.Model):
-    """A snippet/section of a song, belonging to a Playlist"""
-
-    def random_code():
+    """A snippet/section of a song, belonging to a Playlist
+
+    Attributes:
+        playlist (models.ForeignKey): a Many-To-One relationship to a Playlist object
+        song (models.ForeignKey): a Many-To-One relationship to a Playlist object (can be null)
```
Evert-R, line 311: Same. Using `Song` will link to the model.

backend/aml/urls.py
Evert-R: I never use it

Evert-R (approved):
Nice. Only one thing:
Using `Playlist` and `Song` as types in the `Section` model will link these types to the models in the documentation.

PR #1306, opened 2024-10-02
Style: Improve styling & stories of profile view, rank & cup

drikusroor:
I noticed the code of the `Profile` page component and the `ProfileView` could be improved upon. I've also expanded on its story, adding more possible scenarios. When I was working on that, I also noticed the styling was a bit off. I subsequently found out some things that could also be improved in `Rank` / `Cup`, which I did. See the screenshots below for comparison.

Additionally, I updated the routes a bit, removing the `exact` prop, which [doesn't exist anymore](https://stackoverflow.com/a/69866593/4496102) in `react-router` v6, and by removing the non-existent route for "session".

# Screenshots

### Before

image

image

### After

image

image

### DIFF
--- a/frontend/src/components/App/App.tsx
+++ b/frontend/src/components/App/App.tsx
@@ -73,7 +73,7 @@ const App = () => {
                         />
 
                         {/* Profile */}
-                        <Route path={URLS.profile} exact element={<Profile />} />
+                        <Route path={URLS.profile} element={<Profile />} />
 
                         {/* Internal redirect */}
                         <Route path={URLS.internalRedirect} element={<InternalRedirect />} />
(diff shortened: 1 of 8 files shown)


frontend/src/components/Cup/Cup.scss
```diff
@@ -7,7 +7,6 @@
     box-shadow: 13px 0 64px $yellow;
     border: 5px solid white;
     box-sizing: border-box;
-    margin-bottom: 20px;
```
drikusroor, line 10: Spacing between elements is best done in parent / container components, and not in the element itself. It should be spacing-agnostic, so to speak :-)

frontend/src/components/App/App.tsx
```diff
@@ -82,13 +82,9 @@ const App = () => {
                         {/* Experiment */}
                         <Route path={URLS.experiment} element={<Experiment />} />
 
-                        {/* Session - ⚠️ What is the purpose of this non-functional route? */}
-                        <Route path={URLS.session} />
```
drikusroor, line 86: I removed this route because `URLS.session` does not exist in the `URLS` object. (It does exist in the `URLS` object in `API.ts`, but I assume that's something different.)

BeritJanssen (approved):

Issue #1136, opened 2024-06-18
Allow admin users to duplicate `Experiment` objects from admin interface


BeritJanssen:
After #1135, the form to create an experiment will be even longer than prior to the refactor. To simplify creating an experiment which is virtually identical, apart from a questionnaire of playlist, it would be a nice feature to add an action from the dropdown menu in the admin interface, which creates a copy of an existing experiment, in which the slug is automatically modified to, e.g. `{original-slug}-copy`. 

## Requirements

- Add copy/duplicate button to `Experiment` section in admin interface
- Clicking the duplicate button will copy the `Experiment` and give it a:
  - Name: `{original-name} (copy)`
  - Slug: `{original-slug}-copy`
- Additionally, it will copy the related:
  - `Phase`s
  - `Block`s related to the `Phase`s
  - `ThemeConfig` related to `Experiment` & `Block`, along with its `HeaderConfig` & `FooterConfig`
  - `Feedback` related to the `Block`s?
  - `QuestionSeries` related to the `Block`s?
  - `SocialMediaConfig`

drikusroor:
What do you mean by "virtual identical, apart from a questionnaire of/or playlist"? Does that mean the duplicate Experiment will have no playlist? Or will the playlist be duplicated as well? Perhaps it would be good to have a list of properties / fields that will or will not be duplicated along with the Experiment.

BeritJanssen:
These are choices I have not thought about enough. In general, I think the only reason to duplicate an experiment *without* modifying the questions or playlists would be to add a different Consent form (e.g., you relaunch the experiment but with different contact persons), so perhaps we could even choose to copy everything but the Consent form.

PR #1330, opened 2024-10-29
Feature er/duplicate experiment

Evert-R:
### **(Updated: See bold text)**

This PR adds a `duplicate` inline action to the experiments on the admin overview.

Clicking on duplicate will first show an input to enter an extention for the slugs. 
(so entering `copy` will add `-copy` to all the slugs.

~~This input will be stripped from non alphanumeric characters, and added  to all the slugs, seperated by an hyphen.~~
**If the extension contains non alphanumeric or uppercase characters, or will cause duplicate slugs in the database, the django messages framework is used to show the user what when wrong when returning to the form.**

New objects will be created for:
- Experiment
- phases
- blocks
- experiment translated content
- block translated content
- **QuestionSeries**
- **QuestionInSeries**

These objects will be reassigned:
- Playlist
- theme_config
- Image
- **Question**

closes: #1136


backend/experiment/admin.py
```diff
@@ -299,6 +302,80 @@ def slug_link(self, obj):
 
     slug_link.short_description = "Slug"
 
+    def duplicate(self, request, obj, parent_obj=None):
+        """Duplicate an experiment"""
+
+        if "_duplicate" in request.POST:
+            # Get slug from the form
+            slug_extention = request.POST.get("slug-extention")
```
BeritJanssen, line 310:
```suggestion
            slug_extension = request.POST.get("slug-extension")
```

backend/experiment/templates/duplicate-experiment.html
```diff
@@ -0,0 +1,55 @@
...
+<p>The following existing objects will be reused and assigned to the duplicated objects:</p>
+<ul>
+    <li>Playlists</li>
+    <li>Consent files</li>
+    <li>Theme config</li>
+</ul>
+
+<form action="" method="post">
+    {% csrf_token %}{% render_inline_action_fields %}
+    <label for="slug-extention">Enter an extention to add to the slugs in the dupicated experiment:</label>
```
BeritJanssen, line 48:
```suggestion
    <label for="slug-extension">Enter an extension to add to the slugs in the duplicated experiment:</label>
```

BeritJanssen (changes requested):
Nice! I had envisioned the duplication to be formless, i.e., you would just click duplicate and get a copy of the experiment, with all slugs getting a `-copy` attached automatically, and it would then be up to the user to change the slug afterwards. It's also nice to let admin users set the extension right away though, so I'm not opposed to this solution for sure.

What I like about the form is that it explains so well what exactly the duplication is going to do. If we want to not use the form, this could be moved (in shortened form) to a help text shown when hovering over the "duplicate" button, perhaps.

One thing I do think this PR needs to still tackle is to check for duplicate slugs before attempting to save the object: if a there's already an `Experiment` or `Block` object with `slug-copy`, we couldn't save another object with `slug-copy` because of the unique constraint. This would throw an error. So then we might still need to attach integers to the slug (`slug-copy1`), and then also check that the integer that's attached hasn't yet been used... This is a nice one for test-driven development.

Evert-R:
> Nice! I had envisioned the duplication to be formless, i.e., you would just click duplicate and get a copy of the experiment, with all slugs getting a `-copy` attached automatically, and it would then be up to the user to change the slug afterwards. It's also nice to let admin users set the extension right away though, so I'm not opposed to this solution for sure.
> 
> What I like about the form is that it explains so well what exactly the duplication is going to do. If we want to not use the form, this could be moved (in shortened form) to a help text shown when hovering over the "duplicate" button, perhaps.

My reasoning was that it prevents accidental clicks on the `duplicate` button, which would duplicate an entire experiment. (Most functions in the admin have a confirm screen) 
The other reason is that you would almost always want to change the `-copy` extension to something else. With multiple blocks this could be a drag. 

> One thing I do think this PR needs to still tackle is to check for duplicate slugs before attempting to save the object: if a there's already an `Experiment` or `Block` object with `slug-copy`, we couldn't save another object with `slug-copy` because of the unique constraint. This would throw an error. So then we might still need to attach integers to the slug (`slug-copy1`), and then also check that the integer that's attached hasn't yet been used... This is a nice one for test-driven development.

Good point ! I now also noticed that you can have multiple experiments with the same slug, as the `unique` argument is not set on the `Experiment` model, like it's set on the `Block` model.
I will look into this as well.

BeritJanssen:
Good point, I agree that it's better to have the form.

We should add the `unique` constraint on the `Experiment.slug`, good that you checked!

BeritJanssen:
BTW, instead of auto-modifying slugs, we might of course also consider returning a validation error on the form: "This slug extension is already taken. Please choose another one."

I also wonder whether we couldn't also reuse the slug validation from Django and throw a validation error to the user, rather than tacitly adjusting anything that isn't an allowed for a `SlugField`. A bit more annoying to users, but then they know why the cannot use "f***Extension!!" or something similar.

BeritJanssen (approved):

BeritJanssen (approved):

PR #1341, opened 2024-10-31
Bugfix: revise logic on start of categorization experiment

### LINKED ISSUE #1333: 🐛 [BUG] - Categorization experiment does not present questions
### Description

 Categorization experiment does not present questions defined in QuestionSeries. Other experiments (e.g. Eurovision) do.

### Reproduction URL

_No response_

### Reproduction steps

```bash
Run a categorization experiment with default QuestionSeries attached.
```

### Logs

_No response_

### Browsers

_No response_

### OS

_No response_

### Additional context

_No response_

BeritJanssen:
Close #1333. Since I'm not quite sure whether the current change will break anything, can  either of you give this an extra look, @Evert-R and @albertas-jn ? I think you designed the original logic. The problems still arise from getting rid of the `first_round` method. NB: I did one playthrough up until the testing phase, and everything seemed to be working fine.

### DIFF
--- a/backend/experiment/rules/categorization.py
+++ b/backend/experiment/rules/categorization.py
@@ -38,19 +38,13 @@ class Categorization(Base):
     def next_round(self, session: Session):
         json_data = session.json_data
 
-        if json_data.get("started"):
+        if not json_data.get("phase"):
             actions = [self.get_intro_explainer()]
             questions = self.get_open_questions(session)
             if questions:
                 actions.extend(questions)
-            session.save_json_data({"started": True})
-            return actions
-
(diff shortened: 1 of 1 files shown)


drikusroor (commented):
Would this be testable using unit tests? 🤔

BeritJanssen:
Yes, the problem is that I'm not clear on the specifications.

Evert-R (approved):
I went through the whole experiment without problems, so I'd say: merge.
Though I do think the rules file could use a cleanup and some tests. (by me in a new issue, that is)

PR #1347, opened 2024-11-04
Add zh-hans language to modeltranslation

albertas-jn:
I copied `zh -> zh-hans` and  left `zh`, because I am not sure if some other functionality depends on it.


BeritJanssen (approved):
Looks good to me. I do believe we can savely delete `zh`, but perhaps good to test this first with an experiment with Chinese translations.

BeritJanssen:
I suspect this branch is still missing a migration, seeing as there are failing tests for missing columns.

backend/question/migrations/0007_modeltranslation_populate_zh_hans.py
```diff
@@ -0,0 +1,29 @@
+from django.db import migrations
+from question.questions import create_default_questions
+from question import questions
+from django.core import management
+from django.utils import translation
+
+
+def populate_translation_fields(apps, schema_editor):
+
+    QuestionH = apps.get_model("question", "Question")
```
BeritJanssen, line 10: Why call the model definitions `{}H`? I don't get what the `H` stands for, and in other migrations we just refer to this as the model name.

backend/question/questions.py
```diff
@@ -57,10 +57,9 @@ def get_questions_from_series(questionseries_set):
     return [create_question_db(key) for key in keys_all]
 
 
-def create_default_questions(overwrite = False):
+def create_default_questions(Question=Question, Choice=Choice, QuestionGroup=QuestionGroup):
```
BeritJanssen, line 60: It's very unusual to use upper/camelcase argument names. Better to give the arguments a descriptive snake-case name, e.g., "question_model". Could we actually circumvent passing the model along if we rewrite this function as a member of the `Question` model? Then we could get `Choice` and `QuestionGroup` via the foreign key lookups, and wouldn't need to be concerned with the "version" of the model/object we are using.

Issue #952, opened 2024-04-15
Replace string format by `os.path.join` wherever we want to create file paths


BeritJanssen:
This surfaced in #950: the `RELOAD_PARTICIPANT_URL` had a leading slash in the production settings, leading to problems with social media sharing. We should go over all the places where format strings are used to create file paths or urls. For file paths, we can use `os.path.join` instead, which would avoid duplicating slashes, and make sure this also works on non-unix systems. For urls, there doesn't seem to be a standard library solution (`urllib.parse` provides some functionality but only works with full urls as far as I can tell).

PR #1349, opened 2024-11-05
use join for file paths

Evert-R:
This PR replaces formatted strings for creating file paths with `os.path.join`

resolves: #952


backend/experiment/management/commands/createruleset.py
```diff
@@ -27,10 +27,10 @@ def create_block_rule_class(self, ruleset_name):
         ruleset_name_snake_case, ruleset_name_snake_case_upper, ruleset_name_pascal_case = self.get_ruleset_name_cases(ruleset_name)
 
         # Create a new file for the block rules class
-        filename = f"./experiment/rules/{ruleset_name_snake_case}.py"
+        filename = join('.', 'experiment', 'rules', ruleset_name_snake_case + '.py')
```
BeritJanssen, line 30:
```suggestion
        filename = join(".", "experiment", "rules", f"{ruleset_name_snake_case}.py")
```

I think the f syntax is more readable than `+` for string formatting.

backend/experiment/management/commands/createruleset.py
```diff
@@ -81,10 +81,10 @@ def create_test_file(self, ruleset_name):
         ruleset_name_snake_case, ruleset_name_snake_case_upper, ruleset_name_pascal_case = self.get_ruleset_name_cases(ruleset_name)
 
         # Create a new file for the block class
-        filename = f"./experiment/rules/tests/test_{ruleset_name_snake_case}.py"
+        filename = join('.', 'experiment', 'rules', 'tests', 'test_' + ruleset_name_snake_case + '.py')
```
BeritJanssen, line 84: See above.

backend/experiment/models.py
```diff
@@ -145,7 +146,7 @@ def consent_upload_path(instance: Experiment, filename: str) -> str:
     folder_name = experiment.slug
     language = instance.language
 
-    return f"consent/{folder_name}/{language}-{filename}"
+    return join('consent', folder_name, language, filename)
```
BeritJanssen, line 149:
```suggestion
    return join('consent', folder_name, f"{language}-{filename}")
```
Note that your replacement is different, since the current setup writes a `en-consent.md` file, not to a folder `en` with `consent.md`.

backend/experiment/utils.py
```diff
@@ -187,5 +188,6 @@ def consent_upload_path(instance: Experiment, filename: str) -> str:
     experiment = instance.experiment
     folder_name = experiment.slug
     language = instance.language
+    new_filename = language + '-' + filename
 
-    return f"consent/{folder_name}/{language}-{filename}"
+    return join('consent', folder_name, new_filename)
```
BeritJanssen, line 193: See above.

BeritJanssen (changes requested):
See comments: perhaps replace the `filename + ".py"` with `f"{filename}.py"`, which I think is more readable. Definitely revisit the formatting for the consent filepath: when merging the PR as-is, they'd be written to a different path than is currently specified.

Evert-R:
> See comments: perhaps replace the `filename + ".py"` with `f"{filename}.py"`, which I think is more readable. Definitely revisit the formatting for the consent filepath: when merging the PR as-is, they'd be written to a different path than is currently specified.

Using f-strings do indeed improve readability a bit here.
The consent path, I thought I fixed this when I noticed it earlier, but must have forgotten. Good catch ! Thanks

BeritJanssen (approved):

Issue #1298, opened 2024-10-01
Add a timestamp for every action in matching pairs


Evert-R:
**Is your feature request related to a problem? Please describe.**
In matching pairs we set a timestamp in the results for the first and second click to calculate the interval.
I was asked by the congo group if we could calculate the time that the second had been playing, so we could also calculate the interval between a turn and the next click on the first card. At the moment there is no way to know how long the second was playing, only the time interval between the second click and the next first click. 

**Describe the solution you'd like**
- Log the timestamp for the third click in the next result. (this is the start of the turn) 
- log the timestamp for the click on the start button of the game in the first result. (this is the start of the first turn)

**Describe alternatives you've considered**

**Additional context**
Related to #965

PR #1350, opened 2024-11-05
Add timestamp for start of turn

Evert-R:
This PR sets a timestamp `start_of_turn` in the result of the first card.
First when the board has been loaded, then on every third click of a turn. 

Closes: #1298

### DIFF
--- a/frontend/src/components/MatchingPairs/MatchingPairs.tsx
+++ b/frontend/src/components/MatchingPairs/MatchingPairs.tsx
@@ -62,6 +62,8 @@ const MatchingPairs = ({
         yPosition.current = posY;
     }
 
+    let startOfTurn = performance.now();
+
     // Show (animated) feedback after second click on second card or finished playing
     const showFeedback = (score: number) => {
 
@@ -142,6 +144,7 @@ const MatchingPairs = ({
                 currentCard.noevents = true;
                 currentCard.boardposition = index + 1;
                 currentCard.timestamp = performance.now();
+                currentCard.start_of_turn = startOfTurn;
                 // reset response interval in case this card has a value from a previous turn
                 currentCard.response_interval_ms = '';
                 // clear feedback text
@@ -152,6 +155,7 @@ const MatchingPairs = ({
     };
 
     const finishTurn = () => {
+        startOfTurn = performance.now();
         finishedPlaying();
         // remove matched cards from the board
         if (score === 10 || score === 20) {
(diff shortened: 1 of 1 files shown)


BeritJanssen (commented):
In terms of analyzing data: I think it would be better to not set the `start_of_turn` variable on each card, but instead, send a `start_of_turn` variable within the `result` object. This will be saved to the results' json data. The `response_interval_ms` can probably be removed (check with @Lia-JX-Li ), as it can be derived from the timestamps.

frontend/src/components/MatchingPairs/MatchingPairs.tsx
```diff
@@ -62,6 +62,8 @@ const MatchingPairs = ({
         yPosition.current = posY;
     }
 
+    let startOfTurn = performance.now();
```
drikusroor, line 65: Does it work like this in React? 🤔 If written like this, I expect `startOfTurn` to be overwritten in every re-render... I would expect a state variable that gets set on initialization and to be set again using `setStartOfTurn` in the `finishTurn` function (same place as you now do the reassignment).

frontend/src/components/MatchingPairs/MatchingPairs.tsx
```diff
@@ -141,7 +142,8 @@ const MatchingPairs = ({
                 currentCard.turned = true;
                 currentCard.noevents = true;
                 currentCard.boardposition = index + 1;
-                currentCard.timestamp = performance.now();
+                setStartOfTurn(performance.now());
+                currentCard.start_of_turn = startOfTurn;
```
BeritJanssen, line 146:
```suggestion
```

frontend/src/components/MatchingPairs/MatchingPairs.tsx
```diff
@@ -152,6 +154,7 @@ const MatchingPairs = ({
     };
 
     const finishTurn = () => {
+        setStartOfTurn(performance.now());
```
BeritJanssen, line 157: Why set this here and at line 145? Wouldn't it be enough to do it in one of those two places?

frontend/src/components/MatchingPairs/MatchingPairs.tsx
Evert-R: It did work but using a state variable would indeed be better.

Evert-R:
> In terms of analyzing data: I think it would be better to not set the `start_of_turn` variable on each card, but instead, send a `start_of_turn` variable within the `result` object. This will be saved to the results' json data. The `response_interval_ms` can probably be removed (check with @Lia-JX-Li ), as it can be derived from the timestamps.

start_of_turn is indeed more clear, so I changed it back.  I removed the response_interval_ms after consulting with Lia.

PR #1356, opened 2024-11-08
Config: Add debugpy support and update Docker configuration

drikusroor:
Integrate debugpy for remote debugging and adjust Docker settings to facilitate its use, including necessary configurations in the Dockerfile and Docker Compose.


docker-compose.yaml
```diff
@@ -27,7 +27,7 @@ services:
         build:
             context: ./backend
             dockerfile: DockerfileDevelop
-            target: base
+            target: base # Set to debug to use debugpy
```
BeritJanssen, line 30: This is not ideal. Can't we make a script for this that would do something like
```
cd backend && docker run debug python -m debugpy ...
```
and leave docker-compose unchanged? Or would this mean we wouldn't get the necessary communication between the Django server and the debugpy server?

BeritJanssen (changes requested):
I'm personally fine with just using vanilla `pdb`, but anything that makes dev life easier is certainly worth adding. I do hope there's a way to add this without having to adjust the `docker-compose` file every time we want to run the debug server. BTW, I would expect the current state of docker-compose.yaml (i.e., having the backend image set to `base` and calling the `debugpy` function) to not run?

docker-compose.yaml
drikusroor: Yeah, good point. I'll try and see if a better solution can be found for this in the form of a script or dedicated environment variable.

docker-compose.yaml
```diff
@@ -53,8 +53,12 @@ services:
             - SQL_PASSWORD=${SQL_PASSWORD}
             - SQL_HOST=${SQL_HOST}
         ports:
+            - 5678:5678
             - 8000:8000
-        command: bash -c "python manage.py migrate && python manage.py bootstrap && python manage.py runserver 0.0.0.0:8000"
+        command: >
+            bash -c "python manage.py migrate &&
+                    python manage.py bootstrap &&
+                    ${DEBUG_COMMAND:-python manage.py runserver 0.0.0.0:8000}"
```
drikusroor, line 61: So this part runs the debug command as configured in your `.env` file if configured and runs the normal startup command if there is no debug command configured.

backend/aml/development_settings.py
```diff
@@ -25,6 +26,26 @@
...
+
+if DEBUG:
+    try:
+        import debugpy
+
+        if not is_debugpy_running(5678):
+            print("Starting debugpy on port 5678 🐛")
+            debugpy.listen(("0.0.0.0", 5678))
+        else:
+            print("Debugpy is already running on port 5678 👍")
```
drikusroor, line 44: This check has been added since the django server (but not the container) will restart once you change a file. The debug server will spawn a separate process as far as I understood and won't get terminated whenever the django server restarts.

PR #1361, opened 2024-11-12
Fix: create Experiment and Phase object in bootstrap.py

BeritJanssen:
I updated the `bootstrap.py` command, and added a unit test.


drikusroor (approved):
LGTM

PR #1363, opened 2024-11-13
Fixed: Add `random` and `interval` props to `Histogram` component

### LINKED ISSUE #1362: 🐛 [BUG] - Honest Histogram might influence Matching Pairs results by giving the participant visual information
### Description

As the `Histogram` component now shows bar changes based on actual audio frequency, it might give the participant too much information about the audio that is currently playing, which could potentially have an impact on the results.

### Reproduction URL

_No response_

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Logs

_No response_

### Browsers

_No response_

### OS

_No response_

### Additional context

_No response_

drikusroor:
Re-added the random histogram bar changes as not to "help" the participant in the matching pairs experiment.

This PR adds the `random` prop (`false` by default) that makes the `Histogram` show random bar changes again. Additionally, the `interval` prop has been re-instated, which is used as the interval for random bar changes.

The `PlayCard` component then uses the `random` feature of the `Histogram` as it is used by the Matching Pairs game.

Fixes #1362

### DIFF
--- a/frontend/src/components/MatchingPairs/PlayCard.tsx
+++ b/frontend/src/components/MatchingPairs/PlayCard.tsx
@@ -56,6 +56,7 @@ const PlayCard = ({ onClick, registerUserClicks, playing, section, view, showAni
                         bars={histogramBars}
                         backgroundColor="purple"
                         borderRadius=".5rem"
+                        random={true}
                     />
                 :
                 <div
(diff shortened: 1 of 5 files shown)


Evert-R (approved):
Nice! 
For my taste it would be better to set the `interval` in the playcard  to `200`
I find `100` to be a bit distracting.

frontend/src/components/MatchingPairs/PlayCard.tsx
```diff
@@ -56,6 +56,8 @@ const PlayCard = ({ onClick, registerUserClicks, playing, section, view, showAni
                         bars={histogramBars}
                         backgroundColor="purple"
                         borderRadius=".5rem"
+                        random={true}
+                        interval={100}
```
Evert-R, line 60: Suggestion: set interval to `200`

drikusroor:
```suggestion
                        interval={200}
```

Issue #1334, opened 2024-10-31
Progress bar in title


BeritJanssen:
**Is your feature request related to a problem? Please describe.**
Instead of showing how many questions, blocks etc. have been already completed numerically (e.g., 1 / 7), include a progress bar in the title of the page.

drikusroor:
For a specific experiment, or for all experiments? And with regards to the technical implementation: Modify the `next_round` to include the current round number and the total number of rounds, or regex the title in the frontend to deduct the progress as a percentage?

drikusroor:
As for the way it should look like, I've been working a bit on a progress bar a bit and came up with this:

<img width="1085" alt="image" src="https://github.com/user-attachments/assets/52e67784-79b7-4756-8995-ca1fd9d77a65">

Do you think something like this would work or should it be more subtle like a little line beneath the title? 🤔

BeritJanssen:
@Lia-JX-Li what do you think?

Lia-JX-Li:
I like this! Is the color, the text inside, and the position of the progress bar be customized later?

BeritJanssen:
The color and text can be customized, but not the position.

PR #1364, opened 2024-11-14
Added: Add `ProgressBar` component

drikusroor:
Introduce a new ProgressBar component that allows customization of properties and styles. Enhance the associated stories with decorators for improved visualization.

The component is not yet configured to be used in the `AppBar` as it kind of depends on the technical approach we want to take with regards to the content. See also the related issue of #1334

### DIFF
--- a/frontend/src/components/ProgressBar/ProgressBar.scss
+++ b/frontend/src/components/ProgressBar/ProgressBar.scss
@@ -1,0 +1,66 @@
+// import variables.scss
+@import '../../scss/variables.scss';
+
+// Progress bar variables
+$progress-height-sm: 20px; // Increased to accommodate text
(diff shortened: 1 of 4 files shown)


BeritJanssen (approved):
Nice!
As we'll typically have max values around 5-10 (i.e., different questions or experiments), I wonder whether percentage is all that useful to have. It's nice to have, but it should be disabled by default.

drikusroor:
> Nice! As we'll typically have max values around 5-10 (i.e., different questions or experiments), I wonder whether percentage is all that useful to have. It's nice to have, but it should be disabled by default.

I'll remove it since we will most likely only use the progress bar in this location. If we need it in the future it's easy to add, but for now it means a feature that we would need to maintain.

Issue #1352, opened 2024-11-05
BUG - Duplicates are possible for Experiment.slug


Evert-R:
While working on the experiment's duplicate function I noticed that the slug field on ```Experiment``` has not been set to Unique, so it's now possible to save multiple experiments with the same slug, which will not run, because only one experiment is expected.

#### I tried to migrate it the same way it is in ```Block```:
```slug = models.SlugField(max_length=64, unique=True, validators=[experiment_slug_validator])```

#### but I got this error:
```
Applying experiment.0059_alter_experiment_slug...Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/django/db/backends/utils.py", line 87, in _execute
    return self.cursor.execute(sql)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/psycopg/cursor.py", line 732, in execute
    raise ex.with_traceback(None)
psycopg.errors.DuplicateTable: relation "experiment_experiment_slug_63cd0f17_like" already exists

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/server/manage.py", line 24, in <module>
    main()
  File "/server/manage.py", line 20, in main
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.11/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.11/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.11/site-packages/django/core/management/base.py", line 412, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.11/site-packages/django/core/management/base.py", line 458, in execute
    output = self.handle(*args, **options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/core/management/base.py", line 106, in wrapper
    res = handle_func(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/core/management/commands/migrate.py", line 356, in handle
    post_migrate_state = executor.migrate(
                         ^^^^^^^^^^^^^^^^^
[...]

PR #1366, opened 2024-11-15
Bugfix er/make experiment slug unique

Evert-R:
This PR sets the `slug` field on the `Experiment` model to be unique.

First the slug data gets migrated to a temporary field, then the original slug field is deleted, after which the temporary slug is renamed to `slug`.

Directly changing the slug field with `unique=True` didn't work, in every which way i tried, even when no experiments were present in the database.

closes: #1352


BeritJanssen (approved):
Nice! I do think you might be able to squash the migrations, so it's just one file?

Issue #1337, opened 2024-10-31
Landing page before Consent


BeritJanssen:
**Is your feature request related to a problem? Please describe.**
At the moment, if a Consent form is attached to an Experiment, it will always get shown first when a new participant arrives on the page. The TuneTwins team requested that it would be possible to show a page before this, with information about the reasoning and team behind the game.

**Describe the solution you'd like**
This could be implemented without extra development time by referring participants to a blog post with relevant information on the Amsterdam Music Lab website.

**Describe alternatives you've considered**
Having a landing page within MUSCLE could still have its merits, though: imagine an Experiment that has one Block that starts immediately with Autoplay. If we have a returning Participant, i.e., they don't get the Consent form, they would not get to hear the audio because the required user interaction is missing. The landing page could provide the extra click needed to get the Autoplay to work. Still, this is a bit of an edge case, and doesn't actually apply for the MatchingPairs game.

BeritJanssen:
I just discussed with @Lia-JX-Li , and the advantages of a landing page over a blog post would be:
- there is only going to be *one* URL that participants, rather than the AML website and the experiment link
- participants won't get distracted by the AML homepage, where they may potentially end up elsewhere by clicking on other content

I was just thinking that perhaps we could use the `aboutContent` field to populate an HTML page with information. Then, the sharing link would be `app.amsterdammusiclab.nl/my-awesome-experiment/about`. That also wouldn't show consent. Clicking on a button in the about page will take the participant to the experiment proper, and the consent page. If participants share the experiment with each other, we can make sure they share the about page by adjusting the link through the admin interface.

PR #1367, opened 2024-11-15
add experiment about page as Router option

BeritJanssen:
This would be an option towards #1337 : the about page can also function as a landing page.

### DIFF
--- a/frontend/src/components/App/App.tsx
+++ b/frontend/src/components/App/App.tsx
@@ -86,7 +86,7 @@ const App = () => {
                         <Route path={URLS.experiment} element={<Experiment />} />
 
                         {/* ExperimentAbout */}
-                        <Route path={URLS.experiment} element={<ExperimentAbout />} />
+                        <Route path={URLS.experimentAbout} element={<ExperimentAbout />} />
 
                         {/* Store profile */}
                         <Route
--- a/frontend/src/components/App/App.tsx
+++ b/frontend/src/components/App/App.tsx
@@ -11,6 +11,7 @@ import { URLS as API_URLS } from "../../API";
 import useBoundStore from "../../util/stores";
 import Block from "../Block/Block";
 import Experiment from "../Experiment/Experiment";
+import ExperimentAbout from "../Experiment/ExperimentAbout/ExperimentAbout";
 import LoaderContainer from "../LoaderContainer/LoaderContainer";
 import ConditionalRender from "../ConditionalRender/ConditionalRender";
 import Profile from "../Profile/Profile";
@@ -83,6 +84,9 @@ const App = () => {
 
                         {/* Experiment */}
                         <Route path={URLS.experiment} element={<Experiment />} />
+
+                        {/* ExperimentAbout */}
+                        <Route path={URLS.experiment} element={<ExperimentAbout />} />
 
                         {/* Store profile */}
                         <Route


frontend/src/components/App/App.tsx
```diff
@@ -84,6 +85,9 @@ const App = () => {
                         {/* Experiment */}
                         <Route path={URLS.experiment} element={<Experiment />} />
 
+                        {/* ExperimentAbout */}
+                        <Route path={URLS.experimentAbout} element={<ExperimentAbout />} />
+
```
drikusroor, line 90: I am a bit confused :-) Is this supposed to be an about page for an experiment? In that case, I don't think this route will work properly as there the route / page component won't know for which experiment they will have to show the about page. I think it might be better to modify the `Experiment` component's logic to decide whether we want to show the consent already if the user wants to go to the experiment's about page.

Either here by checking the current location and see if the current location contains `"/about"`:

https://github.com/Amsterdam-Music-Lab/MUSCLE/blob/b15fd1032a63e746f8ecb818d9ba15ff2b2a8aca/frontend/src/components/Experiment/Experiment.tsx#L59-L71

Or here, by moving the logic that decides whether we want to show the consent form to a separate "Guard Component" (and something similar for the logic that decides whether participant will be redirected to next experiment instead of the dashboard):

```tsx
export const RequireConsent = ({ experiment, hasShownConsent, children }: RequireConsentProps) => {
    const showConsent = experiment?.consent;
    
    if (!hasShownConsent && showConsent) {
        return <Navigate to={`/experiment/${experiment.slug}`} replace />;
    }

    return <>{children}</>;
};
```

```tsx
export const RequireDashboardOrRedirect = ({ experiment, children }: RequireDashboardOrRedirectProps) => {
    const displayDashboard = experiment?.dashboard.length;
    const nextBlock = experiment?.nextBlock;
    const participantIdUrl = experiment?.participantIdUrl;
    
    if (!displayDashboard && nextBlock) {
        return <Navigate 
            to={`/block/${nextBlock.slug}${participantIdUrl ? `?participant_id=${participantIdUrl}` : ""}`} 
            replace 
        />;
    }

    return <>{children}</>;
};
```

```tsx
            <Routes>
                {/* About route is always accessible */}
                <Route
                    path="/about"
                    element={<ExperimentAbout content={experiment?.aboutContent} slug={experiment.slug} />}
                />
                
                {/* Main experiment route with nested guards */}
                <Route
                    path="*"
                    element={
                        <RequireConsent experiment={experiment} hasShownConsent={hasShownConsent}>
                            <RequireDashboardOrRedirect experiment={experiment}>
                                <ExperimentDashboard 
                                    experiment={experiment} 
                                    participantIdUrl={participantIdUrl} 
                                    totalScore={experiment.totalScore} 
                                />
                            </RequireDashboardOrRedirect>
                        </RequireConsent>
                    }
                />
            </Routes>
```

drikusroor (changes requested):
I don't think the proposed code will do what it is supposed to do and my advice would be to improve the logic in the `Experiment` component (see my comment).

BeritJanssen:
Yup, I rushed this PR and forgot to pass the slug in the `Route`, silly mistake. When you do though, this (more specific) route `{slug}/about` should take precedence over the `{slug}` route, right? I do think it's a good idea to rework the `Experiment` component and factor out the consent & redirect/dashboard logic into their own components. Perhaps open an issue for it.

As we now got yet another feature request for a landing page, which cannot be solved with this approach (as it specifies a landing page for two different experiments), closing this PR. I'm now thinking of a dedicated Django `landing` app and route, and a React Landing component, which would probably be no more than a wrapper around HTML entered from the Django admin.

PR #1375, opened 2024-11-18
Docs: Enhance ImageAdminForm with additional fields and help texts

### LINKED ISSUE #1369: Documentation: add descriptions to Image fields
Right now, the purpose of many fields in the `Image` model isn't documented:
Screenshot 2024-11-15 at 14 31 29

Fields such as "alt", "href", "rel" and "target", but also "tags" would need some extra help text.

drikusroor:
Add help texts to the `ImageAdminForm` for improved user guidance during image uploads and management.

## Screenshot

image

Resolves #1369

### DIFF
--- a/backend/image/forms.py
+++ b/backend/image/forms.py
@@ -62,8 +62,7 @@ class ImageAdminForm(forms.ModelForm):
         max_length=255,
         required=False,
         initial="noreferrer",
-        help_text='The relationship attribute for the link (e.g., "nofollow" for SEO, '
-        '"noopener noreferrer" for security when linking to external sites)',
+        help_text='Choose "nofollow" to prevent search engines from following the link, "noopener" to ensure security when opening links in new tabs, or "noreferrer" for stricter privacy when linking to external sites.',
     )
 
     target = forms.ChoiceField(
--- a/backend/image/forms.py
+++ b/backend/image/forms.py
@@ -3,7 +3,7 @@ from django import forms
 from django.core.exceptions import ValidationError
 from django.forms.widgets import ClearableFileInput
 from django.contrib.postgres.forms import SimpleArrayField
-from .models import Image
+from .models import Image, TARGET_CHOICES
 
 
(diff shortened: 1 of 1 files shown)


backend/image/forms.py
```diff
@@ -31,14 +31,57 @@ def value_from_datadict(self, data, files, name):
...
+    description = forms.CharField(
+        required=False,
+        widget=forms.Textarea,
+        help_text="A detailed description of the image content - useful for content management",
+    )
+
+    alt = forms.CharField(
+        max_length=255,
+        required=False,
+        help_text="Alternative text that describes the image - important for accessibility and SEO. "
```
BeritJanssen, line 51:
```suggestion
        help_text="Alternative text that describes the image. "
```
IMO, this part is too technical. With the next sentence, it's already crystal clear.

backend/image/forms.py
```diff
@@ -31,14 +31,57 @@ def value_from_datadict(self, data, files, name):
...
+        required=False,
+        help_text="The URL where the image should link to when clicked. "
+        "Leave empty if the image should not be clickable.",
+    )
+
+    rel = forms.CharField(
+        max_length=255,
+        required=False,
+        help_text='The relationship attribute for the link (e.g., "nofollow" for SEO, '
+        '"noopener noreferrer" for security when linking to external sites)',
```
BeritJanssen, line 65: Even though `rel` is ir*rel*evant if the image is not clickable: perhaps it's sensible to define `noreferrer` as a default? Not all users will know what to do with this. BTW `noopener` is redundant if `noreferrer` is set, if [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noreferrer) may be believed.

backend/image/forms.py
drikusroor:
```suggestion
        help_text="Alternative text that describes the image - important for accessibility and SEO. This text is read aloud by screen readers or whenever the image cannot be loaded."
```

PR #1392, opened 2024-11-22
Added: Add collapsible functionality to block forms with expand/collapse all buttons

### LINKED ISSUE #1380: Blocks in Experiment form are collapsible
Now we can collapse *all* blocks within a phase. Instead, we want to be able to collapse specific blocks, and otherwise show information such as slug and name by which users can easily identify blocks.

The collapsed `Block` form should should allow users to do configure everything for the block, including questions.

drikusroor:
Introduce collapsible functionality to block forms, enhancing user experience with improved visibility and layout. Update styles and documentation to support the new features.

Resolves #1380

### Collapsed state

image

### One inline block form expanded

image

### DIFF
--- a/backend/experiment/admin.py
+++ b/backend/experiment/admin.py
@@ -148,7 +148,6 @@ class BlockAdmin(InlineActionsModelAdminMixin, admin.ModelAdmin):
                 data=str(serializers.serialize("json", all_feedback.order_by("pk"))),
             )
 
-
         # create forced download response
         response = HttpResponse(zip_buffer.getbuffer())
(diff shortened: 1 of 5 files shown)


BeritJanssen:
I like it! As discussed in another channel, I do wonder whether it wouldn't make sense to show the tabular overview of the blocks in a phase at all times, and use the arrow next to "Blocks" to expand/collapse all Block forms instead.

BeritJanssen (approved):
Looks good! Let's perhaps still look at this together next week? We can of course also merge as is then and refine later though.

PR #1396, opened 2024-11-22
catch for duplictate slugs already in database

Evert-R:
This PR adds a catch for experiments with duplicate slugs already in the database when migrating the slug to a unique field.


drikusroor (approved):
Excellent :-) LGTM

PR #1397, opened 2024-11-25
Refactor: Integrate Round management and enhance Histogram component

drikusroor:
Introduce a new `RoundResponse` type for round management in the Block component, streamline the Action type definition, and enhance the Histogram component with improved randomization logic. Fix various type annotations and update tests to ensure proper functionality.

## What does this PR do?

- It commits the `next_round` data to the `RoundSlice` in the zustand store every time a new round is received from a `next_round` http request.
- It also types the array of actions in `next_round` as "`Round`', which is basically an array of `Action`.
- It also keeps track of the "current action" by setting the index of the current action of the current round in the store.
- This store management of next round & action is currently done in parallel to the way the current round & action(s) are managed in the `Block` component. It has no effect on that functionality. Tracking said data in our zustand store is only used for the `Histogram` component to know whether the current action's section(s) are loaded into the buffer (`AudioContext`) or whether it is loaded externally through the "normal" `Audio` element/API.
- The `Histogram` then uses that information to determine whether it should display actual audio frequency data (only possible through buffer) or random bar data.
  - It will show random data whenever the `random` prop is true, or whenever the current action's section is _not_ played through buffer (aka `!isBuffer`)

## Questions

- Should we now or later convert the whole bookkeeping of round & (current) action(s) to use the zustand store?
- In the `Action` type I now import interfaces (used for different action types) from the components directly. Should I import those into the `Action` type file?
- Should we maybe rename the `next_round` prop in the next round request to `actions`? The http response would then become `{ "actions": [{...}, {...}, {...}] }`, which might make more sense.

### DIFF
--- a/frontend/src/components/Block/Block.tsx
+++ b/frontend/src/components/Block/Block.tsx
@@ -71,7 +71,7 @@ const Block = () => {
         const currentAction = newState ? newState : null;
         setCurrentAction({ ...currentAction });
         updateState(newState);
-    }, [updateState]);
+    }, [updateState, setCurrentAction]);
 
     const continueToNextRound = async (activeSession: Session) => {
         // Try to get next_round data from server
--- a/frontend/src/components/Block/Block.tsx
+++ b/frontend/src/components/Block/Block.tsx
@@ -56,6 +56,9 @@ const Block = () => {
     const setTheme = useBoundStore((state) => state.setTheme);
     const resetTheme = useBoundStore((state) => state.resetTheme);
     const setBlock = useBoundStore((state) => state.setBlock);
+    const setRound = useBoundStore((state) => state.setRound);
+    const setCurrentAction = useBoundStore((state) => state.setCurrentAction);
+    const currentActionIndex = useBoundStore((state) => state.currentActionIndex);
 
     const setHeadData = useBoundStore((state) => state.setHeadData);
     const resetHeadData = useBoundStore((state) => state.resetHeadData);
@@ -67,7 +70,7 @@ const Block = () => {
     const playlist = useRef(null);
 
     // API hooks
-    const [block, loadingBlock] = useBlock(slug);
+    const [block, loadingBlock] = useBlock(slug!);
 
     const loadingText = block ? block.loading_text : "";
     const className = block ? block.class_name : "";
@@ -83,6 +86,7 @@ const Block = () => {
     const updateActions = useCallback((currentActions: []) => {
         const newActions = currentActions;
         setActions(newActions);
+        setCurrentAction(0);
         const newState = newActions.shift();
         updateState(newState);
     }, [updateState]);
@@ -93,6 +97,7 @@ const Block = () => {
             session: activeSession
         });
         if (round) {
+            setRound({ ...round.next_round }); // Make a deep copy of the round as the 'round' object will be mutated by the updateActions
             updateActions(round.next_round);
         } else {
             setError(
@@ -106,6 +111,7 @@ const Block = () => {
     const onNext = async (doBreak = false) => {
         if (!doBreak && actions.length) {
             updateActions(actions);
+            setCurrentAction(currentActionIndex! + 1); // Increment current action index
         } else {
             continueToNextRound(session as Session);
         }
--- a/frontend/src/components/Block/Block.tsx
+++ b/frontend/src/components/Block/Block.tsx
@@ -39,9 +39,7 @@ const Block = () => {
     const setTheme = useBoundStore((state) => state.setTheme);
     const resetTheme = useBoundStore((state) => state.resetTheme);
     const setBlock = useBoundStore((state) => state.setBlock);
-    const setRound = useBoundStore((state) => state.setRound);
     const setCurrentAction = useBoundStore((state) => state.setCurrentAction);
-    const currentActionIndex = useBoundStore((state) => state.currentActionIndex);
 
     const setHeadData = useBoundStore((state) => state.setHeadData);
     const resetHeadData = useBoundStore((state) => state.resetHeadData);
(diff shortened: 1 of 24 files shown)


frontend/src/components/Block/Block.tsx
```diff
@@ -93,6 +80,7 @@ const Block = () => {
             session: activeSession
         });
         if (round) {
+            setRound({ ...round.next_round }); // Make a deep copy of the round as the 'round' object will be mutated by the updateActions
```
drikusroor, line 83: As `updateActions` performs a `.shift()`, it mutates the object, which is (unwantedly) also reflected in the store.

frontend/src/components/Block/Block.tsx
```diff
@@ -106,6 +94,7 @@ const Block = () => {
     const onNext = async (doBreak = false) => {
         if (!doBreak && actions.length) {
             updateActions(actions);
+            setCurrentAction(currentActionIndex! + 1); // Increment current action index
```
drikusroor, line 97: There shouldn't be any problems with this whenever a new round occurs since we set the index back to `0` when that happens.

frontend/src/stories/Playback.stories.tsx
drikusroor, line 1: Here, I add & modify the stories so that it shows an example of a `Histogram` portraying audio frequency data (buffer) and one that shows random data (external), both inside of a `Playback` component.

frontend/src/types/Action.ts
```diff
@@ -0,0 +1,26 @@
+import { ExplainerProps } from "@/components/Explainer/Explainer";
+import { FinalProps } from "@/components/Final/Final";
+import { InfoProps } from "@/components/Info/Info";
+import { LoadingProps } from "@/components/Loading/Loading";
+import { PlaylistProps } from "@/components/Playlist/Playlist";
+import { RedirectProps } from "@/components/Redirect/Redirect";
+import { ScoreProps } from "@/components/Score/Score";
+import { TrialProps } from "@/components/Trial/Trial";
```
drikusroor, line 8: Yeah, so here I'm importing the different action type props from their components, but I'm thinking about importing them into here as they are ultimately props/data that come directly from the backend.

frontend/src/types/Round.ts
```diff
@@ -0,0 +1,7 @@
+import { Action } from "./Action";
+
+export type Round = Action[];
```
drikusroor, line 3: This may seem a bit weird, but since the prop is called `next_round`, I consider the array of actions a "round" and therefore type it as such. Perhaps we can think about better naming if this is too confusing. For example, we could also rename (both in the frontend and backend) `next_round` to `actions` and then the `next_round` http request will respond with an object like this:

```json
{
  "actions": [{...}, {...}, {...}]
}
```

which might make more sense.

frontend/src/util/stores.ts
```diff
@@ -96,6 +97,28 @@ const createParticipantSlice: StateCreator<ParticipantSlice> = (set) => ({
...
+    setRound: (round: Round) => set(() => ({ round })),
+    currentActionIndex: null,
+    setCurrentAction: (index: number) => set(() => ({ currentActionIndex: index })),
+    currentAction: () => {
+        const index = get().currentActionIndex;
+        if (index === null) {
+            return null;
+        }
+        return get().round[index];
+    },
```
drikusroor, line 119: This is a computed getter function that we could start using more often to simplify getting data that might need some logic/computation to retrieve.

frontend/src/types/Action.ts
BeritJanssen: I agree that it would be clearer to have all different types of `Action` in one place.

frontend/src/types/Round.ts
BeritJanssen: I'm not too sure about this change. It is indeed weird that a "round" refers to an array of actions - at the same time, it does make sense that a participant has to take a number of actions within a round. If we document it, I don't think it's too confusing to keep handling *round* / *next_round* references. If we were to change the name of the API call, I would argue that would mean we'd also have to change the name of the `next_round` function in all the rule files, and all other references to `round` in the backend. Plus, referring to `currentActions` (i.e., the array of actions last retrieved from the backend) vs. `currentAction` (i.e., the view that's presented to the user right now) would be confusing, too.

BeritJanssen (changes requested):
I like it! I do wonder whether we need to keep information on the current round, as well as the current action, in the store. For all the child components of `Block`, it would be neat to retrieve information about the current *action* for sure, but I don't think we will need double bookkeeping of the *array* of actions, and can just keep that in the `Block` component, right? Even though the `onNext` function that's drilled into the component tree is perhaps also not ideal, my estimate is that the refactoring involved to move this part of state management to the store is a bit too much "hay on the fork".

frontend/src/types/Round.ts
drikusroor: Ah, but I would not change the name of the api call. You request the next round, it gives you back an object that represents the next round, and that object contains actions.

drikusroor:
> I like it! I do wonder whether we need to keep information on the current round, as well as the current action, in the store. For all the child components of `Block`, it would be neat to retrieve information about the current _action_ for sure, but I don't think we will need double bookkeeping of the _array_ of actions, and can just keep that in the `Block` component, right? Even though the `onNext` function that's drilled into the component tree is perhaps also not ideal, my estimate is that the refactoring involved to move this part of state management to the store is a bit too much "hay on the fork".

Well, in this solution the `Histogram` _needs_ the information about the current action from the store as it needs to know whether the current action has buffer audio or not.

What's also possible is simplifying the whole `RoundSlice` and make it only have a current action (which is then not an index, but the action object itself). There would be less bookkeeping and the `Histogram` still gets the right information. I think.

BeritJanssen:
Yes, that's what I meant: let the store only keep tabs on the current action, not the current round.

Issue #1381, opened 2024-11-19
Move Block feedback from admin to Experimenter Dashboard


Evert-R:
User feedback is now shown in  the admin interface, where it can also be modified.
This should be moved to the experimenter dashhbvoard, as read only.

PR #1398, opened 2024-11-26
Move feedback to experimenter dashboard

Evert-R:
This PR moves the feedback from the Block admin, where it could be edited, to the Experimenter dashboard, where it's read only.

closes: #1381

### DIFF
--- a/backend/experiment/admin.py
+++ b/backend/experiment/admin.py
@@ -44,14 +44,6 @@ from participant.models import Participant
 from question.models import QuestionSeries, QuestionInSeries
 
 
-class FeedbackInline(admin.TabularInline):
-    """Inline to show results linked to given participant"""
-
-    model = Feedback
-    fields = ["text"]
-    extra = 0
-
-
 class BlockTranslatedContentInline(NestedTabularInline):
     model = BlockTranslatedContent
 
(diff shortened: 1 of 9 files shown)


drikusroor (changes requested):
Overall, I think it looks really good and clear already. I have no comments on the code whatsoever, but I did notice some things that I think could be improved from a user experience perspective:

image

1. In this screenshot both experiment and block are named `teletunes`, but this part shows the experiment name. I think it would be more interesting for the user to show the block name ("Feedback for {block name}") and not the experiment name, or at least both ("Feedback for {experiment name} -> {block name}"). The user zooming in on the feedback of a specific block after all.
2. I don't think "Feedback for block" adds much value here - we already know we are looking at feedback - and maybe it would be more interesting to:
3. Add the date and time in a readable localized way here.
4. This is just a personal opinion of mine, but I would go for a grayscale shadow instead of a blueish shadow 😅

Evert-R:
1 - The title is the same for the `all feedback` list  and the filtered block feedback list. (it's outside the iteration)
2 - This should have contained the name or slug of the block. I changed this now from `item.block.name` to `item.block`, so if the block doesn't have content providing a name, it will fallback to the slug, as defined in the model. 
3 - The date of the feedback isn't logged. Only the `block.id` and the text. We could add that to the model, ofcourse, but this could be intentionally , so we can't link the feedback to a specific participant.
4 - I can live with grey ;-)

- I also added links to the 'Block' and 'Experiment` edit pages where possible.
- I added the name (or fallback to slug) to the exported blocks, because I noticed we don't export the pure querylist anymore, so we can't rely on the fallback defined in the `Block` model in the template.

Issue #1368, opened 2024-11-15
🐛 [BUG] - ONE PAIR BEING SILENT on original board of Matching Pairs Congo version 


Lia-JX-Li:
### Description

Dear MUSCLE Team,

After successfully collecting data in the jungle using Toughbook, we are now planning to conduct the same experiment with Dutch children. In preparation, I am filing this issue for your attention. Thank you in advance for looking into it!

Bug Description:
It seems that the issue arises specifically when loading the original board. One of the eight pairs does not play any sound. Interestingly, this problem does not occur with every board and is mostly resolved by reloading and generating a new board.

Let me know if you need any additional details or steps to reproduce the issue.

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

BeritJanssen:
Seems to me like one of the files cannot be found. Can you send the playlist csv and files?

Lia-JX-Li:
I will attach it tomorrow morning when I'm in office and have the toughbook at hand!

Lia-JX-Li:
[original.csv](https://github.com/user-attachments/files/17812592/original.csv)

PR #1403, opened 2024-11-26
Use buffers object to check wether a section has already been loaded

Evert-R:
This PR fixes the bug that occurs when you play `matching_pairs_lite` for the second time (via the dashboard) with 8 originals.

When the sections are loaded the buffer list gets cleared before loading the first section.
we used `previousSource`  to determine wether a section had already been loaded. 
A section might still be loaded, but if it isn't added to the `buffers` object, it will not be passed on to matching pairs.

I now changed this it to checking the `buffers` object to check if a section has been loaded. 

Fixes: #1368

### DIFF
--- a/frontend/src/components/Preload/Preload.tsx
+++ b/frontend/src/components/Preload/Preload.tsx
@@ -46,17 +46,15 @@ const Preload = ({ sections, playMethod, duration, preloadMessage, pageTitle, on
 
                 sections.forEach((section, index) => {
 
+                    // skip Preload if the section has already been loaded in the previous action
+                    if (webAudio.checkSectionLoaded(section) && sections.length === 1) {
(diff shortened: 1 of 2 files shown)


BeritJanssen (commented):
As discussed IRL, make sure this change doesn't introduce a loading time in Hooked experiments, for which the check against `previousSource` was introduced. Perhaps the store slice for actions added per #1397 can be the basis of a more robust solution.

Evert-R:
> As discussed IRL, make sure this change doesn't introduce a loading time in Hooked experiments, for which the check against `previousSource` was introduced. Perhaps the store slice for actions added per #1397 can be the basis of a more robust solution.

I made some changes for previously loaded single section requests.

PR #1409, opened 2024-11-28
Fixed(django-select2): Add reinitialization for django-select2 on form change

### LINKED ISSUE #1408: 🐛 [BUG] - The `django_select2` multiselect used to select playlists for blocks in the experiment admin form does not work for newly added blocks
### Description

The multiselect works perfectly for an existing experiment with existing blocks, but whenever you hit the "Add another block" button, the multiselect for that block's playlist _exists_ but no cursor or dropdown appears when you click it. I _assume_ some Javascript function from `django_select2` needs to be initialized or something.

Possibly related to https://github.com/codingjoe/django-select2/issues/249

drikusroor:
Implement a reinitialization function for django-select2 to ensure it handles new inlines correctly. This addresses the issue of select2 widgets not being instantiated on newly added forms.

Resolves #1408

### DIFF
--- a/backend/experiment/static/experiment_form.js
+++ b/backend/experiment/static/experiment_form.js
@@ -1,6 +1,10 @@
 document.addEventListener("DOMContentLoaded", function () {
     initCollapsibleInlineForms();
     floatingSubmitRow();
+
+    // TODO: Remove this in the future once the issue is fixed in django-select2
+    // re-initialize django-select2 on form change
+    reinitializeDjangoSelect2OnFormChange();
 });
 
 function initCollapsibleInlineForms() {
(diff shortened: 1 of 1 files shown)


backend/experiment/static/experiment_form.js
```diff
@@ -60,3 +64,43 @@ function floatingSubmitRow() {
...
+            // Because select2 was already instantiated on the empty form, we need to remove it, destroy the instance,
+            // and re-instantiate it.
+            jqRow.find('.django-select2').parent().find('.select2-container').remove()
+            jqRow.find('.django-select2').djangoSelect2('destroy');
+            jqRow.find('.django-select2').djangoSelect2()
+        }
+
+        // See: https://docs.djangoproject.com/en/dev/ref/contrib/admin/javascript/#supporting-versions-of-django-older-than-4-1
+        django.jQuery(document).on('formset:added', function (event, $row, formsetName) {
+            if (event.detail && event.detail.formsetName) {
```
BeritJanssen, line 96: Since we pinned the Django version, is this `if/else` necessary?

BeritJanssen (approved):

backend/experiment/static/experiment_form.js
drikusroor: In a normal world I would say no, but somehow the `formset:added` event is dispatched twice whenever adding a new block form, one in the >= 4.1 and one in the < 4.1 format... I don't know which one is the "good" one. Both seem to accomplish in initializing the select2 input correctly and having the function be called twice also does not seem to cause any harm. I was thinking of leaving it like this until the library releases a real fix.

PR #1412, opened 2024-11-29
Fixed(th3plink): Fix session method and format code

drikusroor:
Replace the undefined `session.round_passed()` method with the correct `session.get_rounds_passed()` method and format the code in `toontjehoger_3_plink.py`.


BeritJanssen (approved):
We should also add unit tests for th plink (and three other ToontjeHoger rules), but that can be left for another branch.

PR #1413, opened 2024-12-02
Fixed: show correct song in Hooked experiments

### LINKED ISSUE #1405: 🐛 [BUG] - Teletunes incorrectly shows "You have listened to: <song name>" of the next song
### Description

I was just playing Teletunes on Acceptance and heard the intro theme song of Dexter. After I listened to the audio fragment and answered the questions, it told me I was correct, but that I had been listening to the theme song of Game of Thrones (2011).

I suspect the backend is returning a `Score` action that contains the right score but the wrong song name.

image

### Reproduction URL

https://acc.amsterdammusiclab.nl/block/teletunes

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Logs

_No response_

### Browsers

_No response_

### OS

_No response_

### Additional context

_No response_

BeritJanssen:
Close #1405.

### DIFF
--- a/backend/experiment/actions/score.py
+++ b/backend/experiment/actions/score.py
@@ -7,11 +7,23 @@ from session.models import Session
 from .base_action import BaseAction
 
 
-class Score(BaseAction):  # pylint: disable=too-few-public-methods
+class Score(BaseAction):
     """
-    Provide data for an intermediate score view
+    Provide data for a score view, presenting feedback to a participant after a Trial
+    Relates to client component: Score.ts
 
-    Relates to client component: Score.js
+    Args:
+        session: a Session object
+        title: the title of the score page
+        result: the result for which section and/or score should be reported
+        score: the score to report, will override result.score
+        score_message: a function which constructs feedback text based on the score
+        config: a dict with the following settings:
+        - show_section: whether metadata of the previous section should be shown
+        - show_total_score: whether the total score should be shown
+        icon: the name of a themify-icon shown with the view or None
+        timer: int or None. If int, wait for as many seconds until showing the next view
+        feedback: An additional feedback text
     """
 
     ID = 'SCORE'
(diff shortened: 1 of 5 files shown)


backend/experiment/actions/score.py
```diff
@@ -31,7 +46,9 @@ def __init__(self, session, title: str = None, score=None, score_message=None, c
             get_rounds_passed=session.get_rounds_passed(),
             total_rounds=self.session.block.rounds
         )
-        self.score = score or session.last_score()
+        self.score = (
+            score if score else result.score if result else session.last_score()
```
drikusroor, line 50: This line confuses me a bit. Can you maybe split it up in two lines? Also, can we assume that `result` will always be defined if there is no `score` and `result.score` won't throw an error?

backend/experiment/rules/tests/test_hooked.py
```diff
@@ -103,7 +103,11 @@ def test_hooked(self):
                 self.assertEqual(session.result_set.filter(question_key="correct_place").count(), 1)
             elif i == 1:
                 self.assertEqual(len(actions), 4)
-                self.assertEqual(type(actions[0]), Score)
+                score_action = actions[0]
+                self.assertEqual(type(score_action), Score)
+                self.assertIsNotNone(score_action.last_song)
+                # the last_song method returns the most recent result, without filtering
+                self.assertNotEqual(score_action.last_song, session.last_song())
```
drikusroor, line 110: What's the difference between a song and a result in this context?

drikusroor:
You might want to (also) merge this to `main` instead of `develop` as this currently does not work on production

backend/experiment/actions/score.py
BeritJanssen: There is another check `if result`, so both can be undefined. It's a fallback to a fallback, really. I'll write proper if / else clauses which might make this more clear.

PR #1416, opened 2024-12-10
Added: Enhance MatchingPairs game with tutorial overlay

### LINKED ISSUE #1338: Tutorial mode in MatchingPairs
The TuneTwins team decided that the best way to introduce new participants to the MatchingPairs game would be to give feedback during the first game board in the form of overlays.

We could pass along a `tutorial` dict from the backend along those lines:
```python
tutorial = {
  "no_match": "This was not a match, so you get 0 points. Please try again to see if you can find a matching pair.",
  "lucky_match": "You got a matching pair, but you didn't hear both cards before. This is considered a lucky match. You get 10 points.",
  ...
}
```

When the frontend displays the score, it also checks whether a message for the key for the given condition is in the tutorial dict. If so, it pops it and displays the message in an overlay

If we go this route, styling the overlay will also need attention. I.e., contrast to the background? Grey out background? Perhaps the `Feedback` component's styling could be partly reused?

No alternatives come to mind just yet, should discuss with the team. Ideally, we would arrive at a solution which may also get reused in other experiments, and is not purely engineered against MatchingPairs.

**Additional context**

drikusroor:
- Introduce a tutorial overlay for the MatchingPairs game, allowing users to receive guidance. 
- Refactor MatchingPairs to avoid prop mutation.
- Update the Button component to support multiple clicks when specified (so the tutorial overlay can be closed multiple times).
- Integrate MSW for API mocking in Storybook. 
- Add relevant tests for the new features.

Resolves #1338

### DIFF
--- a/backend/experiment/actions/playback.py
+++ b/backend/experiment/actions/playback.py
@@ -5,47 +5,47 @@ from .base_action import BaseAction
 from section.validators import audio_extensions
 
 # player types
-TYPE_AUTOPLAY = 'AUTOPLAY'
-TYPE_BUTTON = 'BUTTON'
-TYPE_IMAGE = 'IMAGE'
-TYPE_MULTIPLAYER = 'MULTIPLAYER'
-TYPE_MATCHINGPAIRS = 'MATCHINGPAIRS'
+TYPE_AUTOPLAY = "AUTOPLAY"
+TYPE_BUTTON = "BUTTON"
+TYPE_IMAGE = "IMAGE"
+TYPE_MULTIPLAYER = "MULTIPLAYER"
+TYPE_MATCHINGPAIRS = "MATCHINGPAIRS"
 
 # playback methods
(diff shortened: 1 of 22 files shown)


drikusroor:
> This works and looks really wel. Two things come to mind:
> 
>     * We should turn off the tutorial in `matching_pairs_lite`

I turned it off for all matching pairs rules files, and created a new `MatchingPairs2025` (which we from now on can regard as the working version of matching pairs 2.0) rules file that does have tutorial overlay text messages.

>     * We now log a timestamp for every action (click) in the game, so maybe we should also log if a turn showed the tuturial popup, because there's going to be some extra reading time.

I looked into this but this would be kind of hard to manage from the frontend (assuming we'd send this information from the `intermediate_score` endpoint) since we use the response of that api endpoint to (not) show a tutorial message in an overlay. 🤔

PR #1420, opened 2024-12-12
Docs: Update client references, add MkDocs build, and refactor social media serialization

drikusroor:
Update the client component reference from `Consent.js` to `Consent.tsx`, introduce a MkDocs build script, and refactor social media serialization with a new `SocialMediaConfigDto` TypedDict. Clean up unused imports and enhance documentation with added docstrings.

I might have gone overboard a bit with the amount of documentation, so feel free to tell me where to cut down on. I do like the newly introduced Dto typing classes though, although I can imagine them to be in a separate file or something, just like we do in the frontend in the types directory. Lmk what you think. :-)

### DIFF
--- a/backend/experiment/actions/final.py
+++ b/backend/experiment/actions/final.py
@@ -1,6 +1,5 @@
 from django.utils.translation import gettext_lazy as _
-from dataclasses import dataclass
-from typing import Optional, Dict
+from typing import Optional, Dict, TypedDict
 
 from experiment.serializers import serialize_social_media_config, SocialMediaConfigResponse
 from session.models import Session
@@ -8,8 +7,7 @@ from session.models import Session
 from .base_action import BaseAction
 
 
-@dataclass
-class ButtonResponse:
+class ButtonResponse(TypedDict):
     """
     Button configuration for an optional call-to-action button.
 
@@ -22,8 +20,7 @@ class ButtonResponse:
     link: str
 
 
-@dataclass
-class LogoResponse:
+class LogoResponse(TypedDict):
     """
     Logo configuration for branding or visual identification on the final screen.
 
@@ -36,8 +33,7 @@ class LogoResponse:
     link: str
 
 
-@dataclass
-class FinalActionResponse:
+class FinalActionResponse(TypedDict):
     """
     FinalActionResponse represents the structure of the final action data.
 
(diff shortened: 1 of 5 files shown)


backend/experiment/actions/final.py
```diff
@@ -1,45 +1,174 @@
...
+from typing import TypedDict
 
-from experiment.serializers import serialize_social_media_config
+from experiment.serializers import serialize_social_media_config, SocialMediaConfigDto
 from session.models import Session
 
 from .base_action import BaseAction
 
 
+class ButtonDto(TypedDict):
```
BeritJanssen, line 10: What does `Dto` stand for?

BeritJanssen: Ah, I see, "data transfer object". Judging from extensive sampling of programmers (n=1) who have no idea what this is about, perhaps we can find other terminology? "Serialization" or "Response" perhaps?

BeritJanssen: On a side note: while it's definitely a good idea to add types to this, I'm not sure if docstrings are really necessary here. After all, users of the interface will not directly call the functions used for serialization.

backend/experiment/actions/final.py
```diff
@@ -1,45 +1,174 @@
...
+        show_participant_link (bool): If True, display a participant-related link or information.
+        show_participant_id_only (bool): If True, only the participant ID is shown, without a link.
+        feedback_info (dict): Optional dictionary containing feedback-related data. For example:
+                              {"header": "Feedback", "prompt": "Tell us what you think", "button_text": "Submit"}.
+        total_score (float): Explicit final score. If None, this is derived from the session.
+        logo (LogoDto): Optional logo configuration for branding. For example:
+                        {"image": "/static/logo.png", "link": "https://example.com"}.
 
-    Relates to client component: Final.js
+    Attributes:
```
BeritJanssen, line 110: Why specify `Args` *and* `Attributes`?

backend/experiment/actions/final.py
```diff
@@ -1,45 +1,174 @@
...
-        total_score: float = None,
-        logo: dict = None,
+        feedback_info: dict | None = None,
+        total_score: float | None = None,
+        logo: LogoDto | None = None,
     ):
+        """
+        Initialize the final action data.
 
+        Args:
```
BeritJanssen, line 158: No, don't repeat this again in the `__init__` method, I would say. It's sufficient to add the arguments to the class, perhaps add an example implementation there. It would clutter the documentation page if we specify the arguments in two (or even three!) places.

backend/experiment/actions/final.py
```diff
@@ -82,12 +218,17 @@ def action(self):
             "logo": self.logo,
         }
 
-    def get_social_media_config(self, session: Session) -> dict:
+    def get_social_media_config(self, session: Session) -> SocialMediaConfigDto | None:
```
BeritJanssen, line 221: Again, this is a function that's *internally* called after initializing this action, and might actually confuse researcher-users of the infrastructure.

BeritJanssen (changes requested):
Thanks for adding docstrings! I would argue that docstrings for actions are necessary on the top level (i.e., telling a user "how do I initialize this action?"), and not adding doc strings for any functions or options that are used by the action internally. The latter would be certainly useful documentation for developers, but I wouldn't like to confuse researchers who want to work with the interface by including docstrings which aren't strictly relevant for them. *Or*, perhaps we can exclude some of the docstrings from mkdocs? I'd say that mkdocs is for researchers wanting to use the interface, and developers are much more likely view the docstrings through their IDE.

backend/experiment/actions/final.py
drikusroor: I went full monty as I didn't know what the documentation conventions are for our project and the fact that Args & Attributes are not the same. But I will remove Attributes now

backend/experiment/actions/final.py
drikusroor: > Ah, I see, "data transfer object". Judging from extensive sampling of programmers (n=1) who have no idea what this is about, perhaps we can find other terminology? "Serialization" or "Response" perhaps?

It's a fairly common concept in the software industry afaik, perhaps not in de python universe? I've called them `...Configuration` and `...Response` now, depending on the context. The `FinalResponse` is really a response, the other classes are both used in the constructor as well as the response.

> On a side note: while it's definitely a good idea to add types to this, I'm not sure if docstrings are really necessary here. After all, users of the interface will not directly call the functions used for serialization.

Related to what I wrote here above, the `ButtonConfiguration` and the `LogoConfiguration` types are used in the constructor of the final action, so it might make sense to show these too in the docs. What do you think?

PR #1422, opened 2024-12-12
Docs: Add documentation for Redirect action and update .gitignore

drikusroor:
Include a docstring for the Redirect action class to clarify its purpose and usage. Also, add `docs-pages-build` to .gitignore to prevent a locally generated mkdocs site to clutter your changes in git.

### DIFF
--- a/backend/experiment/actions/redirect.py
+++ b/backend/experiment/actions/redirect.py
@@ -7,10 +7,6 @@ class Redirect(BaseAction):
 
     This action is used to redirect the user to a specified URL.
 
-    Attributes:
-        ID (str): Identifier for the action, set to 'REDIRECT'.
-        url (str): The URL to which the redirection should occur.
-
     Args:
         url (str): The URL to redirect to.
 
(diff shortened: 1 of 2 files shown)


backend/experiment/actions/redirect.py
```diff
@@ -2,7 +2,23 @@
...
+        url (str): The URL to which the redirection should occur.
+
+    Args:
+        url (str): The URL to redirect to.
+
+    Example:
+        redirect_action = Redirect('https://example.com')
+    """
+
+    ID = "REDIRECT"
```
BeritJanssen, line 21: My vision here was that we'd document only the arguments which are relevant for initialization. Perhaps it's strange to not add this to the `__init__` method instead, but for a researcher who wants to initialize a `Redirect` action, the `Args` / `Example` information is what is relevant.

BeritJanssen (changes requested):
I wouldn't include the `ID` in the docstring, as it's not relevant for initialization.

backend/experiment/actions/redirect.py
drikusroor: But that's what I did, right? Or do you mean that I should remove the Attributes section?

PR #1423, opened 2024-12-12
Docs: Document HTML action class

drikusroor:
Improved the docstring for the HTML action class and its initializer to enhance clarity and provide better attribute and argument descriptions.


BeritJanssen (changes requested):
See my other reviews. Better to document `Args`, and then to include only what needs to be passed for initialization of the `HTML` object. Perhaps also add an `Examples` section here?

PR #1427, opened 2024-12-13
Add tabbed TranslatedContent form

### LINKED ISSUE #1321: Customize display of `TranslatedContent` objects in admin interface
Arguably, editing the `TranslatedContent` objects through the admin interface is not very user-friendly at the moment. It would be nice to display these kinds of objects just as a row of fields and an associated language, without reflecting the ForeignKey relationship visually.

### LINKED ISSUE #1373: Experiment admin: move ThemeConfig & active fields down
Arguably, the most important content for the Experiment page are phases and blocks - we should move ThemeConfig and "active" down, so they appear at the end of the form.

Potentially, it would also be worth adding the `TranslatedContent` stuff *after* the Phases/Blocks form?
Screenshot 2024-11-18 at 08 52 45

### LINKED ISSUE #1399: 🐛 [BUG] - Block translated content causes the page to overflow horizontally on the experiment admin form page
### Description

The block translated content's inline form causes the page to overflow horizontally on the experiment admin form page. If you collapse all the block translated content inline forms on the page, the overflow disappears.

A possible solution is to make the individual inline forms have their own overflow with scroll. If we can make the whole inline form more compact horizontally, that would probably help a bit as well.

Related to #1321

### Overflow

image

### Collapsed block translated content overflows not

image

### Reproduction URL

_No response_

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

### Logs

_No response_

### Browsers

_No response_

### OS

_No response_

### Additional context

_No response_

BeritJanssen:
Close #1321; close #1399; close #1373.

This branch adds a custom template for `TranslatedContent` models, which makes sure we can add/remove translations inline, using tabs. It also adds a custom template for `QuestionSeries` models, which shows a "add default questions" button. It is far from elegant that the forms are now modified copy & paste versions of those defined in django and django-nested-admin, but at this moment, these templates don't have `{% block %}` tags which would make it possible to extend and modify the existing templates more efficiently. (For the future: we might suggest PRs to the django / django-nested-admin maintainers to add these block tags.)

In leu of floating submit buttons (which mask part of the form) and custom expand / collapse of the phases form, this branch makes use of the options of Django to show submit buttons at the top of the form, and to add the `collapse` class to inline formsets.

At this moment, the form is not dynamic, i.e., language changes will only be picked up after saving. At the same time, this branch covers the desired changes for the experiment form in a basic way, piggy-backing Django (libraries) and therefore getting error / validation handling and serialization of objects mostly for free.


BeritJanssen:
NB: I've also experimentally disabled the `initCollapsibleInlineForms` in `experiment_form.js`, and instead added the `['wide', 'collapse']` classes to the Inline classes. Let's compare what we like best after merging. I do think the custom solution looks nicer, but perhaps we can achieve this with styling rather than javascript?

drikusroor:
Just a quick comment on what I see directly in the video, not code-related per se:
- The language is a bit hard to read on a (dark) blue background. Perhaps you can consider making the text white (with a text-shadow)
- The tabs' behavior isn't very dynamic. You have to save the experiment before you see the new sorting order or a newly added language. Would it be an idea to add some javascript to show (at least) the newly added language? Having the sorting order or the current language code of the tab in question be dynamic seems of less importance to me.

BeritJanssen:
I implemented the suggested style changes, and also went down deeper in the form to control the behaviour of `BlockTranslatedContent`: instead of being able to add these independently of the `ExperimentTranslatedContent`, these will now always be set and prefilled with values from the `ExperimentTranslatedContent` objects of the `Experiment` parent.  I decided to leave any dynamic behaviour for later: let's see first if users find the form usable enough without it.

Changing the order dynamically would be, by my estimate, easier than creating an extra tab with a selected language on the fly, which requires dealing with several forms which do not correspond to a database object. Requiring the user to save first before seeing a new `ExperimentTranslatedContent` object reflected as a tab may then just be the lesser evil, rather than potentially creating many forms on the fly, some of which won't be valid if the user only selected a language, but entered no text for the required `Name` field. With the tab / hidden forms setup, it would be hard to find back which of the translated content forms throws the validation error.

backend/experiment/models.py
```diff
@@ -462,10 +457,6 @@ def get_fallback_content(self) -> "BlockTranslatedContent":
         Returns:
             Fallback content
         """
-
-        if not self.phase or self.phase.experiment:
-            return self.translated_contents.first()
```
BeritJanssen, line 467: We don't have blocks which aren't assigned to phases, and no phases which aren't assigned to experiments anymore. If that happens, I think it's better to throw an error.

PR #1431, opened 2024-12-17
Fix er/song duration configuration

Evert-R:
This PR:
- Adds a check to `_update_sections` and `edit_sections` that prevents users from entering a greater duration for an audio file than the actual duration.

- Refactors the `_update_sections` function:
When the `csv.DictReader` would find an error in a row it would return with an error message, but would only jump out of the `for` loop not out of the function, so that the sections it had already processed were still created, while the rest ignored. 
Now all the errors are first gathered and then shown in a list when returned to the admin
Row length errors as well as rows with invalid numbers will be shown in the admin, but will not be processed. 
Exceeding song duration errors will be shown in the admin but the setting will be corrected and saved.

While running tests the duration check will not be executed, because of the absence of actual audio files

### DIFF
--- a/backend/section/models.py
+++ b/backend/section/models.py
@@ -20,6 +20,7 @@ import datetime
 import random
 import csv
 from os.path import join
+import audioread
 
 from django.db import models
 from django.utils import timezone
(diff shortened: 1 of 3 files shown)


backend/section/models.py
```diff
@@ -137,76 +139,103 @@ def is_number(string):
...
+
+            if "test" not in sys.argv:
+                # while running tests this would throw an error
+                with audioread.audio_open(file_path) as f:
+                    actual_duration = f.duration
+                if float(row['duration']) > actual_duration:
+                    # Add or edit this row, but show an error message containing the actual saved duration
+                    row['duration'] = actual_duration
+                    global_errors += 1
+                    csv_messages.append(f"Error: The duration of {row['filename']} exceeds the actual duration of the audio file and has been set to {actual_duration} seconds.")
```
drikusroor, line 174: Was gonna ask about this situation, very nice

drikusroor (approved):
LGTM

PR #1432, opened 2024-12-17
Draft: Implement prototype of analytics dashboard that can be expanded upon

drikusroor:
Introduce a prototype of an analytics dashboard using TypeScript and Chart.js, along with a new API endpoint for exporting block data in the admin interface that can be used to build the real analytics dashboard.

As this is a prototype, it is also an invitation to criticise this setup and / or ask questions about things that aren't clear yet.

## How to get this to work?

For the typescript project that contains the scripts for the charts, check https://github.com/Amsterdam-Music-Lab/MUSCLE/blob/76f5440707907f867bc4519f94fb3aa066f9a4b9/backend/experiment/static/analytics/README.md

The transpiled Javascript files are committed into the repository so they should work "out of the box". See also: https://github.com/Amsterdam-Music-Lab/MUSCLE/blob/76f5440707907f867bc4519f94fb3aa066f9a4b9/backend/experiment/static/analytics

Navigate to `http://localhost:8000/admin/experiment/analytics/block/{id}` to see the dashboard of a block in question. Be sure to use a block id of a block that has at least _some_ sessions and results.

### DIFF
--- a/backend/experiment/admin.py
+++ b/backend/experiment/admin.py
@@ -10,6 +10,7 @@ from django.core import serializers
 from django.shortcuts import render, redirect
 from django.forms import CheckboxSelectMultiple
 from django.http import HttpResponse
+from django.urls import path
 
 from inline_actions.admin import InlineActionsModelAdminMixin
 from django.urls import reverse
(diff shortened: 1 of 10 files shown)


drikusroor:
@Solcany I've assigned it to you since you are working on it now (or later) :-)

Issue #1417, opened 2024-12-10
🐛 [BUG] - The playlist select input appears in a weird (top-left of page) position for blocks on a new experiment


drikusroor:
### Description

My first hunch is that its container needs a `position: relative` so that the absolute position of the input works correctly. But it might also be a `django_select2` issue.

### Reproduction steps

```bash
1. Go to '...'
2. Click on '....'
3. Scroll down to '....'
4. See error
```

BeritJanssen:
Btw, the select widget also doesn't have a proper dark mode: while the background of the select element stays white, the font becomes lighter, which makes it unreadable.

PR #1434, opened 2024-12-17
Initialize playlist input for new block

Evert-R:
This PR initializes the playlist input `django-select2` when creating a new block, so that it appears in the right place.

closes: #1417


drikusroor:
Oh, that was easier than expected it seems :D

Issue #1428, opened 2024-12-16
Code documentation todo list


Evert-R:
#### - actions

- [x] experiment.actions.base_action #1444

- [x] experiment.actions.consent #1445

- [x] experiment.actions.explainer #1430 

- [x] experiment.actions.final #1420

- [x] experiment.actions.form #1446

- [x] experiment.actions.frontend_style #1447

- [x] experiment.actions.html #1423

- [x] experiment.actions.info #1438

- [x] experiment.actions.playback #1448

- [x] experiment.actions.redirect #1422

- [x] experiment.actions.score #1438

- [x] experiment.actions.styles #1450

- [x] experiment.actions.trial #1451 

- [x] experiment.actions.utils #1452

- [x] experiment.actions.wrappers #1453

#### - experiment app

- [x] experiment.models

- [ ] experiment.rules

#### - image app

- [x] #1469

#### - participant app

- [x] #1477

- [x] #1483

#### - question app

- [x] #1485

- [x] #1486

#### - result app

- [x] #1457

#### - section app

- [x] section.models

#### - session app

- [x] session.models

#### - theme app

- [x] #1484

PR #1446, opened 2024-12-24
Docs: Improve documentation and typing for `Form` and `Question` classes

drikusroor:
Enhance the documentation and typing for the `Form` action class and all `Question` classes.

Related to #1428

### DIFF
--- a/backend/experiment/actions/form.py
+++ b/backend/experiment/actions/form.py
@@ -1,3 +1,4 @@
+from typing import Dict, List, Union, Optional, Any, Literal
 from django.conf import settings
 from django.utils.translation import gettext_lazy as _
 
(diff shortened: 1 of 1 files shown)


backend/experiment/actions/form.py
```diff
@@ -38,91 +62,280 @@ def __init__(
...
+        choices (Optional[Dict[str, str]]): Custom yes/no labels
+        **kwargs: Additional Question arguments
+
+    Example:
+        ```python
+        question = BooleanQuestion(
+            key="is_student",
+            question="Are you a student?",
+            choices={
+                "no": "No", # or _("No") for translation
```
BeritJanssen, line 122: Perhaps mention that `_("No")` is the default? And perhaps use "Nope" and "Yep" as examples, to make sure that users won't think it's necessary to define choices if they are happy with (translatable) yes/no.

backend/experiment/actions/form.py
```diff
@@ -38,91 +62,280 @@ def __init__(
...
+
+    def __init__(self, choices: Dict[str, str], min_values: int = 1, **kwargs) -> None:
         super().__init__(**kwargs)
         self.choices = choices
         # minimal number of values to be selected, 1 or more
+        # TODO: enforce (raise ValueError), or math.floor it
         self.min_values = min_values
 
 
 class DropdownQuestion(Question):
```
BeritJanssen, line 170: Should these all inherit `ChoiceQuestion`, rather than plan `Question`? `BooleanQuestion` always comes with yes/no buttons at this point.

backend/experiment/actions/form.py
```diff
@@ -151,41 +364,82 @@ def __init__(self, scale_steps=7, explainer=_("How much do you agree or disagree
 
 
 class LikertQuestionIcon(Question):
```
BeritJanssen, line 366: Perhaps this can also inherit from `LikertQuestion` instead?

BeritJanssen (approved):
Nice! Only wonder whether we can make the relationships between different question types clearer by more specific inheritance. No precondition for merging though.

PR #1447, opened 2024-12-24
Docs: Enhance documentation for the frontend style classes

drikusroor:
Improve type annotations and documentation for frontend style classes.

Related to #1428


BeritJanssen (commented):
LGTM. To my knowledge, we're actually not using the `FrontendStyle` class always as it's documented though. We're not using the `root_style` option anywhere. I think this is something that should still be streamlined somewhat, but that can be left for later.

PR #1449, opened 2024-12-27
Fixed: Fix deletion of blocks related to translated content

### LINKED ISSUE #1436: 🐛 [BUG] - Cannot delete blocks after having created `BlockTranslatedContent`
### Description

It seems that something goes wrong in the order objects are being deleted when checking "delete" on a form from the `BlockInline` admin. After saving, we get a juicy IntegrityError. After some hours of debugging, I arrived at the conclusion that this is probably a bug in the `django-nested-admin` package, which may just not foresee the level of nesting we're doing here (`Experiment`->`Phase`->`Block`->`BlockTranslatedContent`).

### Screenshots
Screenshot 2024-12-18 at 14 56 04
Screenshot 2024-12-18 at 14 56 19

### Reproduction URL

tst.amsterdammusiclab.nl/server/admin

### Reproduction steps

```bash
1. Log in to the test or your local admin interface
2. Click on 'Experiments'
3. Edit an experiment: go to blocks and mark a block with a `BlockTranslatedContent` for deletion
4. Click "Save"
```

### Logs

_No response_

### Browsers

_No response_

### OS

_No response_

### Additional context

_No response_

drikusroor:
Address an issue preventing the deletion of blocks due to a block without a primary key being unable to relate to translated content.

Fixes #1436

### DIFF
--- a/backend/experiment/models.py
+++ b/backend/experiment/models.py
@@ -227,6 +227,12 @@ class Block(models.Model):
     theme_config = models.ForeignKey(ThemeConfig, on_delete=models.SET_NULL, blank=True, null=True)
 
     def __str__(self):
+        # If this block is unsaved or being deleted (has no PK),
+        # avoid calling get_fallback_content() (which does a DB query).
+        if not self.pk:
+            # Provide a fallback label or just return self.slug if present.
+            return self.slug or "Deleted/Unsaved Block"
+
         content = self.get_fallback_content()
         return content.name if content and content.name else self.slug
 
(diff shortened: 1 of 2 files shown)


BeritJanssen (approved):
Good catch! Perhaps a unit test to confirm saving / deleting blocks works would be a good idea?

drikusroor:
> Good catch! Perhaps a unit test to confirm saving / deleting blocks works would be a good idea?

I wasn't able to reproduce it in an admin test oddly enough 🤔 So I made a test for the __str__ function of the Block model

PR #1450, opened 2024-12-27
Docs: Improve documentation and enhance Trial story UX

drikusroor:
Add docstrings for style constants in `styles.py` for better clarity. These don't seem to show up in mkdocs though... Maybe because they aren't classes/functions?

Fix the initial display issue of the Trial story to improve user experience by ensuring it shows correctly without requiring a click.

Related to #1428

### DIFF
--- a/backend/experiment/actions/styles.py
+++ b/backend/experiment/actions/styles.py
@@ -1,8 +1,27 @@
-STYLE_NEUTRAL = 'neutral'
-STYLE_NEUTRAL_INVERTED = 'neutral-inverted'
-STYLE_BLUE = 'blue'
-STYLE_PINK = 'pink'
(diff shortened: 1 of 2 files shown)


BeritJanssen (commented):
Indeed, styles is something to be revisited. I see two approaches now:
- a dictionary with multiple `'class-name': True` combinations, which essentially means these classes will be slapped onto the class list of the corresponding frontend component's element. This is used in the Categorization rules.
- a single string, which is evaluated by `FrontendStyle` against a number of possible values.

I like the first approach better, as it's less involved. However, I also struggle with how to document the possible class names that could be set from the backend, and their effect on the frontend styling. It would be good that, if a user sets the class name to `neurtal`, the backend raises a warning to catch the typo.

PR #1455, opened 2025-01-06
Draft: Experiment form using React

drikusroor:


backend/experiment/views.py
```diff
@@ -14,13 +17,21 @@
...
 
 logger = logging.getLogger(__name__)
 
 
+class ExperimentViewSet(viewsets.ModelViewSet):
+    queryset = Experiment.objects.all()
+    serializer_class = ExperimentSerializer
+    permission_classes = [permissions.IsAuthenticated]
+
+
```
drikusroor, line 34: This creates GET, GET(id), POST, PUT & DELETE endpoints for the experiment (and its relations as defined in the serializers)

backend/section/views.py
```diff
@@ -51,11 +55,18 @@ def get_section(request: HttpRequest, section_id: int) -> Section:
...
 
     except Section.DoesNotExist:
         raise Http404("Section does not exist")
+
+
+@api_view(["GET"])
+def playlists(request):
+    """Return a list of all playlists"""
+    playlists = [{"id": playlist.id, "name": playlist.name} for playlist in Playlist.objects.all()]
+    return Response(playlists)
```
drikusroor, line 72: This is used by the experiment form in react to be able to show all playlists and pick one from a select input

backend/aml/urls.py
```diff
@@ -26,31 +32,32 @@
...
+    path("question/", include("question.urls")),
+    path("participant/", include("participant.urls")),
+    path("result/", include("result.urls")),
+    path("section/", include("section.urls")),
+    path("session/", include("session.urls")),
+    path("theme/", include("theme.urls")),
+    path("admin_interface/", include("admin_interface.urls")),
+    path("admin/", admin.site.urls),
+    path("api/token/", TokenObtainPairView.as_view(), name="token_obtain_pair"),
+    path("api/token/refresh/", TokenRefreshView.as_view(), name="token_refresh"),
```
drikusroor, line 45: This is used by the experiment form react application to obtain a JWT token that is used to authenticate the user in Rest API requests

backend/experiment/serializers.py
```diff
@@ -3,14 +3,138 @@
...
+
+class PhaseSerializer(serializers.ModelSerializer):
+    blocks = BlockSerializer(many=True, required=False)
+
+    class Meta:
+        model = Phase
+        fields = ["id", "index", "dashboard", "randomize", "blocks"]
+
+
+class ExperimentSerializer(serializers.ModelSerializer):
```
drikusroor, line 49: This is the weak part about this setup (for now) I think. It seems to be a bit convoluted but I do think it should be able to simplify it. It's like this right now because I had to deal with different scenarios like many to many relationships, or just FKs, do the related objects have PKs already or not, and so on. Maybe a good step is just to create and save an empty experiment before you get to the form. Then at least the experiment already exists and we can remove the `create` method in here.

backend/experiment/templates/form/experiment-form/index.html
drikusroor, line 1: If we load this file and the built JS + CSS files, I think we can integrate it into the Django admin interface quite easily and maybe even remove the whole JWT mumbo jumbo.

BeritJanssen:
My preference would be to handle this form through the *frontend*, so we don't have a (potentially confusing) React section in the backend. To my knowledge, authentication should still work through Django session cookies. This has the drawback of not using tailwind, but I think it's going to make the codebase more organized.

Issue #214, opened 2022-08-30
Publish AML on eScience Research Software Directory


BeritJanssen:
There is an initiative of the eScience center, in collaboration with SURF, to provide a [Research Software Directory](https://research-software-directory.org/), in which universities and other research institutes can showcase their software. It would be good to publish AML there also.

Issue #212, opened 2022-08-30
Publish AML on CLARIAH/ODISSEI


BeritJanssen:
Some insights from the Clariah Tech Day on March 30th:
### Publishing software
We can register software, once our repo is open, via source registry in Clariah's tool-discovery:
[Github](https://github.com/CLARIAH/tool-discovery)
Repos Harvested & converted to linked open data
[Requirements for resuse](https://github.com/CLARIAH/clariah-plus/blob/main/requirements/software-requirements.md)
[Overview of tools](https://tools.clariah.nl/)

### Publishing data
[Harvard Dataverse](https://data.harvard.edu/dataverse) seems to have become the standard for storing data. DANS maintains a national hub for Dataverse. Linked open data is preferred, but requires an onthology, reusing existing vocabularies.
For presenting data and research results, [Clariah Data stories](https://stories.datalegend.net/) may also be interesting: these provide Jupyter notebooks along with some explanatory text. The Jupyter notebooks could also be a useful resource to share analysis scripts with colleagues and students.

PR #1471, opened 2025-01-20
add CITATION.cff file

BeritJanssen:
This file will be used by #212 and #214 to extract information about the software. @jaburgoyne , can you give feedback on the abstract and keywords?


jaburgoyne:
My instinct: Author names using current developers in order of hiring, followed by Werner, followed by Henkjan and me (will check on order between the two of us and get back to you). If possible, it might be nice to add the ICMPC citation to the README file (or elsewhere in this CFF file if the format supports it).

jaburgoyne (changes requested):
See my comment outside the review

Issue #1335, opened 2024-10-31
MatchingPairs: improve feedback after clearing the board


BeritJanssen:
## Requirements

- [ ] Update `Final` component, don't create new component
- [ ] Add rank / percentile bar
- [ ] Keep text color the same (don't make it dependent on cup color)
- [ ] Should we replace old gold color with the gold color from the design?
- [ ] Show the new score informations not as separate HTML, but use an `HTML` component to display `feedback_info` and format it in the new matching pairs rules file with the right information.

## Original description & mockup

The "Final" page after clearing a board in MatchingPairs should provide additional information to participants:
- [ ] how hard was the board? Are the upcoming boards more or less challenging than the current one?
- [ ] how did participants perform, compared to other participants?
- [ ] what was their highest score so far?
- [ ] what was their average score?

![MockupFinal](https://github.com/user-attachments/assets/539e67ea-55b5-4125-9172-df9806fd332e)

BeritJanssen:
To display how participants performed compared to others, we won't add a visual display just yet, but say "you perfomed X percent of players". This message should be suppressed if the participant's percentile is below 25%.

BeritJanssen:
Ordering:
- This game
- Personal best
- Average score
- Total score (i.e., over all sessions)

BeritJanssen:
To implement: redesign MatchingPairs rules so that it's one board per session, then we can organize the successive boards as blocks attached to one experiment.

jaburgoyne:
See also #518

PR #1472, opened 2025-01-20
Added: Add FinalV2 component with enhanced feedback and ranking features

drikusroor:
Introduce the FinalV2 component, which includes additional feedback information, a rank bar section, and new styling elements. Update the color scheme with a new goldmp2 variable for improved visual representation.

Related to #1335

### DIFF
--- a/backend/experiment/actions/final.py
+++ b/backend/experiment/actions/final.py
@@ -1,6 +1,7 @@
 from django.utils.translation import gettext_lazy as _
 from typing import Optional, Dict, TypedDict
 
+from experiment.actions.types import FeedbackInfo
 from experiment.serializers import serialize_social_media_config, SocialMediaConfigConfiguration
 from session.models import Session
 
@@ -36,6 +37,7 @@ class LogoConfiguration(TypedDict):
 class FinalActionResponse(TypedDict):
     view: str
     score: float
+    percentile: Optional[float]
     rank: Optional[str]
     final_text: Optional[str]
     button: Optional[ButtonConfiguration]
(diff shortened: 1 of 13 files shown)


frontend/src/components/Final/Final.tsx
```diff
@@ -49,10 +51,13 @@ const Final = ({
         <div className="aha__final d-flex flex-column justify-content-center">
             {rank && (
                 <div className="text-center">
-                    <Rank cup={{ className: rank.class, text: rank.text }} score={{ score, label: points }} />
+                    <Rank cup={{ className: rank?.class, text: rank.text }} score={{ score, label: points }} />
                 </div>
             )}
-            <div className="text-center">
+            <div className={classNames("aha__final-rank-bar-section", rank?.class)}>
```
BeritJanssen, line 57: Hmm, wouldn't this mean that we'd get the percentile bar in every experiment from now on? I think it would be best to be able to configure whether or not to show the rank bar cursor.

backend/experiment/actions/final.py
```diff
@@ -36,6 +37,7 @@ class LogoConfiguration(TypedDict):
 class FinalActionResponse(TypedDict):
     view: str
     score: float
+    percentile: float
```
BeritJanssen, line 40: I think `percentile` should be optional (it's not going to be applicable in every scenario and/or not every researcher might want to show it to the participants).

BeritJanssen (changes requested):
Nice! It would be nice though if you could still tweak the `Final` component such that the display of the percentile can be set, so it's not going to effect all experiments, only MatchingPairs for now. Perhaps this can be achieved by (not) setting `percentile` when creating the `Final` action, and let displaying in the frontend depend on that?

docker-compose.yaml
```diff
@@ -1,7 +1,7 @@
 # Docker-compose definition for development
 services:
   db:
-    image: postgres:14
+    image: postgres:16
```
BeritJanssen, line 4: 😢  Maybe let's not? Or what would be the arguments for changing the version again?

drikusroor: Oh, that was an accidental commit because I have to do this every time to make my local setup work 😅

frontend/src/components/Final/Final.tsx
drikusroor: Agree

PR #1474, opened 2025-01-20
Fix/multiple default questions

BeritJanssen:
This branch fixes the problem of not being able to add default questions for any blocks than the first in the `ExperimentForm`, and also reintroduces the floating submit row (together with a `margin-bottom` on the last field in the form, so the buttons don't obscure the form), as well as the custom collapsible forms. Both Django built-in options (i.e., putting `save_on_top=True` in the admin, and adding `classes=['collapse']`, are as easy to use as the custom functions. `collapse` will always show the form as collapsed by default, with now options to customize that, other than writing custom Javascript code -- so the original goal I had in mind with using the Django tweaking options, i.e., avoid custom Javascript code, is kind of defunct.


drikusroor:
I'm writing some remarks here for things that I noticed while giving the experiment form a quick test. Not all of them might have something to do with this PR per se but we can create separate issues out of those.

image

- The Playlists input (still / again) does not work for newly added unsaved Blocks.
- I can (somehow) type in the _Add default questions_ button (see screenshot) 😅 

image

When I try to save a new question series + question. It tells me that the index inputs are required. For ease of use I would make them either _not_ required (and have them default to 0 in the save method) or make the inputs have a default value of 0 (if possible).

drikusroor (approved):
Code LGTM but see my comment here for some things I noticed during a quick test: https://github.com/Amsterdam-Music-Lab/MUSCLE/pull/1474#issuecomment-2604043210

PR #1478, opened 2025-01-21
Feature/document result

### LINKED ISSUE #1457: Document result.models

BeritJanssen:
Close #1457 . The branch also adds docstrings to the `score`, `views` and `utils` modules, and increases test coverage.

### DIFF
--- a/backend/result/models.py
+++ b/backend/result/models.py
@@ -1,18 +1,38 @@
-import json
-
 from django.db import models
 from django.utils import timezone
 
 from django.core.exceptions import ValidationError
 
 
-# Create your models here.
 class Result(models.Model):
-    """Score for each step in a session"""
-    session = models.ForeignKey('session.Session', on_delete=models.CASCADE, blank=True, null=True)
-    participant = models.ForeignKey('participant.Participant', on_delete=models.CASCADE, blank=True, null=True)
+    """A model to register participant responses to the database.
+
+    Note that a Result can be either tied to a Session, which is the common way to log responses to tasks during an experiment,
+    or to a Participant, which can be a consent form or a demographic question.
+
+    Attributes:
+        session (Optional[Session]): session for which this result will be registered
+        participant (Optional[Participant]): participant for which this result will be registered
+        section (Optional[Section]): a section tied to the result, usually applicable for Trials with Playback
+        created_at (datetime): a timestamp, set automatically at creation time
+        question_key (str): a description by which to identify the result during analysis
+        expected_response (str): if there is a correct response for a given Trial, it can be logged here
+        given_response (str): set as a result of the participant's response
+        comment (str): optional comment to help analysis
+        score (float): numerical value to reflect the participant's success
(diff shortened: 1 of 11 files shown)


backend/result/views.py
```diff
@@ -22,26 +43,43 @@
...
         # Create a result from the data
-        result = handle_results(result_data, session)
-        if not result:
-            return HttpResponseServerError("Could not create result from data")
-    except ValueError:
-        return HttpResponseServerError("Invalid data")
-
+        handle_results(result_data, session)
+    except Exception as e:
+        return HttpResponseServerError(f"Invalid data: {e}")
```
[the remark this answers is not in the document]
drikusroor: 😵 That's pretty neat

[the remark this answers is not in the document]
BeritJanssen: Yup, I fixed this. Any other comments before I merge?

backend/result/models.py
```diff
@@ -1,18 +1,38 @@
...
 
-# Create your models here.
 class Result(models.Model):
-    """Score for each step in a session"""
-    session = models.ForeignKey('session.Session', on_delete=models.CASCADE, blank=True, null=True)
-    participant = models.ForeignKey('participant.Participant', on_delete=models.CASCADE, blank=True, null=True)
+    """A model to register participant responses to the database.
+
+    Note that a Result can be either tied to a Session, which is the common way to log responses to tasks during an experiment,
+    or to a Participant, which can be a consent form or a demographic question.
```
drikusroor, line 11: mkdocs does have a

```mkdocs
Note:
    Lorem ipsum
```

syntax. Would it make sense to use that here too or would you prefer it to be shown on top?

backend/result/score.py
```diff
@@ -1,74 +1,105 @@
...
+    After continuation of audio, participants need to decide whether it was continued in the correct place.
+    If answered incorrectly, this function modifies the reaction time score of the previous step.
+    """
     previous_result = result.session.last_result(["recognize"])
-    if check_expected_response(result) != result.given_response:
-        previous_result.score *= -1
+    if result.expected_response != result.given_response:
+        previous_result.score *= (
+            -1
+        )  # will raise exception if `previous_result.score = None`
```
drikusroor, line 102: Should this be a TODO: as well then?

backend/result/utils.py
```diff
@@ -19,63 +34,103 @@ def get_result(session, data):
     return result
 
 
-def handle_results(data, session):
+def handle_results(data: dict, session: Session) -> bool:
```
drikusroor, line 37: Does this really return a bool? 🤔

drikusroor (approved):
LGTM with some comments here and there

backend/result/utils.py
BeritJanssen: Ah, whoops, not anymore indeed!

backend/result/score.py
BeritJanssen: Hm, perhaps it would be good to catch the exception and log a message about the special conditions for this scoring rule.

PR #1491, opened 2025-01-28
Fixed: Add `show_percentile` option to `Final` action & component

### LINKED ISSUE #1488: Do not show percentiles by default
https://github.com/Amsterdam-Music-Lab/MUSCLE/pull/1472 introduced the option to display the percentile of the participant, but this now shows everywhere (screenshot below from ToontjeHoger "Het eerste luisteren"). The `percentile` variable is now always set in the `Final` action, but this behaviour should be controllable when initializing a `Final` action.

Image

drikusroor:
Introduce a new option to display the participant's percentile rank in the Final action and component, along with enhanced tests to verify this functionality.

Fixes #1488

### DIFF
--- a/backend/experiment/actions/final.py
+++ b/backend/experiment/actions/final.py
@@ -47,6 +47,7 @@ class FinalActionResponse(TypedDict):
     social: Optional[SocialMediaConfigConfiguration]
     show_profile_link: bool
     show_participant_link: bool
+    show_percentile: bool
     feedback_info: Optional[Dict[str, str]]
     participant_id_only: bool
     logo: Optional[LogoConfiguration]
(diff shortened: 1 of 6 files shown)


backend/experiment/actions/final.py
```diff
@@ -154,13 +158,13 @@ def action(self) -> FinalActionResponse:
             "social": self.get_social_media_config(self.session),
             "show_profile_link": self.show_profile_link,
             "show_participant_link": self.show_participant_link,
+            "show_percentile": self.show_percentile,
             "feedback_info": self.feedback_info,
             "participant_id_only": self.show_participant_id_only,
             "logo": self.logo,
         }
 
         if self.percentile is None:
```
BeritJanssen, line 167: This will actually never be `None`, the way it's configured now?

backend/experiment/actions/final.py
```diff
@@ -122,6 +125,7 @@ def __init__(
         self.show_profile_link = show_profile_link
         self.show_participant_link = show_participant_link
         self.show_participant_id_only = show_participant_id_only
+        self.show_percentile = show_percentile
         self.feedback_info = feedback_info
         self.logo = logo
         self.percentile = session.percentile_rank(exclude_unfinished=False)
```
BeritJanssen, line 131: IMO it's not even necessary to pass the `show_percentile` boolean to the frontend. We could set `self.percentile` to None if this is `False`, right? Or are there scenarios in which we don't want to show the percentile, but still pass the value?

BeritJanssen (changes requested):
Nice! I do wonder whether we couldn't save a couple of lines of code by letting the presence of the `percentile` variable depend on the `show_percentile` setting, and check against that in the frontend, see my comments. If there are good reasons to send `percentile` even if `show_percentile=False`, go ahead and merge though.

PR #1493, opened 2025-01-29
Docs: Add documentation for UTC timestamps in JSON exports

### LINKED ISSUE #1071: Document that we enforce UTC for timestamps
So far as I can see, the timestamps in our JSON exports have always been and are still in the UTC time zone. For consistency, I fully support this choice!

I'm not sure, however, that we formally *enforce* this anywhere or have it documented. Could we do that?

drikusroor:
Include a note clarifying that all timestamps in JSON exports are in UTC and provide guidance on server timezone configuration.

Resolves #1071


Evert-R (approved):

jaburgoyne:
Thanks! I still find this a bit confusing: why is it relevant that the server is on Amsterdam time? What I was hoping for in the original issue is that an end-user would be working 100% with UTC (and *maybe* a customisable Django admin option to represent the underlying UTC in a different time zone of choice, for web display only)

drikusroor:
> Thanks! I still find this a bit confusing: why is it relevant that the server is on Amsterdam time? What I was hoping for in the original issue is that an end-user would be working 100% with UTC (and _maybe_ a customisable Django admin option to represent the underlying UTC in a different time zone of choice, for web display only)

The `TIME_ZONE` setting is used to render date times in templates and in the Django Admin dashboard, and when time gets serialized in Rest API endpoints. So it does not affect the way datetime objects are stored in the database.

Perhaps we could make it an environment variable setting though, to make it more flexible for other researchers that want to use MUSCLE?

jaburgoyne:
An environment variable would be ideal if it's still possible! I don't have a sense of how much work that is. If it's going to delay other things, then perhaps just note that the admin interface will show Amsterdam time but that all data exports will be in UTC

drikusroor:
> An environment variable would be ideal if it's still possible! I don't have a sense of how much work that is. If it's going to delay other things, then perhaps just note that the admin interface will show Amsterdam time but that all data exports will be in UTC

I've created issue #1497 for this

backend/docs/11_Exporting_result_data.md
```diff
@@ -377,3 +377,12 @@ You will be presented with a screen that lets you choose the fields that you wan
 ## The database structure
 [![muscle_db_visualized](https://github.com/user-attachments/assets/f0d97029-ec49-4bcc-8c84-8ed35b0937c0)](https://github.com/user-attachments/assets/f0d97029-ec49-4bcc-8c84-8ed35b0937c0)
 [*click to open the image to enable zoom functionality*](https://github.com/user-attachments/assets/f0d97029-ec49-4bcc-8c84-8ed35b0937c0)
+
+## Note on Timestamps
+
+All timestamps in JSON exports are in UTC time. This ensures consistency across different systems and time zones. Please also note that the currently configured timezone of the server is 'Europe/Amsterdam'. You can change these settings in the `base_settings.py` file in the `backend/aml` directory. Look for:
```
BeritJanssen, line 383: Hm, it may not be so clear to outside collaborators what "the server" is. Perhaps something like this is clearer?
```suggestion
All timestamps in JSON exports are in UTC time. This ensures consistency across different systems and time zones. Please also note that the currently configured timezone for this application is 'Europe/Amsterdam'. If you would like to fork and spin your own instance with another timezone setting, you can change the `base_settings.py` file in the `backend/aml` directory. Look for:
```

BeritJanssen (approved):
Thanks for this! May still reformulate the part about adapting timezones a bit.

EDIT: I see that I missed above discussion. Environment variable is indeed also a good solution. This avoids having to fork the repo.

PR #1494, opened 2025-01-30
Feature er/add participant model documentation

Evert-R:
This PR adds documentation, examples and tests for the `Participant` models and utils.

Closes: #1477, #1483


drikusroor:
> ## 🔍 Existing Issues For Review
> 
> Your pull request is modifying functions with the following pre-existing issues:
> 
> 📄 File: **backend/participant/utils.py**
> Function 	Unhandled Issue
> **`get_participant`** 	[**Participant.DoesNotExist: No participant in session**](https://uva-aml.sentry.io/issues/6098029958/?referrer=github-open-pr-bot) /server/experiment/{slu...
> `Event Count:` **1**
> 
> Did you find this useful? React with a 👍 or 👎

@Evert-R is this something we need to investigate perhaps?

backend/participant/models.py
```diff
@@ -63,12 +131,36 @@ def profile_object(self):
                     profile.question_key)] = profile.score
         return profile_object
 
-    def is_dutch(self):
-        """Return if participant is tagged with Netherlands (nl) country code"""
+    def is_dutch(self) -> bool:
+        """Return if a participant is tagged with Netherlands (nl) country code
```
drikusroor, line 135:
```suggestion
        """Return if a participant is tagged with the country code of the Netherlands (nl)
```

backend/participant/models.py
```diff
@@ -63,12 +131,36 @@ def profile_object(self):
...
+
+        Returns:
+            Wether a participant is located in the Netherlands or not.
```
drikusroor, line 138: Is this technically what it does? Do we decide this using the ip2country service or by the locale settings from the client's OS/Browser? In the latter case a Dutch person might live abroad with its laptop set to Dutch / NL.

backend/participant/tests.py
```diff
@@ -131,3 +132,102 @@ def test_score_sum(self):
...
+
+    def test_session_count(self):
+        Session.objects.bulk_create(
+            [
+                Session(block=self.block, participant=self.participant),
+                Session(block=self.block, participant=self.participant),
+                Session(block=self.block, participant=self.participant),
+            ]
+        )
+        self.assertEqual(self.participant.session_count(), 4)
```
drikusroor, line 144: Why is this 4 if you are only creating 3 sessions? 🤔

drikusroor (approved):
LGTM. Added some comments but nothing too serious

Evert-R:
> > ## 🔍 Existing Issues For Review
> > Your pull request is modifying functions with the following pre-existing issues:
> > 📄 File: **backend/participant/utils.py**
> > Function 	Unhandled Issue
> > **`get_participant`** 	[**Participant.DoesNotExist: No participant in session**](https://uva-aml.sentry.io/issues/6098029958/?referrer=github-open-pr-bot) /server/experiment/{slu...
> > `Event Count:` **1**
> > Did you find this useful? React with a 👍 or 👎
> 
> @Evert-R is this something we need to investigate perhaps?

I had a quick look at it, but I figure it's intentionally created that way. Maybe @BeritJanssen knows?

backend/participant/models.py
Evert-R: > Is this technically what it does? Do we decide this using the ip2country service or by the locale settings from the client's OS/Browser? In the latter case a Dutch person might live abroad with its laptop set to Dutch / NL.

We use ip2country for that. To restrict playing certain audio files licensed only for the Netherlands.

backend/participant/tests.py
Evert-R: There's already one in the SetupTestData ;-)

drikusroor: I thought about that, but I also think a test should communicate intended behavior and this might cause confusion :-D

Evert-R: That's a fair point. I've updated the test.

backend/participant/models.py
```diff
@@ -96,8 +188,21 @@ def scores_per_experiment(self):
 
         return scores
 
-    def score_sum(self, question_group_key):
-        """ Sums scores of all profile results with questions in a question group"""
+    def score_sum(self, question_group_key) -> float:
+        """ Sums scores of all profile results with questions in a question group
+
+        attributes:
```
albertas-jn, line 194: Here, and similarly in other methods, `question_group_key` is an argument, not an attribute.

albertas-jn (changes requested):
In methods, "attributes" should be changed to "Args", according to our own docstring guidelines. (It should be called "Parameters" though).

backend/participant/models.py
Evert-R: Missed that ! Thanks

PR #1495, opened 2025-01-30
Feature er/add theme model documentation

Evert-R:
This PR adds documentation for the `theme` models

closes: #1484


drikusroor (approved):
LGTM, but I was thinking if it would be better to specify the Args: instead of the Attributes:. They're mostly the same, but then you'd know how to initialize one. Not sure which one is more convenient. Also, it might be nice to add an example of initialization of every model.

Evert-R:
> LGTM, but I was thinking if it would be better to specify the Args: instead of the Attributes:. They're mostly the same, but then you'd know how to initialize one. Not sure which one is more convenient. Also, it might be nice to add an example of initialization of every model.

Examples are indeed always nice to have, but I figured that they are mostly useful for functions that researchers use in their rules files. Developers should know their way around a Django model, but i'll look into it ;-)

Attributes are used for classes and Args for functions. We use the google style with mkdocs.
More info here:
[https://amsterdam-music-lab.github.io/MUSCLE/this_doc/](https://amsterdam-music-lab.github.io/MUSCLE/this_doc/)

PR #1496, opened 2025-01-30
Added: Scaffold playlist selection for MP2.0

drikusroor:


backend/experiment/rules/matching_pairs_2025.py
```diff
@@ -55,15 +59,151 @@ def next_round(self, session):
...
                 rank=self.rank(session, exclude_unfinished=False),
                 feedback_info=feedback_info,
                 percentile=session.percentile_rank(exclude_unfinished=False),
             )
             return [score]
 
-    def final_text(self, session):
+    def get_matching_pairs_trial(self, session):
+        player_sections = self._select_sections(session)
+        random.seed(self.random_seed)
```
BeritJanssen, line 72: Is this for testing purposes?

drikusroor: No, I renamed the `final_text` method to `_final_text` and `git diff` got confused

backend/experiment/rules/matching_pairs_2025.py
```diff
@@ -50,26 +61,202 @@ def next_round(self, session):
...
+            button={"text": "Next game", "link": self.get_experiment_url(session)},
+            rank=self.rank(session, exclude_unfinished=False),
+            percentile=session.percentile_rank(exclude_unfinished=False),
+        )
+
+        return [score]
+
+    def get_matching_pairs_trial(self, session: Session):
+        player_sections = self._select_sections(session)
+        random.seed(self.random_seed)
```
BeritJanssen, line 89: My previous question was aiming at why you would set a ramdom seed here before shuffling. That was used for enforcing a fixed order, but I don't think it's needed here?

BeritJanssen (approved):
Don't forget to remove the change to `PlayCard.tsx`, as per the TODO, prior to merging.

Issue #1489, opened 2025-01-28
Move documentation from the wiki to mkdocs


Evert-R:
There are still topics in the wiki that should be moved to mkdocs

PR #1501, opened 2025-02-04
Feature er/move remaining wiki items to mkdocs

Evert-R:
This PR adds the remaining items from the wiki to the mkdocs documentation under a new section : `Developers documentation`

Closes: #1489


backend/docs/D4_Installing_python_packages.md
```diff
@@ -0,0 +1,28 @@
...
+* `requirements.in/dev.txt` for packages that are only needed in *development*
+
+* `requirements.in/prod.txt` for packages that are only needed in *production*
+
+Then, run the following `pip-compile` commands to generate the `requirements.txt` files:
+
+```bash
+# Install all packages for development
+docker compose run --rm server pip-compile --output-file=requirements/dev.txt requirements.in/dev.txt
+
```
BeritJanssen, line 16: This actually doesn't install the packages, but will check the `requirements.in` file, and check
- which versions are compatible with the current Python version
- which dependencies are needed for the packages listed in `requirements.in`

backend/docs/D6_Running_tests.md
```diff
@@ -0,0 +1,34 @@
...
+# or use the script
+./scripts/test-frontend-watch
+```
+
+## Backend tests
+
+To run the backend tests, run the following command:
+
+```sh
+docker-compose run --rm server bash -c "python manage.py test"
```
BeritJanssen, line 30:
```suggestion
docker compose run --rm server ./manage.py test
```
This works for me, or is this not possible on every OS?

BeritJanssen (approved):
NOICE!

backend/docs/D6_Running_tests.md
Evert-R: That doesn't work for me on windows:

This does work:
`docker compose run --rm server python ./manage.py test`

But I personally never use it. I use the docker desktop terminal, or attach vscode to the server container and use the terminal there.

I can list these options as well, including the modified `docker compose run --rm server python ./manage.py test`, if that works for all platforms.

BeritJanssen: I do think without the `bash -c` it becomes slightly less wordy. Documenting other ways to run tests via Docker is indeed also a good addition!

PR #1503, opened 2025-02-04
Fixed: Bring back final cup and center align text

### LINKED ISSUE #1502: 🐛 [BUG] - Experiments don't show trophies anymore
### Description

I just wondered whether to make a release branch, but realized that right now we are not release ready:

### Screenshots

Image

(End screen Toontjehoger Kids - Relatief gehoor)

Should be:

Image

BeritJanssen:
Close #1502 . @drikusroor , can you still check if the `text-align:center` will muddle anything when the content here is html rather than plain text?


drikusroor (commented):
Comparison with & without `text-center`:

## Without

image

## With

image

BeritJanssen:
For plain text, centered text is better. So I need to add a class perhaps depending on whether or not the text is html. 🤔

drikusroor:
> For plain text, centered text is better. So I need to add a class perhaps depending on whether or not the content is html. 🤔

You also experiment this element in the backend and always render html but wrap the plain text in a `<center>` element?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/center

drikusroor:
#1506

I've created this PR target to this PR's branch that might help you with testing the content alignment in Storybook.

backend/experiment/actions/final.py
```diff
@@ -160,13 +162,19 @@ def action(self) -> FinalActionResponse:
...
         return None
+
+    def wrap_plain_final_text(self):
+        '''check if `final_text` starts with a html tag
+        If not, wrap it in a `<center>` element for better alignment
+        '''
+        tag_pattern = re.compile('<[a-z]*>')
+        if not re.match(tag_pattern, self.final_text):
+            return f'<center>{self.final_text}</center>'
+        return self.final_text
```
BeritJanssen, line 180: @drikusroor do you see any adders under the grass with this solution?

docker-compose.yaml
```diff
@@ -103,6 +103,7 @@ services:
       - VITE_STRICT=${FRONTEND_STRICT}
     ports:
       - 3000:3000
+      - 6006:6006
```
BeritJanssen, line 106: This is needed to run Storybook via Docker, right? At least I couldn't get it to work without.

frontend/DockerfileLegacy
BeritJanssen, line 1: This file was still hanging around doing nothing. I decided to also get rid of it in this branch.

docker-compose.yaml
drikusroor: Yes, that's correct. To expose Storybook's port

backend/experiment/actions/final.py
drikusroor: Not immediately, but perhaps some test scenarios could be useful here?

PR #1511, opened 2025-02-10
Adjust name of continuation->verification scoring rule

### LINKED ISSUE #1462: We should document `SongSync` and `Hooked` in terms of the vocabulary that was already used in publications to describe the stratagem
> We should document `SongSync` and `Hooked` in terms of the vocabulary that was already used in publications to describe the stratagem - not sure if this does so, can be left for later.

_Originally posted by @BeritJanssen in https://github.com/Amsterdam-Music-Lab/MUSCLE/pull/1453#discussion_r1904009282_

BeritJanssen:
Close #1462 .

### DIFF
--- a/backend/experiment/actions/wrappers.py
+++ b/backend/experiment/actions/wrappers.py
@@ -139,7 +139,7 @@ def song_sync(
                         "correct_place",
                         session,
                         section=section,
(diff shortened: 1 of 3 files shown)


jaburgoyne (approved):
This seems like a good change to me – thank you!

PR #1515, opened 2025-02-11
Added: Add conditional support for Plausible Analytics domain configuration

drikusroor:
Introduce a new environment variable for Plausible Analytics domain, enabling conditional loading of the analytics script based on its validity. I think it would be better if the platform wouldn't have hard-coded references to the domains that _we_ as AML use, nor should we be too opiniated about whether to use an analytics platform and if so, which one.

### So how does it work?

Whenever you add the `FRONTEND_PLAUSIBLE_DOMAIN` environment variable in your `.env` file (or in GitHub's environment variables) and set it to a valid domain, such as `amsterdammusiclab.nl`, the MUSCLE frontend will dynamically add that script tag to the page. I've verified it locally and I can see it being added and making contact with plausible:

image

image


.env.dist
```diff
@@ -30,3 +30,4 @@ FRONTEND_HTML_OG_URL="" # optional: information for web crawlers
 FRONTEND_HTML_BODY_CLASS="" # optional: assign a css class to the <body> tag in index.html, for custom styling
 FRONTEND_SENTRY_DSN="" # optional: link to sentry instance, e.g. https://xxx@xxx.ingest.sentry.io/xxx
 FRONTEND_STRICT=False # optional: use StrictMode for development
+FRONTEND_PLAUSIBLE_DOMAIN="" # domain for plausible analytics (optional)
```
drikusroor, line 33:
```suggestion
FRONTEND_PLAUSIBLE_DOMAIN="" # optional: domain for plausible analytics
```

PR #1516, opened 2025-02-11
Make `emphasizeTitle` a style option

### LINKED ISSUE #175: Add styling option to frontend `Question` component
#170 introduces a mechanism to display question texts in larger font - to avoid repetition of code, this functionality should be moved to the `Question` parent component.

BeritJanssen:
Close #175 .


frontend/src/components/Question/Question.scss
```diff
@@ -28,7 +28,7 @@
             font-size: 1.3em;
         }
 
-        &:not(.title){
+        &:not(.emphasize-title){
             text-align: center;
         }
     }
```
drikusroor, line 34: I don't know the entire context, but wouldn't it be better to align to the center by default and do something when `.emphasize-title` is present?

drikusroor (approved):
LGTM

PR #1520, opened 2025-02-12
Util: Add PostgreSQL data migration script

drikusroor:
Introduce a script to facilitate the migration of PostgreSQL data between versions, automating the process of dumping and restoring databases using Docker containers.

### DIFF
--- a/scripts/migrate-postgres-data
+++ b/scripts/migrate-postgres-data
@@ -1,0 +1,116 @@
+#!/bin/bash
+set -e
+
+# ----------------------------------------------------------------------------
+# Step 0: Shut down all running containers for the project.
+#
+echo "Shutting down all project containers..."
+docker compose down
+
+# ----------------------------------------------------------------------------
+# Step 1: Run 'docker compose up db' to capture version error output.
+#
+echo "Running 'docker compose up db' to capture version error output..."
+LOGS=$(docker compose up db 2>&1)
+echo "$LOGS"
+
+# ----------------------------------------------------------------------------
+# Remove any lingering database container so the volume isn’t in use.
+#
+echo "Removing any lingering 'db' container from docker compose..."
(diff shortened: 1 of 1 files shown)


scripts/migrate-postgres-data
```diff
@@ -0,0 +1,116 @@
...
+#
+echo "Starting a temporary PostgreSQL $OLD_VERSION container for data dump..."
+docker run --name pg_old -d \
+  -v muscle_db_data:/var/lib/postgresql/data \
+  postgres:"$OLD_VERSION"
+
+# ----------------------------------------------------------------------------
+# Step 4: Wait until PostgreSQL inside the container is ready.
+#
+echo "Waiting for PostgreSQL to be ready..."
```
Evert-R, line 51: The script hangs here, with this error in the container in docker desktop:

```
2025-02-12 13:59:18 Error: Database is uninitialized and superuser password is not specified.
2025-02-12 13:59:18        You must specify POSTGRES_PASSWORD to a non-empty value for the
2025-02-12 13:59:18        superuser. For example, "-e POSTGRES_PASSWORD=password" on "docker run".
2025-02-12 13:59:18 
2025-02-12 13:59:18        You may also use "POSTGRES_HOST_AUTH_METHOD=trust" to allow all
2025-02-12 13:59:18        connections without a password. This is *not* recommended.
2025-02-12 13:59:18 
2025-02-12 13:59:18        See PostgreSQL documentation about "trust":
2025-02-12 13:59:18        https://www.postgresql.org/docs/current/auth-trust.html
```

BeritJanssen: I ran into something similar when doing this manually. I think what happens is that the env variables aren't loaded in this way. I solved it by running docker compose up db and then running the script afterwards.

PR #1523, opened 2025-02-17
Use Circle rather than ListenFeedback for Preload

### LINKED ISSUE #1394: Rename and/or reorganize Listen / ListenFeedback / Circle / ListenCircle components
There are still some incomplete refactors left in the frontent components used for Preload and Autoplay:
- the `ListenFeedback` component defines Yes / No buttons on top of the animated circle during playback. This used to be the view that was used for all `Hooked` experiments, but is obsolete now, as the setup has become more modular. The only component that uses `ListenFeedback` at the moment is `Preload`. `Listen` seems to be used nowhere except for in the `ListenFeedback` component. As both of them are wrappers around `Circle` and `ListenCircle`, we might also let Preload directly use those two components, which means we may be able to get rid of `Listen(Feedback)` altogether.
- Autoplay has nested components `Circle` and `ListenCircle`. I think it would be clearer if the last component was called `CircleContent` instead (that is also the class name of its encapsulating div).

BeritJanssen:
Close #1394 . All the components in the `Listen` folder were more or less deprecated (they included form elements and/or combinations of instructions and `Circle` elements). With this change, `Preload` loses the ability to present a countdown. We should revisit this component should an experiment require a countdown prior to a playback.


frontend/src/components/AnimatedCircleContent/AnimatedCircleContent.tsx
```diff
@@ -1,17 +1,17 @@
...
 
-const ListenCircle = ({
+const AnimatedCircleContent = ({
     duration,
     countDownRunning = true,
     histogramRunning = true
-}: ListenCircleProps) => {
+}: AnimatedCircleContentProps) => {
     return (
         <>
```
drikusroor, line 16:
```suggestion
```

frontend/src/components/AnimatedCircleContent/AnimatedCircleContent.tsx
```diff
@@ -1,17 +1,17 @@
 import Histogram from "../Histogram/Histogram";
 import CountDown from "../CountDown/CountDown";
 
-interface ListenCircleProps {
+interface AnimatedCircleContentProps {
     duration: number;
     countDownRunning?: boolean;
     histogramRunning?: boolean;
```
drikusroor, line 7:
```suggestion
```

frontend/src/components/AnimatedCircleContent/AnimatedCircleContent.tsx
```diff
@@ -1,17 +1,17 @@
...
 }
 
-const ListenCircle = ({
+const AnimatedCircleContent = ({
     duration,
     countDownRunning = true,
     histogramRunning = true
```
drikusroor, line 13:
```suggestion
```

frontend/src/components/AnimatedCircleContent/AnimatedCircleContent.tsx
```diff
@@ -1,17 +1,17 @@
...
 
-interface ListenCircleProps {
+interface AnimatedCircleContentProps {
     duration: number;
     countDownRunning?: boolean;
     histogramRunning?: boolean;
 }
 
-const ListenCircle = ({
+const AnimatedCircleContent = ({
```
drikusroor, line 10: You could also consider making this a dumb component that just shows children with a circle around. Then this component wouldn't need props and you could just use it like this (in `AutoPlay` for example):

```tsx
<AnimatedCircle>
  <Countdown duration={duration} countDownRunning={countDownRunning} />
</AnimatedCircle>
```

That would reduce complexity in my opinion and also allows for re-use of this component when you want to show different things inside of it.

drikusroor (changes requested):
I would seriously consider making the `AnimatedCircle` a dumb component. Aside from that, both tests & storybook build are failing. :-)

frontend/src/components/AnimatedCircleContent/AnimatedCircleContent.tsx
BeritJanssen: `Circle` is the dumb component you describe. `AnimatedCircleContent` is animation that groups together a histogram and countdown, as used in `Autoplay`. Granted, we don't really *need* the `AnimatedCircleContent`. Just showing the `Countdown` and `Histogram` components in `Autoplay` would reduce the number of lines. I decided to keep the outfactor, but then I also couldn't conceive of scenarios in which we would want to show `Countdown` and `Histogram`, *without* the surrounding `Circle`.

BeritJanssen:
Alright, I removed the `AnimatedCircleContent` component. Another 350 lines gone. 😸

PR #1546, opened 2025-03-25
Tunetwins

BeritJanssen:


frontend/src/components/Final/Final.tsx
```diff
@@ -13,101 +13,126 @@ import FinalButton from "./FinalButton";
...
+        {/* <div class="list-group-item">
+          {rank && <div className="text-center">
+            <Rank cup={{ className: rank?.class, text: rank.text }} score={{ score, label: points }} />
+          </div>}
+        </div> */}
+        
+        {/* Also contains the statistics */}
+        {/* <div className="aha__final-text">
+              <div dangerouslySetInnerHTML={{ __html: final_text }} />
+          </div> */}
```
BeritJanssen, line 85: Commented out code is not a good idea in general, it makes it very hard to understand what has been changed.

frontend/src/components/MCGTheme/logos/TuneTwins.tsx
BeritJanssen, line 1: This bellongs into the assets, or, even better, should be uploaded through the admin interface.

BeritJanssen (commented):
@bacor , I looked over the code. Some general comments:
- logos should be uploaded through the backend
- svgs like star, dot etc. perhaps to be moved to assets, and let the component only do the animation / sizing etc.
- don't comment out code. If you're not sure whether it should be dropped, version control is always there to bring it back.
- I'm rather opposed to calling the theme for TuneTwins "MCG Theme", since the original theme has also been developed by the MCG. :-)
- As the code stands, it will interfere with non-TuneTwins projects, so I vote against merging it into the existing code base
- Seeing as the code also introduces more than 2'000 lines which would need to be integrated and maintained, I would advise keeping it on a separate branch for now, and deploying it to its own server or container network.

PR #1580, opened 2025-08-13
That's My Song: ask questions every round after first playthrough

BeritJanssen:
This PR ensures that players who've already completed a game will get questions after every trial when playing again. That is, until all questions have been asked. @jaburgoyne is this what you had in mind?

NB: The two demographic questions in the beginning yet have to be removed. I am planning to move these to a separate Questionnaire block, but that would mean these questions will appear after Consent, before the first Explainer. If that is as intended, the logic to ask these questions can be removed from the `thats_my_song` rules file.

### DIFF
--- a/backend/experiment/rules/base.py
+++ b/backend/experiment/rules/base.py
@@ -90,6 +90,17 @@ class BaseRules(object):
         message = "You correctly identified {} out of {} recognized songs!".format(correct, total)
         return score_message + " " + message
 
+    def has_played_before(self, session):
+        """Check if the current participant has completed this game previously."""
+        previous_games = Session.objects.filter(
+            participant=session.participant,
+            block=session.block,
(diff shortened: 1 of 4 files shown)


jaburgoyne:
So I can confirm that we would like the two demographic questions *before* the Explainer

BeritJanssen:
There are no more questions attached to this experiment, so once all questions are answered, participants won't get any more. One question I still had: given the length of the experiment, the condition in this PR -- namely, that after a *full playthrough**, questions will be asked after every round, seems a noop: the list of 15 questions will be all asked within the 30-5 rounds after the question offset. So perhaps it's fine to get rid of the hocus pocus checking if a player plays again? Or, if the request is that a participant will get questions every time if they come back to the game at a later stage, without having it finished previously, I would still have to change the condition for getting a question after every round accordingly.

